PR #9, opened 2021-06-15
chore(messaging): Correct clippy complaints in messaging/client

connec:
Some of the docs are light/vapid because I don't have enough context to write anything insightful – happy to update!


src/messaging/client/errors.rs
```diff
@@ -65,7 +65,7 @@ pub enum Error {
     DuplicateEntryKeys,
     /// The list of owner keys is invalid
     #[error("Invalid owner key: {0}")]
-    InvalidOwners(sn_data_types::PublicKey),
+    InvalidOwners(PublicKey),
     /// No Policy has been set to the data
```
dirvine, line 69: For these self-describing fields we are fine with a `#[allow(missing_docs)] ` decoration

connec: I did not add these docs, but that makes sense especially for `Error` derives that are better-documented by their attributes. I can remove them since I'm in the area!

connec: Hrm, on second thought the annotation isn't visible in the documentation (e.g. https://docs.rs/sn_messaging/36.0.1/sn_messaging/client/enum.Error.html) so some commentary might be worthwhile.

I'm inclined to leave this as-is for now since it's not causing any CI issues.

oetyng:
The added rows between fields in structs, are those intentional?
I can't remember a vocal convention, but in practice I think there are never added rows between the struct fields.

joshuef (approved):

connec:
@oetyng I did add them deliberately. I saw both styles in the code, and prefer with spacing (I find it easier to read). Happy to stick with convention if there is one, though.

PR #11, opened 2021-06-15
refactor(relocation): make the relocate payload non-optional for a JoinAsRelocatedRequest

bochaco:


src/messaging/node/relocation.rs
```diff
@@ -6,23 +6,23 @@
...
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord, Hash, Clone)]
 pub struct RelocateDetails {
-    /// Public id of the node to relocate.
-    pub pub_id: XorName,
-    /// Relocation destination - the node will be relocated to a section whose prefix matches this
-    /// name.
-    pub dst: XorName,
+    /// Name of the node to relocate.
+    pub name: XorName,
+    /// New name the node will adopt, it will be relocated to
```
maqi, line 23: To confirm, this `new_name` is not the exact name that the relocated node shall use when join the target section.
The relocated node is only expected to join the target section with a name of matching the target section (i.e. the first prefix.len() bits of this new_name)

src/routing/routing/bootstrap/relocate.rs
```diff
@@ -34,40 +33,54 @@ pub(crate) struct JoiningAsRelocated {
...
 impl JoiningAsRelocated {
     pub fn new(
-        node: Node,
+        mut node: Node,
         genesis_key: BlsPublicKey,
         relocate_details: SignedRelocateDetails,
     ) -> Result<Self> {
         let section_key = relocate_details.dst_key()?;
 
+        // FIXME: we need to be provided with the keypair, generating a random one
```
maqi, line 49: age is bounded witnin name, which is part of the keypair.
So, if using the passed down keypair, that keypair needs to ensure fullfil the criterias of `name_prefix + age` .

src/routing/routing/bootstrap/relocate.rs
```diff
@@ -83,20 +96,52 @@ impl JoiningAsRelocated {
...
+                }))
+            }
+            JoinAsRelocatedResponse::Retry(section_auth) => {
+                let new_name = self.relocate_payload.details.new_name()?;
+                if !self.check_autority_provider(&section_auth, &new_name) {
+                    return Ok(None);
+                }
+
+                // FIXME: verify variant/join_response
+                // if !verify_message(&routing_msg, None) {
```
maqi, line 141: when comment out this block of code ?

src/routing/routing/bootstrap/relocate.rs
```diff
@@ -107,25 +152,29 @@ impl JoiningAsRelocated {
...
             }
-            Some(JoinAsRelocatedResponse::Redirect(section_auth)) => {
+            JoinAsRelocatedResponse::Redirect(section_auth) => {
+                let new_name = self.relocate_payload.details.new_name()?;
+                if !self.check_autority_provider(&section_auth, &new_name) {
+                    return Ok(None);
+                }
+
+                // FIXME: verify variant/join_response
+                // if !verify_message(&routing_msg, None) {
```
maqi, line 174: same question, why comment out this block of code ?
do you mean the function is only for normal join response, not for JoinAsReloatedResponse ?

PR #23, opened 2021-06-16
chore: rename Signed to KeyedSig

oetyng:
The struct `Signed` wraps a signature that we've supplemented with the corresponding key. Therefore, simple and straight forward way is to call it `KeyedSig`.

### Background
"Signed" means the object we're speaking of, that we created a signature over. It's not the sig itself (not even if coupled with the pubkey).
So for example, this struct below is (eh, more or less, still programming..) grammatically correctly named:

```
pub struct SectionSigned<T: Serialize> {
    /// some value to be agreed upon by elders
    pub value: T,
    /// signature over the value
    pub signed: Signed,
}
```
.. as we have a _value_ that is _signed_ (signed is a **property** of the value, not an object),
but then the second member cannot also be _signed_, rather it is _signature_.

Additional updates across the code to align with this language convention.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ homepage = "https://maidsafe.net"
 license = "GPL-3.0"
 name = "safe_network"
 readme = "README.md"
-repository = "https://github.com/maidsafe/the_safe_network"
+repository = "https://github.com/maidsafe/safe_network"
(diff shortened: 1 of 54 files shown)


joshuef (approved):
lgtm

PR #24, opened 2021-06-16
ci: allow all jobs to run, but only test with specific patterns

joshuef:

### DIFF
--- a/.github/workflows/pr_client.yml
+++ b/.github/workflows/pr_client.yml
@@ -1,11 +1,6 @@
 name: Client Changes
 
-
-on: 
-  pull_request:
-    paths: 
-      - 'src/client/**'
-      - 'src/messaging/**'
+on: pull_request
 
 
 env:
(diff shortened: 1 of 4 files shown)


connec (approved):
LGTM. Still some other failures (udeps, test publish) which I guess can be fixed separately?

Also, it seems there are some E2Es still running after ~4 hours (https://github.com/maidsafe/safe_network/actions/workflows/e2e.yml) – maybe something to keep an eye on?

joshuef:
> LGTM. Still some other failures (udeps, test publish) which I guess can be fixed separately?
yeh, i'm not sure what the udeps situation is. Is that what @S-Coyle was fixing recently?

>Also, it seems there are some E2Es still running after ~4 hours
Aye! i removed that blob example test. I havent dug into why it's hanging (no output from the GHA there). But that seems to have been the cause of that

connec:
In this case it seems `tracing-appender` could be a legitimate unused dependency (e.g. `cargo udeps` is running properly).

PR #25, opened 2021-06-16
Fix routing examples

connec:
- 496596514 **fix(examples): Migrate routing examples to crate root**

  It's not possible to use `cargo` to run examples in nested directories,
  so the routing examples have been promoted to the crate root. This
  required some changes to the examples to make them compatible with the
  renamed crate, as well as the addition of `tokio-util` as a
  dev-dependency to the root crate.

- 40a480aa7 **chore: Format dev-dependencies in Cargo.toml**

  They were out of order and there were some duplicates.

- b5eb048ed **ci: Remove routing/Cargo.toml to fix publish test**


S-Coyle (approved):
LGTM 👏

PR #34, opened 2021-06-16
Move sn deps into repo

oetyng:


joshuef:
@oetyng for sure we'll need PR.yml for running any specific tests against changes in these subs. (And probably update the other workflows to trigger on changes to these if they are dependent; ie DTs should trigger all tests) 

eg: https://github.com/maidsafe/safe_network/blob/main/.github/workflows/pr_node.yml#L32-L44

joshuef (approved):

joshuef (approved):

PR #36, opened 2021-06-16
refactor: reorganising files in msg handling module

bochaco:
This is also moving the `core` module out of `routing_api` mod, so keep the core isolated from any public api.

This PR doesn't change any logic and/or flow in routing codebase, it's just for reorganising files and modules.


src/routing/core/msg_handling/join.rs
```diff
@@ -0,0 +1,236 @@
...
+        }])
+    }
+
+    pub(crate) fn handle_join_as_relocated_request(
+        &mut self,
+        peer: Peer,
+        join_request: JoinAsRelocatedRequest,
+    ) -> Result<Vec<Command>> {
+        debug!("Received {:?} from {}", join_request, peer);
+        let payload = if let Some(payload) = join_request.relocate_payload {
```
maqi, line 144: just a mark
this shall be updated once https://github.com/maidsafe/safe_network/pull/11 got merged.

maqi (commented):
the commit history showing there is a git mv of 
`src/routing/routing_api/core/connectivity.rs -> src/routing/core/msg_handling/dkg.rs`
which is little bit confusing.
Not a blocking issue though.

src/routing/core/msg_handling/join.rs
bochaco: it will be yes

maqi (approved):

PR #38, opened 2021-06-17
fix: various master workflow fixes

joshuef:


connec (approved):
Seems fine to me, for what that's worth 😄 I guess the `Makefile` from `sn_node` was lost during the repo merge?

joshuef:
yup. i over enthusiastically deleted it. I thought it wasnt used (might be good to get rid of it proper, but lets get it all going first!)

PR #39, opened 2021-06-17
Minor test fixes

connec:
Some small issues encountered when looking through the `balance_management` tests.

---

- 4e1e3d6f2 **chore(client): Fix some comments in tests**

  The comments were out of sync with the asserted values.

- 51049baec **test(client): Remove redundant statement**

  The loop will continue as long as `receiving_bal != target_tokens` which
  subsumes the explicit `receiving_bal > target_tokens` test, making the
  latter redundant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2323,7 +2323,7 @@ checksum = "71d301d4193d031abdd79ff7e3dd721168a9572ef3fe51a1517aba235bd8f86e"
 
 [[package]]
 name = "safe_network"
-version = "0.2.5"
+version = "0.2.6"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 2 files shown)


joshuef (dismissed):

connec:
Not sure what to do about the failing E2Es. They seem to be totally unrelated.

joshuef:
They are indeed. It's an intermittent error that I've only seen on CI. 

You can restart tests and it'll probably go away. (if not let's see if we can get more info)

It is something we need to dive into at some point (reads like qp2p connection failed.. but we should probably deal w/ that error somewhere and retry then eg.)

joshuef (approved):

PR #43, opened 2021-06-18
Fixes to routing and node

Yoga07:


src/types/keys/public_key.rs
```diff
@@ -89,7 +89,7 @@ impl PublicKey {
     }
 
     /// Returns the bytes of the underlying public key
-    pub fn to_bytes(&self) -> Vec<u8> {
+    pub fn to_bytes(self) -> Vec<u8> {
```
lionel1704, line 92: Is this change due to clippy? Doesn't make sense :thinking:

oetyng (approved):

PR #45, opened 2021-06-18
refactor(transfers): integrate dbc

oetyng:


src/client/client_api/transfers/mod.rs
```diff
@@ -165,88 +166,14 @@ impl Client {
...
-            }))?;
-
-        let payment_proof: TransferAgreementProof =
-            self.await_validation(cmd, signed_transfer.id()).await?;
-
-        debug!("Payment proof retrieved");
-        Ok(payment_proof)
+    ///
+    pub async fn generate_payment(&self) -> Result<Payment, Error> {
+        Ok(BTreeMap::new())
```
dirvine, line 176: prob best use `todo!()` here

src/node/node_api/mod.rs
```diff
@@ -61,6 +65,49 @@ impl NodeInfo {
...
+}
+
+impl BlsKeyManager {
+    pub fn new(network: Network) -> Self {
+        Self { network }
+    }
+}
+
+use futures::executor::block_on as block;
+
```
dirvine, line 80: I think we need to keep our use statements at top of file. It's just convention

oetyng: They are always placed in top eventually. While working on the code all sort of non conventional things can be seen.
 
This PR is still in draft and not ready for review yet.

src/client/client_api/transfers/mod.rs
oetyng: This is a draft PR and not marked as ready for review yet.

This is absolute rudimentary things that compile, no more.

There will be no todos, unimplemented or empty methods when it is marked as ready for review.

src/node/node_api/mod.rs
oetyng: Especially the
`use futures::executor::block_on as block;` will be placed right over methods where it is used, since it is never kept to production, and only used when waiting for async impl in some other place. In this case it is the `sn_dbcs` `KeyManager` trait, thst is not yet even in master. Only after that will it get async implemented.

It is not a good point in the process to look for these kind of things.

PR #46, opened 2021-06-19
Log rotation

Yoga07:


src/node/utils.rs
```diff
@@ -111,12 +111,17 @@ pub fn init_logging(config: &Config) {
 
     let level_filter = config.verbose().to_level_filter();
     let module_log_filter = format!("{}={}", NODE_MODULE_NAME, level_filter.to_string());
-    let logger = Logger::with_env_or_str(module_log_filter)
+    let logger = Logger::try_with_env_or_str(module_log_filter)
+        .map_err(|e| Error::Configuration(format!("{:?}", e)))?
         .format(do_format)
-        .suppress_timestamp();
+        .rotate(
+            Criterion::Size(1024 * 1024 * 50), // 50 mb
```
joshuef, line 118: 50mb files is _a lot_. I'd say 1mb or 5 or so. 

I'd also just keep all of them for now?

joshuef (changes requested):

joshuef (approved):

PR #52, opened 2021-06-21
Ae for transfers

joshuef:

### DIFF
--- a/src/client/client_api/mod.rs
+++ b/src/client/client_api/mod.rs
@@ -137,8 +137,6 @@ impl Client {
 
         let transfer_actor = Arc::new(RwLock::new(TransferActor::new(keypair.clone(), elders)));
 
-        let _transfer_actor_for_error_handling = transfer_actor.clone();
-
         let mut client = Self {
             keypair,
             transfer_actor,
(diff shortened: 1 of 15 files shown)


src/client/client_api/mod.rs
```diff
@@ -162,6 +170,39 @@ impl Client {
...
+
+                if expected < received {
+                    info!("An Elder is missing transfer history, updating if we have newer");
+                    // TODO: target specific elder by src
+                    // refactor send_cmd to allow optional src
+                    let our_history = transfer_actor.read().await.history();
+                    for (i, debit) in our_history.debits.iter().enumerate() {
+                        if i >= expected as usize {
+                            trace!("Sending transfer #${:?} to elder@{:?}", i, src);
+                            let _ = client
```
Yoga07, line 193: Do we want to suppress the return?

src/client/connections/messaging.rs
```diff
@@ -113,11 +113,22 @@ impl Session {
...
+        cmd: Cmd,
+        client_sig: ClientSig,
+        send_to_specific_elder: Option<SocketAddr>,
+    ) -> Result<(), Error> {
         let msg_id = MessageId::new();
         let endpoint = self.endpoint()?.clone();
 
-        let elders: Vec<SocketAddr> = self.connected_elders.read().await.keys().cloned().collect();
+        let mut elders: Vec<SocketAddr> = vec![];
+        if let Some(socket) = send_to_specific_elder {
```
Yoga07, line 126:
```suggestion
   let elders = if let Some(socket) = send_to_specific_elder {
           vec![socket]
      } else {
            elders = self.connected_elders.read().await.keys().cloned().collect::<Vec<SocketAddr>>()
        }
```

src/client/client_api/mod.rs
joshuef: i'll log it uppp

Yoga07 (approved):

PR #61, opened 2021-06-22
Fs async

joshuef:

### DIFF
--- a/src/client/client_api/map_apis.rs
+++ b/src/client/client_api/map_apis.rs
@@ -668,8 +668,8 @@ mod tests {
     use super::*;
     use crate::client::utils::test_utils::{create_test_client, gen_ed_keypair};
     use crate::messaging::client::{CmdError, Error as ErrorMessage};
-    use crate::retry_loop_for_pattern;
     use crate::types::{MapAction, MapKind, Token};
+    use crate::{retry_loop, retry_loop_for_pattern};
     use anyhow::{anyhow, bail, Result};
     use std::str::FromStr;
     use std::time::Duration;
@@ -855,10 +855,10 @@ mod tests {
         let tag = 15001;
         let mapref = MapAddress::Unseq { name, tag };
 
-        let client = create_test_client().await?;
-        let owner = client.public_key();
+        let some_client = create_test_client().await?;
+        let owner = some_client.public_key();
 
-        let _ = client.store_unseq_map(name, tag, owner, None, None).await?;
+        let _ = retry_loop!(some_client.store_unseq_map(name, tag, owner, None, None));
 
         let mut client = create_test_client().await?;
 
@@ -887,16 +887,24 @@ mod tests {
         let _ = permissions.insert(random_user, permission_set);
 
         let test_data_name = XorName(rand::random());
-        let address = client
-            .store_seq_map(test_data_name, 15000u64, random_pk, None, Some(permissions))
-            .await?;
+        let address = retry_loop!(client.store_seq_map(
+            test_data_name,
+            15000u64,
+            random_pk,
+            None,
+            Some(permissions.clone())
+        ));
+
         let res = client.get_map_shell(address).await;
         match res {
             Err(Error::ErrorMessage {
                 source: ErrorMessage::DataNotFound(_),
                 ..
             }) => (),
-            Ok(_) => bail!("Unexpected Success: Validating owners should fail"),
+            Ok(data) => bail!(
+                "Unexpected Success: Validating owners should fail.  Data received : {:?}",
+                data
+            ),
             Err(e) => bail!("Unexpected: {:?}", e),
         };
 
--- a/src/client/client_api/sequence_apis.rs
+++ b/src/client/client_api/sequence_apis.rs
@@ -788,7 +788,7 @@ mod tests {
 
         let _ =
             retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("0")?);
-        let balance_before_delete = retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("10")?)?;
+        let balance_before_delete = retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("10")? && *bal != Token::from_str("0")?)?;
 
         client.delete_sequence(sequence_address).await?;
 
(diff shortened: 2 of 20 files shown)


Yoga07 (dismissed):

Yoga07 (approved):

src/client/client_api/map_apis.rs
```diff
@@ -887,16 +887,24 @@ mod tests {
...
         let test_data_name = XorName(rand::random());
-        let address = client
-            .store_seq_map(test_data_name, 15000u64, random_pk, None, Some(permissions))
-            .await?;
+        let address = retry_loop!(client.store_seq_map(
+            test_data_name,
+            15000u64,
+            random_pk,
+            None,
+            Some(permissions.clone())
```
grumbach, line 895: Just thinking out loud, `clone()` in a loop, wouldn't it be possible to pass a ref?

src/client/client_api/sequence_apis.rs
```diff
@@ -783,27 +783,18 @@ mod tests {
         let client = create_test_client().await?;
         let owner = client.public_key();
         let perms = BTreeMap::<PublicKey, SequencePrivatePermissions>::new();
-        let sequence_address = client
-            .store_private_sequence(None, name, tag, owner, perms)
-            .await?;
-
-        let mut balance_before_delete = client.get_balance().await?;
+        let sequence_address =
+            retry_loop!(client.store_private_sequence(None, name, tag, owner, perms.clone()));
```
grumbach, line 787: Same comment about `clone()`

src/client/client_api/sequence_apis.rs
```diff
@@ -783,27 +783,18 @@ mod tests {
...
 
-        while balance_before_delete == Token::from_str("0")?
-            || balance_before_delete == Token::from_str("10")?
-        {
-            sleep(Duration::from_millis(200)).await;
-            balance_before_delete = client.get_balance().await?;
-        }
+        let _ =
+            retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("0")?);
+        let balance_before_delete = retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("10")?)?;
```
grumbach, line 791: This is one more `get_balance()` than the impl before, I suppose `get_balance()` problably has no side effects, but could it change between the 2 calls?

src/client/client_api/sequence_apis.rs
```diff
@@ -875,38 +852,22 @@ mod tests {
...
+            retry_loop!(client.get_sequence_private_permissions_for_user(address, owner));
 
         assert!(user_perms.is_allowed(SequenceAction::Read));
         assert!(user_perms.is_allowed(SequenceAction::Append));
 
         let mut seq_res = client
             .get_sequence_permissions(address, SequenceUser::Key(owner))
             .await;
+
         while seq_res.is_err() {
```
grumbach, line 869: Why not use the `retry_loop!` marco this time?

src/node/data_store/tests.rs
```diff
@@ -161,10 +161,31 @@ async fn delete() -> Result<()> {
             value: data.clone(),
         };
         data_store.put(the_data).await?;
+
+        let mut used_space = data_store.total_used_space().await;
+
+        while used_space != *size {
+            used_space = data_store.total_used_space().await;
+            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        }
```
grumbach:
```suggestion
        loop {
            let used_space = data_store.total_used_space().await;
            if used_space == *size {
                break;
            }
            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
        }
```
No mut, only one call. What do you think?

src/node/data_store/tests.rs
```diff
@@ -161,10 +161,31 @@ async fn delete() -> Result<()> {
...
+        assert!(!data_store.has(&the_data.id).await);
+
+        let mut used_space = data_store.total_used_space().await;
+
+        while used_space != 0 {
+            used_space = data_store.total_used_space().await;
+            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        }
+
```
grumbach:
```suggestion
        loop {
            let used_space = data_store.total_used_space().await;
            if used_space == 0 {
                break;
            }
            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
        }
```

src/node/data_store/tests.rs
```diff
@@ -211,8 +232,16 @@ async fn overwrite_value() -> Result<()> {
                 value: data.clone(),
             })
             .await?;
+
+        let mut used_space = data_store.total_used_space().await;
+
+        while used_space != size {
+            used_space = data_store.total_used_space().await;
+            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
+        }
```
grumbach, line 249:
```suggestion
        loop {
            let used_space = data_store.total_used_space().await;
            if used_space == size {
                break;
            }
            tokio::time::sleep(std::time::Duration::from_millis(200)).await;
        }
```

grumbach (commented):
A couple of comments and code suggestions but nothing game changing. Maddening amount of await additions! 🥇

src/client/client_api/map_apis.rs
joshuef: hmm could be. But we'd then have to refactor that API as well. Which maybe we want, but this is just tests so I'd lean on not doing that just for that

src/client/client_api/sequence_apis.rs
joshuef: yeh, this is basically the same as previous, but the loop pattern matching doesnt let us (or I didn't figure out), how to get it to match against an or here.

Balance could change, but wont go to 0 during this test.

grumbach:
```suggestion
                let balance_before_delete = retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("10")? || *bal != Token::from_str("0")?)?;
```
Wouldn't this do the trick?

grumbach, line 791:
```suggestion
        let balance_before_delete = retry_loop_for_pattern!(client.get_balance(), Ok(bal) if *bal != Token::from_str("10")? && *bal != Token::from_str("0")?)?;
```

PR #65, opened 2021-06-22
fix: merge remote section info during Sync

maqi:

### DIFF
--- a/src/routing/network/mod.rs
+++ b/src/routing/network/mod.rs
@@ -162,7 +162,26 @@ impl NetworkUtils for Network {
             key_sig,
         };
 
-        if !info.verify(section_chain) {
+        // With the change of AE, the voting of OtherSection is removed, which means the info of
+        // remote section is no longer being signed by own section.
+        // As this passed in section_chain is just our own chain during sync, which in a high chance
+        // won't contain the section_key of a remote section.
+        // So the handling logic is now changed to:
+        //     1, if section_chain validates, update as it is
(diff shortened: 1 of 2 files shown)


src/routing/network/mod.rs
```diff
@@ -162,7 +162,26 @@ impl NetworkUtils for Network {
...
+        //     1, if section_chain validates, update as it is
+        //     2, if chain not validate but the if is self-validated and we don't have such entry,
+        //        we take that entry in.
+        // In case the entry turned out to be an outdated one, it will got updated via AE flow.
+        //
+        // Note this is just a temp resolvement. It's still being discussed whether shall bring back
+        // the re-vote to improve the security.
+
+        if !info.verify(section_chain)
+            && (!info.self_verify() || self.sections.get(&section_auth.value.prefix).is_some())
```
Yoga07, line 179: Instead of self verifying, can we pass the last known key for the incoming prefix and verify from that common point to the latest key?

maqi: not sure what you mean of `verify from that common point to the latest key?` 
the `last known key` we hold of the remove section is just a `leaf` and could be `not in our chain` already by high chance.

Yoga07: I mean, the last key we know of the received section's parent prefix is trusted. Can we use that key and to verify the incoming chain.

joshuef (approved):

src/routing/network/mod.rs
maqi: >Can we use that key and to verify the incoming chain.

PR #66, opened 2021-06-22
fix: remove the client timeout

grumbach:
Remove the client timeout (that made many tests fail), and adapt the pattern matching accordingly.


joshuef (approved):
lgtm, nice one!

PR #81, opened 2021-06-23
ci: split up e2etests, dont log so aggresively

joshuef:

### DIFF
--- a/src/client/client_api/transfers/balance_management.rs
+++ b/src/client/client_api/transfers/balance_management.rs
@@ -276,7 +276,7 @@ mod tests {
             let _ =
                 retry_loop_for_pattern!( client.get_balance(), Ok(bal) if *bal == correct_balance);
 
-            if count == 50 {
+            if count == 7 {
                 break;
             }
         }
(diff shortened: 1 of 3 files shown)


src/client/client_api/transfers/balance_management.rs
```diff
@@ -276,7 +276,7 @@ mod tests {
             let _ =
                 retry_loop_for_pattern!( client.get_balance(), Ok(bal) if *bal == correct_balance);
 
-            if count == 50 {
+            if count == 7 {
```
connec, line 279: How did you choose 7, out of curiosity?

connec (approved):
Only issue I see with this would be new test modules not being picked up automatically, but not a big deal.

Would it make sense, as a further future enhancement, to run the different modules in parallel, or would that not work against a single network?

src/client/client_api/transfers/balance_management.rs
joshuef: hmm, i think this is what we were doing before.

but it's a random number. Everything is in seq, 7 in a row means stuff is going okay. 50 in a row takes like 5 mins on my machine, so probs adds 15 mins to CI

connec (approved):

joshuef:
> Would it make sense, as a further future enhancement, to run the different modules in parallel, or would that not work against a single network?

Im not sure we can have one network used by many jobs. But it would make sense if we could. They _might_ slow each other down though.

Down the line having a constant DO testnet to test against may be the answer here (when not having breaking changes...)

PR #95, opened 2021-06-25
feat(responses): drop data error responses if NoSuchChunk is encountered

Yoga07:


src/node/metadata/map_storage.rs
```diff
@@ -296,6 +297,7 @@ impl MapStorage {
             .map(|data| data.version())
         {
             Ok(res) => Ok(res),
+            Err(Error::NoSuchChunk(addr)) => return Err(Error::NoSuchChunk(addr)),
```
lionel1704, line 300: What happens to this error? Will the node try to handle it somewhere?

src/node/metadata/chunk_records.rs
```diff
@@ -182,6 +182,12 @@ impl ChunkRecords {
             if !response.is_success() && self.capacity.is_full(&src).await {
                 // We've already responded already with a success
                 // so do nothing
+            } else if !matches!(
+                response,
+                QueryResponse::GetChunk(Err(crate::messaging::client::Error::DataNotFound(_)))
```
lionel1704, line 187: If `!matches!(response, Error(...))`
I think the `!` is extra there?

Yoga07: Ah yes! IDE issues, will be removing 👍

src/node/metadata/map_storage.rs
Yoga07: We map that error to do nothing [here](https://github.com/Yoga07/sn/blob/5c4668a31e158ea35bd9c36c181b929851c68416/src/node/metadata/mod.rs#L76).

lionel1704 (approved):

src/client/client_api/queries.rs
```diff
@@ -11,6 +11,8 @@ use crate::client::{connections::QueryResult, errors::Error};
 use crate::messaging::client::{ClientSig, Query};
 use crate::types::{PublicKey, Signature};
 use log::debug;
+use tokio::time::Duration;
+const QUERY_TIMEOUT: u64 = 20; // 20 seconds
```
connec, line 15: No opinion on the value,  but `Duration::from_secs` is `const` so you should be able to do:

```rs
const QUERY_TIMEOUT: Duration = Duration::from_secs(20);
```

Yoga07: Oo! Been playing around with that value before pushing, will update :)

src/client/client_api/mod.rs
```diff
@@ -41,6 +42,9 @@ pub struct Client {
     simulated_farming_payout_dot: Dot<PublicKey>,
     incoming_errors: Arc<RwLock<Receiver<CmdError>>>,
     session: Session,
+    query_timeout: Duration,
+    #[cfg(test)]
+    pub override_timeout: Option<Duration>,
```
lionel1704, line 47: Doesn't `pub(crate) query_timeout` make this simpler?

lionel1704 (dismissed):

lionel1704 (approved):

oetyng (approved):

PR #98, opened 2021-06-28
Fix interrupted join loop

lionel1704:
- fix(comm): properly close temp endpoint created for is_reachable() test
- refactor(main): refactor main function to use `tokio::main` and it's default runtime options


maqi (approved):

Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "GPL-3.0"
 name = "safe_network"
 readme = "README.md"
 repository = "https://github.com/maidsafe/safe_network"
-version = "0.2.20"
```
joshuef, line 11: i dont think we should be changing this here. might break the realease/bump flow

joshuef (changes requested):

Cargo.toml
Yoga07: Hmm, that wasn't intentional 🤔 Seems IDE does that or cargo.

lionel1704: Yea I just made the change to see if `Cargo.toml` update will not use the cache. Blind attempt 😬

oetyng (approved):

PR #99, opened 2021-06-28
Fix ci

joshuef:


src/transfers/wallet.rs
```diff
@@ -100,7 +100,7 @@ impl Wallet {
 
     /// Mutates state.
     pub fn apply_credit(&mut self, credit: Credit) -> Result<()> {
-        debug!("Wallet applying credit");
+        trace!("Wallet applying credit");
```
maqi, line 103: the logged info seems for debugging only. 
id do want to move it into `trace` level, maybe better include more stuff within the log msg, say `Which wallet is applying how much credit` ?

maqi (commented):
suprised the lock file got so many diffs.

Also, seems the PR contains some un-necessary merging commits.
better got the work `rebased` to the main head, instead of `merge` ?

maqi (approved):

PR #108, opened 2021-06-29
tests: tweak timing

joshuef:


lionel1704:
Could we add the timeout as an argument for `create_test_client_for(....)` and `retry_loop(....)` ? so keep it short for other tests and increase it for the 10mb tests.

joshuef:
I've added the option to pass timeout into ``create_test_client`, which seems fine. the macroo i'm less sure about. That's just wait time between retries... for blobs that shouldn't make any difference (we've already waited Xs), for other ops it should still be fine

I fear it'll make the macro messy...

Cargo.toml
```diff
@@ -45,7 +45,7 @@ sn_launch_tool = "~0.2.0"
 thiserror = "1.0.23"
 itertools = "0.10.0"
 async-trait = "0.1.42"
-secured_linked_list = "~0.2.0"
+secured_linked_list = "~0.3.0"
 bls_dkg = "~0.4.0"
```
Yoga07, line 49: Latest `bls_dkg` is `v0.5.0`

src/client/client_api/mod.rs
```diff
@@ -148,6 +150,7 @@ impl Client {
             session,
             incoming_errors: Arc::new(RwLock::new(err_receiver)),
             query_timeout: Duration::from_secs(query_timeout),
+            blob_cache: Arc::new(Mutex::new(LruCache::new(150))),
```
Yoga07, line 153: Have the 150 as a const?

src/client/utils/test_utils/test_client.rs
```diff
@@ -31,22 +31,20 @@ pub fn init_logger() {
...
 
 /// Create a test client optionally providing keypair and/or bootstrap_config
 /// If no keypair is provided, a check is run that a balance has been generated for the client
-pub async fn create_test_client_with(optional_keypair: Option<Keypair>) -> Result<Client> {
+pub async fn create_test_client_with(
+    optional_keypair: Option<Keypair>,
+    timeout: Option<u64>,
+) -> Result<Client> {
     init_logger();
+    let timeout = timeout.unwrap_or(20);
```
Yoga07, line 45: Can use the `DEFAULT_QUERY_TIMEOUT` const here.

Yoga07 (dismissed):

Yoga07 (approved):

PR #119, opened 2021-07-01
Metadata less locks

joshuef:

### DIFF
--- a/src/node/metadata/elder_stores.rs
+++ b/src/node/metadata/elder_stores.rs
@@ -17,15 +17,16 @@ use crate::messaging::{
 use crate::node::{node_ops::NodeDuty, Error, Result};
 use crate::routing::Prefix;
 use crate::types::PublicKey;
+use std::sync::Arc;
+use tokio::sync::RwLock;
 use tracing::info;
-
 /// The various data type stores,
 /// that are only managed at Elders.
 pub(super) struct ElderStores {
-    chunk_records: ChunkRecords,
-    map_storage: MapStorage,
-    sequence_storage: SequenceStorage,
-    register_storage: RegisterStorage,
+    chunk_records: Arc<RwLock<ChunkRecords>>,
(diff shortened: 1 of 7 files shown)


src/node/metadata/elder_stores.rs
```diff
@@ -36,38 +37,54 @@ impl ElderStores {
...
         msg_id: MessageId,
         requester: PublicKey,
         origin: EndUser,
     ) -> Result<NodeDuty> {
         match &query {
-            DataQuery::Blob(read) => self.chunk_records.read(read, msg_id, origin).await,
-            DataQuery::Map(read) => self.map_storage.read(read, msg_id, requester, origin).await,
+            DataQuery::Blob(read) => {
+                self.chunk_records
+                    .write()
```
dirvine, line 57: Should this not be `read`

joshuef: no as we (currently) write the liveness checks inside here. 

Something to do (was leaving it for another PR) is to return another `NodeDuty` for writing read receipts for adults so we can have that compeltely separate from chunk storage.

src/node/metadata/elder_stores.rs
```diff
@@ -78,40 +95,53 @@ impl ElderStores {
             DataCmd::Blob(write) => {
                 info!("Writing Blob");
                 self.chunk_records
+                    .write()
```
lionel1704, line 98: I think this can be `read()` since there is nothing being written here.

lionel1704: We only send the data to Adults. (no liveliness checks too)

joshuef: not yet, that'll be possible with the https://github.com/maidsafe/safe_network/pull/121 in i think

dirvine (approved):

joshuef:
Closing as this should get atop transfer removal for simplicity (or wont be needed there)

PR #141, opened 2021-07-06
Tidy lints

connec:
- 8dc76bbff **chore: Remove redundant lint configuration**

  Many of the lints specified for `#[warn(...)]` are warn-by-default and
  have been removed. Additionally:
  
  - `clippy::wrong_pub_self_convention` is deprecated and is now a no-op.
  - comments have been updated to describe the purpose of the
    configuration, on the assumption that folk can discover lint meanings
    themselves (whereas why we're using them can't be Googled).

- 67bf67d58 **chore: Remove duplicate attributes from crate root**

  These were presumably duplicated when the crates were merged.

- 7e02bf3d0 **chore: Update favicon URL**

  The old URL no longer exists. The new URL was taken from the
  https://maidsafe.net site.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,16 +8,11 @@
 
 //! Implementation of the "Node" node for the SAFE Network.
 
-#![doc(
-    html_logo_url = "https://raw.githubusercontent.com/maidsafe/QA/master/Images/maidsafe_logo.png",
-    html_favicon_url = "https://maidsafe.net/img/favicon.ico",
-    test(attr(deny(warnings)))
-)]
 // For quick_error
 #![recursion_limit = "256"]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/maidsafe/QA/master/Images/maidsafe_logo.png",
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -8,28 +8,23 @@
...
 #![recursion_limit = "256"]
 #![doc(
     html_logo_url = "https://raw.githubusercontent.com/maidsafe/QA/master/Images/maidsafe_logo.png",
     html_favicon_url = "https://maidsafe.net/img/favicon.ico",
     test(attr(deny(warnings)))
 )]
-// For explanation of lint checks, run `rustc -W help` or see
-// https://github.com/maidsafe/QA/blob/master/Documentation/Rust%20Lint%20Checks.md
+// Forbid some very bad patterns. Forbid is stronger than `deny`, preventing us from suppressing
+// suppressing the lint with `#[allow(...)]` et-all.
```
bochaco, line 19: Suppress one "suppressing" :)

connec: I really need to get a plugin that will format comments for me 😄

joshuef (approved):

PR #147, opened 2021-07-07
Resource logging

joshuef:
- adds resource logging at trace level
- optionally can print to stdout with `--resource-logs` flag
- adds flag to use json logging `-j` or `--json-logs`


src/node/logging/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+    let _ = tokio::task::spawn(async move {
+        loop {
+            let elapsed = last_log.elapsed();
+            if elapsed >= LOG_INTERVAL {
+                log(&mut last_log, &mut system, &ctx, print_resources_usage).await;
+            } else {
+                tokio::time::sleep(LOG_INTERVAL - elapsed).await;
+            }
+        }
+    });
```
connec, line 34: suggestion: We could simplify this by using [`tokio::time::Interval`](https://docs.rs/tokio/1.8.1/tokio/time/fn.interval.html) for the interval:

```rust
// ...
initial_log(&mut system, &ctx).await;
let _ = tokio::task::spawn(async move {
    let mut interval = tokio::time::interval(LOG_INTERVAL);
    interval.set_missed_tick_behaviour(MissedTicketBehaviour::Skip); // default is `Burst`, probably not what we want
    loop {
        interval.tick().await;
        log(...).await;
    }
});
```

src/node/logging/system.rs
```diff
@@ -0,0 +1,76 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use std::ffi::OsString;
+use sysinfo::{DiskType, ProcessExt};
+
+/// Struct containing a disk information.
+#[derive(Debug)]
+pub struct Disk {
```
connec, line 14: nitpick: This is unused.

src/node/logging/system.rs
```diff
@@ -0,0 +1,76 @@
...
+    /// Number of written bytes since the last refresh.
+    pub written_bytes: u64,
+    /// Total number of read bytes.
+    pub total_read_bytes: u64,
+    /// Number of read bytes since the last refresh.
+    pub read_bytes: u64,
+}
+
+#[derive(Debug)]
+pub struct Network {
```
connec, line 36: nitpick: This is unused.

src/node/logging/system.rs
```diff
@@ -0,0 +1,76 @@
...
+    pub total_errors_on_transmitted: u64,
+}
+
+/// Struct containing a process' information.
+#[derive(Debug)]
+pub struct Process {
+    pub memory: u64,
+    pub virtual_memory: u64,
+    pub cpu_usage: f32,
+    pub disk_usage: DiskUsage,
```
connec, line 58: nitpick: This `DiskUsage` is identical to `sysinfo::DiskUsage`, could we not use that directly rather than defining our own copy?

src/node/logging/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+use tracing::trace;
+
+const LOG_INTERVAL: Duration = std::time::Duration::from_secs(30);
+
+pub async fn run_system_logger(ctx: LogCtx, print_resources_usage: bool) {
+    let mut system = System::new_all();
+    let mut last_log = Instant::now();
+    initial_log(&mut system, &ctx).await;
+    log(&mut last_log, &mut system, &ctx, print_resources_usage).await;
+    let _ = tokio::task::spawn(async move {
```
Yoga07, line 25: Shouldn't we hold the thread handle until the process ends?

joshuef: only if you want to do something with the thread. But this would (does) jsut keep going in the new thread

connec (dismissed):

src/node/logging/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+        one_min_load_avg,
+        five_min_load_avg,
+        fifteen_min_load_avg,
+        "System load avg",
+    );
+
+    let our_pid = &(std::process::id() as usize);
+
+    for (pid, proc_) in system.processes() {
+        if *pid != *our_pid as i32 {
```
connec, line 89: `pid` is a `Pid`, which is a different alias on different platforms (e.g. [linux](https://docs.rs/sysinfo/0.19.0/sysinfo/type.Pid.html) vs [windows](https://docs.rs/sysinfo/0.19.0/x86_64-pc-windows-msvc/sysinfo/type.Pid.html)). We should use [`as_u32`](https://docs.rs/sysinfo/0.19.0/sysinfo/type.Pid.html#method.as_u32) to get a stable comparator:

```suggestion
    let our_pid = &(std::process::id() as u32);

    for (pid, proc_) in system.processes() {
        if pid.as_u32() != our_pid {
```

src/node/logging/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+use sysinfo::{System, SystemExt};
+use system::Process;
+use tokio::time::MissedTickBehavior;
+use tracing::trace;
+
+const LOG_INTERVAL: Duration = std::time::Duration::from_secs(30);
+
+pub async fn run_system_logger(ctx: LogCtx, print_resources_usage: bool) {
+    let mut system = System::new_all();
+    let mut last_log = Instant::now();
```
oetyng, line 23: `let mut last_log = Instant::now();`  can be removed now that interval is used (nice one btw).

src/node/logging/mod.rs
joshuef: aha, nice

connec (approved):

happybeing:
@joshuef once you're happy with this let me know and I'll try to find time to add it to vdash. If you can send me a sample of the logfile output that would be great too.

Also feel free to either add it to vdash yourself, or suggest how you'd like it to appear etc. 

My first thought is that some would be useful in the main page, but some might need to be on a separate second page per node log. With a single character keypress to flip between the two views.

PR #148, opened 2021-07-07
chore: remove meta_data locks in handling

joshuef:


lionel1704 (dismissed):

src/node/metadata/elder_stores.rs
```diff
@@ -67,7 +67,7 @@ impl ElderStores {
     }
 
     pub async fn write(
-        &mut self,
+        &self,
```
Yoga07, line 70: Not sure how we can make this non-mutable. 🤔 We insert an Op down the line eg: [here](https://github.com/maidsafe/safe_network/blob/078c2b4c017c13f7c5a13ecdb6555d8511f6413f/src/node/metadata/map_storage.rs#L102) for Map which requires it to be mutable.

src/node/metadata/elder_stores.rs
```diff
@@ -137,7 +137,7 @@ impl ElderStores {
         })
     }
 
-    pub async fn update(&mut self, data: DataExchange) -> Result<(), Error> {
+    pub async fn update(&self, data: DataExchange) -> Result<(), Error> {
```
Yoga07, line 140: Same as above.

src/node/metadata/map_storage.rs
```diff
@@ -60,7 +65,7 @@ impl MapStorage {
     }
 
     /// On receiving data from Elders when promoted.
-    pub async fn update(&mut self, map_data: MapDataExchange) -> Result<()> {
+    pub async fn update(&self, map_data: MapDataExchange) -> Result<()> {
```
Yoga07, line 68: Same as above

Yoga07 (approved):

PR #153, opened 2021-07-07
Ci job flow improvements

joshuef:


joshuef:
complicated and not much gain as things stand. going to close this

PR #158, opened 2021-07-08
Remove hand-written debug impls

connec:
- cb0d94b80 **chore: Remove hand-written `Debug` impls**

  In almost every case, our hand-written `Debug` impls existed exclusively
  to skip/reformat sensitive/overly verbose fields. Recently we added the
  `custom_debug` crate to our dependencies, which allows us to make these
  tweaks as annotations in the type definition, which is a massive
  reduction in boilerplate.
  
  Additionally, this commit takes the somewhat stronger stance that we
  should prefer not to customise `Debug` output, except for security
  reasons (e.g. don't show private keys) or *extremely* verbose fields
  (such as payload or chunk bytes). This is on the basis that, if you are
  looking at a type's `Debug` output, you probably want more information,
  rather than less.
  
  This will probably mean significantly more verbose `Debug` output, which
  is likely to have a detrimental impact on logging clarity. Follow-up
  commits will seek to tackle this with the following interventions:
  
  - Be more specific about what to log. E.g., rather than logging an
    entire `Map` struct, just log its address if that's what we care
    about. The idea here is to handle verbosity at the call-site, rather
    than globally reducing verbosity for the type.
  - Otherwise, apply verbosity reductions as 'far out' as possible. This
    is likely to affect messaging enums, for example, which sometimes
    contain fairly nested structure which could wind up being particularly
    verbose. Again, the principle would be to make the messaging enum
    responsible for setting its desired verbosity, rather than 'infecting'
    all of its members with those concerns.
  
  In both cases, we should be able to achieve a satisfactory level of
  verbosity using `custom_debug`, perhaps with the addition of a couple of
  formatting helpers, so we would ideally still have very few hand-written
  `Debug` impls.

- 7fbfcaddb **chore: Use abbreviated `Debug` format for byte arrays**

  The default `Debug` output for byte arrays is not very readable. The new
  impl uses `HexFmt` to print an abbreviated hexadecimal format.

- fd0e94b31 **chore(types): Short `Debug` output for `{PublicKey,Signature}::Ed25519`**

  `ed25519_dalek` uses the default derived `Debug` output, which is very
  verbose for these types as they contain large byte arrays. It's also
  inconsistent with the other variants, from `blsttc`, which use an
  abbreviated slice of the byte representation. `custom_debug` has been
  used to apply the abbreviated format to the `Ed25519` variants.

- eb15cbac9 **chore: Update blsttc to get new `Debug` impl for `PublicKeySet`**

  This brings down the verbosity of the `Debug` output for types
  containing `PublicKeySet`.

- 221c2b6fd **chore: Exclude verbose fields from `Debug` impls**

  Fields that are byte arrays or other large, opaque types are not very
  useful in `Debug` output and create a lot of clutter. We can skip the
  with `custom_debug`.

- 4e513491c **chore(messaging): Simplify `Debug` format for `NodeMsg::Sync`**

  The `Section` and `Network` types are very large, so their `Debug`
  output is quite verbose. This is not great when we're logging messages
  going through the network. The new format is based on the one removed
  previously.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -318,9 +318,9 @@ dependencies = [
 
 [[package]]
 name = "blsttc"
-version = "2.0.1"
+version = "2.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e2cf27c82f62fd876518ea11d7a29120a410efb04d8b291cbec2ac0151d8dbd"
+checksum = "731c2cd7030d34eb735f27f803508c34e03456222733fe6660467cd00e8c6b72"
 dependencies = [
  "blst",
  "byteorder",
--- a/src/messaging/data/mod.rs
+++ b/src/messaging/data/mod.rs
@@ -333,8 +333,7 @@ mod tests {
     fn debug_format_functional() -> Result<()> {
         if let Some(key) = gen_keys().first() {
             let errored_response = QueryResponse::GetSequence(Err(Error::AccessDenied(*key)));
-            assert!(format!("{:?}", errored_response)
-                .contains("GetSequence(Err(AccessDenied(PublicKey::"));
+            assert!(format!("{:?}", errored_response).contains("GetSequence(Err(AccessDenied("));
             Ok(())
         } else {
             Err(anyhow!("Could not generate public key"))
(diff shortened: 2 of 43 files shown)


connec:
Looks like we have at least one test that's asserting on debug output, which I'll review.

joshuef (approved):

PR #159, opened 2021-07-09
chore(routing): Remove an unused `IntoIter` struct

connec:
- 828ec58cd **chore(routing): Remove an unused `IntoIter` struct**

  Sadly the Rust compiler can't detect this because it is 'used' in its
  `Iterator` implementation (even though *that* is unused, Rust can't
  tell).

### DIFF
--- a/src/routing/section/section_peers.rs
+++ b/src/routing/section/section_peers.rs
@@ -13,11 +13,7 @@ use crate::messaging::{
 };
 use crate::routing::{peer::PeerUtils, section::NodeStateUtils};
 use itertools::Itertools;
-use std::{
-    cmp::Ordering,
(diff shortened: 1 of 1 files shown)


dirvine (approved):

src/routing/section/section_peers.rs
```diff
@@ -178,16 +174,6 @@ impl SectionPeersUtils for SectionPeers {
     }
 }
 
-pub struct IntoIter(btree_map::IntoIter<XorName, SectionSigned<NodeState>>);
```
connec, line 181: Note that although this is `pub`, it isn't reachable from the crate root since `routing::section::section_peers` is private and `routing::section` doesn't re-export it. So, no breaking changes.

PR #166, opened 2021-07-13
ci: Introduce ARM Build/Test and Deploy

jacderida:
Workflows have been updated as follows:

* Standard PR Checks: run both ARM and ARMv7 build checks
* Client Tests: run tests for both ARM and ARMv7
* Messaging Tests: run tests for both ARM and ARMv7
* URL Tests: run tests for both ARM and ARMv7
* Routing Tests: run tests for both ARM and ARMv7
* Main:
    - Build ARM and ARMv7 artifacts
    - Package ARM and ARMv7 artifacts
    - Add ARM and ARMv7 checksums to release description
    - Attach ARM and ARMv7 artifacts to the release

There were also a couple of misc changes:
* Removed verbose output on agreement it wasn't very useful
* Cosmetic name changes from copy/paste (e.g. "Routing Tests" -> "URL Tests")

The builds and test runs for ARM are performed with the `cross` project, which is nicely wrapping the build configuration for different architectures inside Docker containers, which in turn use qemu inside them.

The Data Types tests include 'prop' tests, which take too long to run on ARM (at least on the Github build agents). Cargo doesn't allow us to disable a subset of tests based on their namespace, so the data type tests haven't been enabled for ARM.

The Node tests also haven't been updated to include tests for ARM. This is due to some issue with the configuration of the container used by `cross`. There's some code in qp2p that creates a directory for storing configuration and the code is being denied access for creating it.

We can come back and enable both of these sets of tests at a later point.

I tested the ARM additions by running a deployment on my fork:
https://github.com/jacderida/safe_network/releases/tag/v0.7.12


joshuef (approved):
lgtm

PR #176, opened 2021-07-14
refactor(routing): removing event_tx from Comm instance since it's not needed

bochaco:


src/routing/core/comm.rs
```diff
@@ -12,21 +12,21 @@ use bytes::Bytes;
...
-    // messages handler every time we establish new connection. It's kept in an `Option` so we can
-    // take it out and drop it on `terminate` which together with all the incoming message handlers
-    // terminating closes the corresponding receiver.
-    event_tx: RwLock<Option<mpsc::Sender<ConnectionEvent>>>,
+}
+
+impl Drop for Comm {
+    fn drop(&mut self) {
+        // Close all existing connections and stop accepting new ones.
+        self.endpoint.close();
```
connec, line 28: Doesn't affect this PR, but should `Endpoint` impl `Drop` and close itself do you think?

bochaco: I was thinking the exact same thing, I agree!

bochaco: ...I just double checked and currently qp2p doesn't do it

src/routing/routing_api/dispatcher.rs
```diff
@@ -261,29 +261,7 @@ impl Dispatcher {
                 }
                 .map_err(|e: Error| e)?
             }
-            MsgKind::DataMsg(_) => {
```
connec, line 264: Not sure I understand this change, I guess it's unrelated to `event_tx`?

connec (approved):

src/routing/core/comm.rs
connec: Yes I also checked first 😄 I think we can merge this first and change it after?

src/routing/routing_api/dispatcher.rs
bochaco: yes, it's just to prevent from cloning `wire_msg` as we don't need to call `send_on_existing_connection` in a loop, we are already logging an error within `send_on_existing_connection` for each recipient we have an issue when attempting to send.

PR #186, opened 2021-07-15
fix: improved retry loop with max iterations

grumbach:
This is an attempt to fix the hanging tests that use the `retry_loop!` macro which loops indefinitely, making tests hang and hiding the repeating error.
This is my proposal for a fix: it retries a given number of times (defaults to 10 if n is not provided) and `bail`s after retrying those n times.

### DIFF
--- a/src/client/utils/test_utils/mod.rs
+++ b/src/client/utils/test_utils/mod.rs
@@ -101,14 +101,23 @@ pub fn read_network_conn_info() -> AnyhowResult<HashSet<SocketAddr>> {
 #[macro_export]
 /// Helper for tests to retry an operation awaiting for a successful response result
 macro_rules! retry_loop {
-    ($async_func:expr) => {
+    ($n:literal, $async_func:expr) => {{
+        let mut retries = $n;
         loop {
             match $async_func.await {
                 Ok(val) => break val,
-                Err(_) => tokio::time::sleep(std::time::Duration::from_millis(2000)).await,
+                Err(_) if retries > 0 => {
+                    retries -= 1;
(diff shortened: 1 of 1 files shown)


src/client/utils/test_utils/mod.rs
```diff
@@ -101,14 +101,23 @@ pub fn read_network_conn_info() -> AnyhowResult<HashSet<SocketAddr>> {
...
-    ($async_func:expr) => {
+    ($n:literal, $async_func:expr) => {{
+        let mut retries = $n;
         loop {
             match $async_func.await {
                 Ok(val) => break val,
-                Err(_) => tokio::time::sleep(std::time::Duration::from_millis(2000)).await,
+                Err(_) if retries > 0 => {
+                    retries -= 1;
+                    tokio::time::sleep(std::time::Duration::from_millis(200)).await;
```
grumbach, line 111:
```suggestion
                    tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
```

Yoga07 (approved):

bochaco (approved):

joshuef (changes requested):
We already ahve retry w backoff which does this

joshuef:
@grumbach , could be good here to unify everything; either just straight up using the funcs, or using them via macro. 

I'm not sure why we'd use one over the other to be honest (in terms of dev ux... we want the backoff property I think)... but going for one is probably a good idea

joshuef:
Perhaps it should not be a test util either and just a helper we cna export for consumers to use too?

connec:
I think this can be closed, if for no other reason than `retry_loop!` isn't used.

PR #187, opened 2021-07-15
Client authority by construction

connec:
- e355f5628 **feat(messaging): Add a `ClientAuthority` 'proof' struct**

  The `ClientAuthority` struct can only be constructed by verifying a
  `ClientSigned` for a particular payload. Thus, instances of
  `ClientAuthority` can be trusted as proof that the payload indeed had a
  valid signature, and was signed by the owner of `public_key` (e.g. a
  holder of the corresponding secret key).
  
  Note that it's important that `ClientAuthority` **does not** implement
  `Deserialize`, or any other trait that could allow unverified
  construction. It therefore cannot be a drop-in replacement for
  `ClientSigned`, which is kind of the point as we can now introduce
  `ClientAuthority` in areas where we expect the signature to be valid,
  then work our way back towards verification. Ideally we should be able
  to verify immediately and use `ClientAuthority` in `MessageType`.

- 412984884 **refactor(messaging)!: Use `ClientAuthority` in `MessageType`**

  This is pretty non-functional currently. In fact, we end up verifying
  the signature *twice*, since we also verify in
  `routing::Core::handle_message`. We should deduplcate this in future.
  
  BREAKING CHANGE: The `MessageType::Client` variant's `client_signed`
  field has been removed, and replaced with `client_auth`.

- 9c11f12e2 **fix(node): Ensure storage write op authority is verified**

  Previously, we would only verify the signatures of the original client
  message, not the signatures of replicated ops. With this change, a
  `ClientAuthority` is now needed to apply an op, meaning the signature
  needs to be verified in order to call those functions.
  
  Currently, verifications happens quite late, inside each individual
  store. This is because the `DataExchange` content is not iterated until
  this point. We really want to do this verification earlier, however.

- b0e8385bd **fix(node): Stores always verify client authority**

  We use `ClientAuthority` in `NodeDuty` to reflect the fact that, by the
  time we have constructed a `NodeDuty`, we expect the client authority to
  be valid. Note that authentication is still a separate step, but the
  public key in the authority can be trusted to have signed the op.
  
  Ultimately, this makes it impossible to call any of the store functions
  without a `ClientAuthority`, which can only be constructed by verifying
  a `ClientSigned` for the op.

- 85a01c19e **refactor(node): Rework storage APIs to avoid double-verifying**

  This doesn't completely remove double-verification (e.g. when dealing
  with the contents of a `MessageType::Client`, which is still verified on
  deserialization and again on conversion).

### DIFF
--- a/src/node/event_mapping/node_msg.rs
+++ b/src/node/event_mapping/node_msg.rs
@@ -8,12 +8,12 @@
 
 use super::{Mapping, MsgContext};
 use crate::messaging::{
-    data::QueryResponse,
+    data::{DataCmd, DataMsg, ProcessMsg, QueryResponse},
     node::{
         NodeCmd, NodeDataQueryResponse, NodeMsg, NodeQuery, NodeQueryResponse, NodeSystemCmd,
         NodeSystemQuery,
     },
-    DstLocation, MessageId, SrcLocation,
+    ClientAuthority, ClientSigned, DstLocation, MessageId, SrcLocation, WireMsg,
 };
 use crate::node::{
     error::convert_to_error_message,
@@ -79,26 +79,39 @@ fn match_node_msg(msg_id: MessageId, msg: MessageReceived, origin: SrcLocation) 
         MessageReceived::NodeQuery(NodeQuery::Metadata {
             query,
             client_signed,
-            origin,
+            origin: query_origin,
         }) => {
-            // FIXME: ******** validate client signature!!!! *********
-            NodeDuty::ProcessRead {
-                query,
+            match verify_data_authority(
                 msg_id,
-                client_signed,
                 origin,
+                client_signed,
+                ProcessMsg::Query(query.clone()),
+            ) {
+                Ok(client_auth) => NodeDuty::ProcessRead {
+                    query,
+                    msg_id,
+                    client_auth,
+                    origin: query_origin,
+                },
+                Err(duty) => duty,
             }
         }
         MessageReceived::NodeCmd(NodeCmd::Metadata {
             cmd,
             client_signed,
-            origin,
-        }) => NodeDuty::ProcessWrite {
-            cmd,
-            msg_id,
-            client_signed,
-            origin,
-        },
+            origin: cmd_origin,
+        }) => {
+            match verify_data_authority(msg_id, origin, client_signed, ProcessMsg::Cmd(cmd.clone()))
+            {
+                Ok(client_auth) => NodeDuty::ProcessWrite {
+                    cmd,
+                    msg_id,
+                    client_auth,
+                    origin: cmd_origin,
+                },
+                Err(duty) => duty,
+            }
+        }
         //
         // ------ Adult ------
         MessageReceived::NodeQuery(NodeQuery::Chunks { query, .. }) => NodeDuty::ReadChunk {
@@ -107,11 +120,21 @@ fn match_node_msg(msg_id: MessageId, msg: MessageReceived, origin: SrcLocation) 
         },
         MessageReceived::NodeCmd(NodeCmd::Chunks {
             cmd, client_signed, ..
-        }) => NodeDuty::WriteChunk {
-            write: cmd,
-            msg_id,
-            client_signed,
-        },
+        }) => {
+            match verify_data_authority(
+                msg_id,
+                origin,
+                client_signed,
+                ProcessMsg::Cmd(DataCmd::Chunk(cmd.clone())),
+            ) {
+                Ok(client_auth) => NodeDuty::WriteChunk {
+                    write: cmd,
+                    msg_id,
+                    client_auth,
+                },
+                Err(duty) => duty,
+            }
+        }
         // this cmd is accumulated, thus has authority
         MessageReceived::NodeCmd(NodeCmd::System(NodeSystemCmd::ReplicateChunk(chunk))) => {
             NodeDuty::ReplicateChunk { chunk, msg_id }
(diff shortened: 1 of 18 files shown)


src/node/metadata/mod.rs
```diff
@@ -170,3 +170,10 @@ fn build_client_error_response(error: CmdError, msg_id: MessageId, origin: EndUs
         aggregation: false,
     }
 }
+
+// TODO: verify earlier so that this isn't needed
+fn verify_op(client_sig: ClientSig, cmd: DataCmd) -> Result<ClientSigned> {
```
bochaco, line 175: In cases like this, I always try to receive a ref to prevent from unnecessary data cloning, i.e. `client_sig: &ClientSig`

bochaco (commented):
One thing I was considering, and trying to achieve, it is to have the messaging mod/crate to not contain any impl and simply be structs and enums with their documentation of the flow they belong to. Then all type of utilities to instantiate them to be part of other mods, e.g. routing, client, etc. Create a very clear separation of what messages and the protocol is, and what are our own set of utilities and/or impl of the protocol.

This is just a proposal, mainly so we can then publish messaging on its own and can be used not only as a protocol document but also to be used by 3rd party developers implementing their own version of nodes and clients, either in Rust or any other languages.

src/node/event_mapping/node_msg.rs
```diff
@@ -137,21 +160,34 @@ fn match_node_msg(msg_id: MessageId, msg: MessageReceived, origin: SrcLocation)
...
+        ),
     }
 }
+
+fn verify_data_authority(
+    msg_id: MessageId,
+    origin: SrcLocation,
+    client_signed: ClientSigned,
+    msg: ProcessMsg,
+) -> Result<ClientAuthority, NodeDuty> {
```
bochaco, line 177: I don't think this should return a `Result` but simply a NodeDuty ?

src/client/client_api/commands.rs
```diff
@@ -23,30 +22,24 @@ impl Client {
     pub async fn send_signed_command(
         &self,
         cmd: DataCmd,
-        client_pk: PublicKey,
+        client_auth: ClientAuthority,
```
bochaco, line 25: This will be a problem for offline signing, remember this API is intended to support that, and hardware wallets. It seems that won't be possible as the user cannot build a `ClientAuthority` without providing its secret key?

bochaco (changes requested):

src/client/client_api/queries.rs
```diff
@@ -23,32 +22,26 @@ impl Client {
     pub(crate) async fn send_signed_query(
         &self,
         query: DataQuery,
-        client_pk: PublicKey,
+        client_auth: ClientAuthority,
```
bochaco, line 25: Same issue as with the change in `send_signed_command`.

src/client/client_api/commands.rs
bochaco: ....hmmm reviewing it again, i see why it returns a `Result`...so I'll take back my comment

connec: Yeah, this is a mild nuisance with the approach. The crux of the issue is that construction of `ClientAuthority` can't be possible with invalid parts, so the only options are:

- Construct with parts and validate them (`ClientAuthority::verify(...)`) – this would work with a hardware wallet scenario, at the cost of a 'redundant' sig check ("I just generated the sig, why do I have to verify it?").
- Generate on construction, as per `ClientAuthority::new`, but this requires the keypair.

So, when I wrote this I debated between using `ClientSigned` in the API, meaning callers could provide whatever they want even if it's bogus, or using `ClientAuthority`, meaning callers would need to choose how they acquire it. I guess the way I'm imaging using the current API would look like:

```rust
let signature = sign_with_my_wallet(&payload);
let client_auth = ClientAuthority::verify(wallet_pk, signature, &payload).unwrap(); // we just created the sig
send_signed_command(...)
```

All that said, I guess we don't need the signature to be valid here, so yeah we could just use `ClientSigned`.

src/node/event_mapping/node_msg.rs
bochaco: ....hmmm reviewing it again, i see why it returns a Result...so I take back my comment

src/client/client_api/commands.rs
bochaco: This is just a client and its API, so any enforcement made here is only to help the user at most. The actual check will be on the network side, so the client shouldn't be too much concerned about an invalid signature, that's not in our/its control, the user needs to provide a valid signature to get the message accepted by the network. So I think either ClientSigned or just the two parts separate as current code is good enough. 
Having the client to do such verification is something I'd doubt we want, let's let that responsibility to the user of this lib/API.

src/node/event_mapping/node_msg.rs
connec: Hrm, I don't thinks so, since the duty for the `Ok` case depends on the query/cmd? We only know the right `NodeDuty` based on what's received, so we could match again against `msg` but there would only be three cases and an `unimplemented!` or `unreachable!`, e.g.

```rust
fn verify_data_authority(
    msg_id: MessageId,
    origin: SrcLocation,
    msg_origin: Option<EndUser>,
    client_signed: ClientSigned,
    msg: ProcessMsg,
) -> NodeDuty {
    /* ... */
    match msg {
        ProcessMsg::Query(query) => NodeDuty::ProcessRead {
            query,
            msg_id,
            client_auth,
            origin: msg_origin.unwrap(),
        },
        ProcessMsg::Cmd(DataCmd::Chunk(cmd)) => NodeDuty::WriteChunk {
            write: cmd,
            msg_id,
            client_auth,
        },
        ProcessMsg::Cmd(cmd) => NodeDuty::ProcessWrite {
            cmd,
            msg_id,
            client_auth,
            origin: msg_origin.unwrap(),
        },
        _ => unreachable!(),
    }
}
```

Not sure I prefer that, especially for the `unreachable!`. Unless you were thinking of something else?

src/client/client_api/commands.rs
connec: Cool, that makes sense to me, I'll switch this back then.

src/node/event_mapping/node_msg.rs
bochaco: no, you are right, we cross with comments here, i understand the need of a Result there, I definitely wouldn't prefer `unreachable!()`

bochaco (approved):

bochaco (approved):

PR #201, opened 2021-07-19
refactor(routing): simplify anti-entropy logic and remove duplicated AE logic

bochaco:
- Unify `routing::core::messaging::check_lagging` functionality with AE check functionality made on top level NodeMsg's `DstLocation`.
- Do not handle JoinRequests in AE check step, and simply skip AE check so it gets properly handled by the Join handler later on in the flow.


src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -27,205 +19,164 @@ impl Core {
...
-            let cmd = self.send_direct_message(
-                (sender_name, sender),
-                node_msg,
-                section_auth.section_key(),
-            )?;
-
-            Ok((Some(cmd), false))
+        // For the case of receiving a JoinRequest not matching our prefix,
+        // we just let the JoinRequest handler to deal with it later on.
+        if let NodeMsg::JoinRequest(_) = node_msg {
```
maqi, line 34: here there is no longer check of prefix. at least the comment shall be refactored?

bochaco: I can improve it, the comment is still trying to tell that a join msg which is not matching the section/prefix/pk won't be handled here, but I'll try to improve it, it's just so it's clear we are not ignoring those checks but doing them as part of the join handler.

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,428 @@
...
+            Err(_) => {
+                trace!(
+                    "Anti-Entropy: cannot find dst_section_pk {:?} sent by {} in our chain",
+                    dst_section_pk,
+                    sender
+                );
+
+                // Let's try to find a section closer to the destination section key,
+                // otherwise we just drop the message.
+                let name = XorName::from(PublicKey::Bls(*dst_section_pk));
```
maqi, line 217: section_key 's name is different to that section's name, which means closest to the key won't land the msg to that section.
Don't get the point why doing so ?

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,428 @@
...
+                match self.network.closest(&name) {
+                    Some(section_auth)
+                        if &section_auth.value != self.section.authority_provider() =>
```
maqi, line 220: the difference could be within the `key` itself ? why not just check `prefix` ?

src/messaging/node/mod.rs
```diff
@@ -43,6 +43,31 @@ use xor_name::XorName;
...
+        /// Current `SectionAuthorityProvider` of the sender's section.
+        section_auth: SectionAuthorityProvider,
+        /// Sender's section signature over the `SectionAuthorityProvider`.
+        section_signed: KeyedSig,
+        /// Sender's section chain truncated from the dest section key found in the `bounced_msg`.
+        proof_chain: SecuredLinkedList,
+        /// Message bounced due to outdated destination section information.
+        bounced_msg: Box<NodeMsg>,
+    },
+    /// Message sent to a peer when a message needs to be sent to a different
```
maqi, line 59: sounds like it shall be done by the `relay` ?
not clear in which scenario why this `AntiEntropyRedirect` is required.

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,428 @@
...
+                        bounced_msg
+                    );
+                    return Ok(vec![]);
+                }
+            }
+        };
+
+        // We send the msg to the Elder closest to the dest section key,
+        // just to pick one of them in a random but deterministic fashion.
+        let name = XorName::from(PublicKey::Bls(dst_section_pk));
```
maqi, line 123: name of section_key is different to the section_name

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,428 @@
...
+        dst_section_pk: &BlsPublicKey,
+        sender: SocketAddr,
+    ) -> Result<Option<Command>> {
+        if self
+            .section
+            .chain()
+            .cmp_by_position(dst_section_pk, self.section.chain().last_key())
+            != Ordering::Less
+        {
+            // Destination section key matches our current section key
```
maqi, line 186: as long as dst_section_pk doesn't equal to self.section.chain().last_key(), the resulted result will always be Ordering::Less

src/messaging/node/mod.rs
```diff
@@ -43,6 +43,31 @@ use xor_name::XorName;
 #[allow(clippy::large_enum_variant)]
 /// Message sent over the among nodes
 pub enum NodeMsg {
+    /// Message sent to a peer when a message with outdated section
+    /// information was received, attaching the bounced message so
+    /// the peer can resend it with up to date destination information.
+    AntiEntropyRetry {
+        /// Current `SectionAuthorityProvider` of the sender's section.
+        section_auth: SectionAuthorityProvider,
```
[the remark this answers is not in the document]
bochaco: basically because we don't want constructs which are not simple structs or primitive types in our messages, mainly because other implementations of nodes and clients may use different programming languages. Also, having that type of definition doesn't enforce it, validations shall be made anyways when deserialising the message.

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,428 @@
...
+        ) {
+            Ok(updated) => {
+                if updated {
+                    info!(
+                        "Anti-Entropy: updated remote section SAP updated for {:?}",
+                        section_auth.prefix
+                    );
+                } else {
+                    debug!(
+                        "Anti-Entropy: discarded SAP for {:?} since it's the same as the one in our records: {:?}",
```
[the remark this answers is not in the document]
bochaco: no, only when we are not updating due to SAP being the same. Failures return an `Err`

src/routing/core/msg_handling/anti_entropy.rs
bochaco: ah nice catch! since the key we received as the destination pk is not in our chain we simply return a AE-Redirect msg to the sender so it can contact a closer section, so we just try to find a closer section for the destination of the msg. It seem we need to find that out using the DstLocation instead.

src/routing/core/msg_handling/anti_entropy.rs
bochaco: yes, i guess we can simply check the key since any diff in the elders means a new key. The prefix could still be the same though, a section could have same prefix and different elders. Anyhow, I'm now thinking that in theory we don't need this check since we don't keep our own SAP in the `self.network`, so we shouldn't be getting our own SAP when querying from `self.network`.

src/messaging/node/mod.rs
bochaco: we won't be doing relay pr forwarding, at least that's my understanding, so each peer will talk directly to the destination.

src/routing/core/msg_handling/anti_entropy.rs
bochaco: yes, in this case it's just to have a deterministic way of choosing one of the elders where to re-send the msg to, thus I'm using the dst key just for that, the msg will be sent directly

src/routing/core/msg_handling/anti_entropy.rs
bochaco: the dst_section_pk may not be in our chain returning `Ordering::Greater`

bochaco: ...I think it's wrong anyway as we need to handle the case for `Ordering::Greater` with an AE-Redirect msg.

maqi: >the dst_section_pk may not be in our chain returning Ordering::Greater

when a key is not in the chain, the cmp result will be `Ordering::Less`

src/messaging/node/mod.rs
maqi: > so each peer will talk directly to the destination.

Then, why this `AntiEntropyRedirect` is needed?

src/routing/network/mod.rs
```diff
@@ -168,46 +142,54 @@ impl NetworkUtils for Network {
...
+                    )));
                 }
             }
         }
 
+        // Make sure the proof chain can be trusted,
+        // i.e. check each key is signed by its parent/predecesor key.
+        if !proof_chain.self_verify() {
+            return Err(Error::UntrustedProofChain(format!(
+                "invalid signature: {:?}",
```
maqi, line 173: invalid proof_chain?

src/routing/network/mod.rs
```diff
@@ -168,46 +142,54 @@ impl NetworkUtils for Network {
...
+                proof_chain
+            )));
+        }
+
+        // Check the SAP's key is the last key of the proof chain
+        if proof_chain.last_key() != &signed_section_auth.value.public_key_set.public_key() {
```
maqi, line 179: might be an edge case or even a `pointless` attack, 
however, if this `signed_section_auth` is using different section group to sign the `last_key` , this will pass this check however introduce a SAP signed by wrong parent to be recorded?

src/routing/core/msg_handling/anti_entropy.rs
bochaco: ah yes, you are right! I've removed that anyways, and now we compare the keys directly as that's all we care in that check really.

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -8,305 +8,432 @@
...
+                let section_signed = section_signed_auth.sig;
+
+                NodeMsg::AntiEntropyRetry {
+                    section_auth,
+                    section_signed,
+                    proof_chain,
+                    bounced_msg: Box::new(node_msg.clone()),
+                }
+            }
+            Err(_) => {
```
maqi, line 213: I am not sure about this branch.
The `dst_section_pk` is supposed to about our section.
If we cannot find it in our chain, it shall only means the sender has no knowledge of us at all, hence populated with a `random` key.
So, here we only need to send back a chain started from `genesis_key` so that our current SAP can be verified and accepted.
I think finding a closer section to dstination will only loop the msg back to self.
Hence, I don't think AntiEntropyRedirect is necessary at all.

src/messaging/node/mod.rs
bochaco: to return the SAP of a closest section to the sender

src/routing/network/mod.rs
bochaco: i don't think it matter if it's pointless attack or not, we just need to make sure we can trust the SAP and proof-chain. The SAP's key has to be the last one in the chain, and we also check that the chain is connected to our chain or previous SAP, thus we can trust the SAP, that's what this is about. Not sure if I'm missing your point.

src/routing/core/msg_handling/anti_entropy.rs
bochaco: > hence populated with a random key.
we don't know that and cannot assume that. It could be a key which belongs to our chain and we are not aware of yet, it could be a split we are not aware of yet

bochaco: > here we only need to send back a chain started from genesis_key so that our current SAP can be verified and accepted.

perhaps we do this if we cannot find a closer SAP/section

src/routing/network/mod.rs
maqi: What I mean is that the `signer` of the SAP is not the `parent` in the proof_chain.
This will still pass all the checks we have here. 
And we will record this SAP directly, not the correct one from the proof_chain.

Though, I don't get a clear example of such `bad signer` may cause any trouble.
Just point it out to avoid any potential abuse of such.

bochaco: if the SAP's key is the last one in the chain, and the SAP signature is valid, and proof-chain signatures are verified, then it's all good....not sure how the signer of the SAP cannot be a prent in the proof chain if those checks passed.

bochaco: I've added an additional validation which makes sure the SAP's section key matches the key in its signature, I think with this addition we cover the scenario you are describing above.

bochaco:
Closing this as it's been merged to https://github.com/maidsafe/safe_network/tree/Feat-ChunksToRouting branch that will go into `main` branch afterwards.

PR #206, opened 2021-07-20
refactor(types)!: More compact ed25519 signature serialisation

connec:
- 65067d1d5 **refactor(types)!: More compact ed25519 signature serialisation**

  `ed25519_dalek::Signature` serializes as a 64 element tuple, which
  `rmp_serde` serializes as a regular array in which each byte is
  serialized individually. Concretely, this meant any byte larger than
  0x7f/127 would be serialized as *two* bytes (one tagging it as a uint 8,
  and the second for the actual byte).
  
  Message Pack does define a byte array type, which serializes as a tag,
  length, and bytes. `rmp_serde` supports this through serde's `*_bytes`
  methods, which we can use `serde_bytes` to hook into.
  
  BREAKING CHANGE: Although the API doesn't change, signatures are
  serialized in messages, so this break the message format.


src/types/keys/signature.rs
```diff
@@ -36,7 +37,12 @@ pub struct SignatureShare {
 #[allow(clippy::large_enum_variant)]
 pub enum Signature {
     /// Ed25519 signature.
-    Ed25519(#[debug(with = "Self::fmt_ed25519")] ed25519_dalek::Signature),
+    Ed25519(
+        #[debug(with = "Self::fmt_ed25519")]
+        #[serde(deserialize_with = "Signature::deserialize_ed25519")]
```
connec, line 42: `Self` doesn't work here, since internally serde generates a wrapper struct, which `Self` ends up referring to 😢

bochaco (approved):

connec:
I've merged this, but there's some activity in the `ed25519` that could see this fixed or simplified upstream.

PR #207, opened 2021-07-21
fix: retain the original dst_location

maqi:

### DIFF
--- a/src/node/metadata/chunk_records.rs
+++ b/src/node/metadata/chunk_records.rs
@@ -22,6 +22,7 @@ use crate::routing::Prefix;
 use crate::types::{Chunk, ChunkAddress, PublicKey};
 use tracing::{info, warn};
 
+use bls::PublicKey as BlsPublicKey;
 use std::{
     collections::BTreeSet,
     fmt::{self, Display, Formatter},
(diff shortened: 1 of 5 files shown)


src/node/metadata/mod.rs
```diff
@@ -149,6 +155,44 @@ fn build_client_query_response(
...
+        msg: MsgType::Node(NodeMsg::ForwardDataMsg {
+            msg,
+            user: origin,
+            data_signed,
+        }),
+        // TODO: Here shall use the section_pk of the target section.
+        //       But the one passed in is a the section_pk of own section.
+        dst: DstLocation::Section {
+            name: origin.xorname,
+            section_pk,
```
Yoga07, line 190: Is this correct? We should be populating this with the recipient section's PK(i.e. Client Elders) right? Or else, the AE check won't pass.

joshuef (approved):

src/node/metadata/chunk_records.rs
```diff
@@ -186,6 +186,18 @@ impl ChunkRecords {
...
                 } else {
+                    let section_pk_option = if let Ok(section_pk) =
+                        network.get_section_pk_by_name(&end_user.xorname).await
+                    {
+                        section_pk.bls()
+                    } else {
+                        None
+                    };
+                    let section_pk = if let Some(section_pk) = section_pk_option {
+                        section_pk
```
Yoga07, line 197: Can we error out here instead of sending our own key? Because that'd anyhow be rejected at the recipient and be a wasteful round of messaging.

joshuef (approved):

Yoga07 (approved):

PR #210, opened 2021-07-21
Node authority by construction

connec:
- 1f774e29e **refactor(messaging)!: Add an `Authority` 'proof' struct**

  This generalises the approach taken with `ClientAuthority` of wrapping a
  `*Signed` struct in such a way that they can only be constructed
  (outside the `authority` module) with a valid signature for a particular
  payload.
  
  It has the same advantage as `ClientSigned`: an `Authority<NodeSigned>`
  is guaranteed to contain a valid `NodeSigned`. By extension, a
  `NodeSigned` could be assumed to be unverified.
  
  This was much easier to 'plumb in' than `ClientAuthority`, since it's
  used in fewer places. As with `ClientAuthority`, signatures would
  currently be double-verified (once by `WireMsgUtils::check_signature`
  and once again in `WireMsg::into_message`). This can change once all
  `*Signed` types have a corresponding `*Authority`.
  
  BREAKING CHANGE: Some public types that contained `NodeSigned` now
  contain `Authority<NodeSigned>`. Usage may require unwrapping the value
  with `into_inner`.

- e8009c4c2 **refactor(messaging)!: Use `Authority<BlsShareSigned>`**

  We now use `Authority<BlsShareSigned>` wherever we expect a valid
  `BlsShareSigned`. As with `NodeSigned`, this was quite straightforward
  to plumb in.
  
  BREAKING CHANGE: Some public types that contained `BlsShareSigned` now
  contain `Authority<BlsShareSigned>`. Usage may require unwrapping the
  value with `into_inner`, or using the new getters to read fields.

- 1ab2dba95 **refactor(messaging)!: Use `Authority<SectionSigned>`**

  We now use `Authority<SectionSigned>` wherever we expect a valid
  `SectionSigned`. As with `NodeSigned`, this was quite straightforward to
  plumb in.
  
  BREAKING CHANGE: Some public types that contained `SectionSigned` now
  contain `Authority<SectionSigned>`. Usage may require unwrapping the
  value with `into_inner`, or using the new getters to read fields.

- 20983efa4 **refactor(routing): Remove sig check from `Core::handle_message`**

  Signatures are now checked as part of deserialization in
  `WireMsg::into_message`.

- e7e736fca **refactor(messaging)!: Replace `CientAuthority` with `Authority<ClientSigned>`**

  Now that we have a generic `Authority<T>` wrapper, we can get rid of the
  specific `ClientAuthority` type.
  
  BREAKING CHANGE: The `ClientAuthority` type is dead. Long live the
  `Authority<ClientSigned>` type.


src/messaging/authority.rs
```diff
@@ -101,6 +107,23 @@ pub struct NodeSigned {
     pub signature: EdSignature,
 }
 
+impl NodeSigned {
```
bochaco, line 110: Assuming `NodeSigned`, `BlsSharedSigned` and `ClientSigned` contain signature which are not verified, plus they are our message defs, I don't think we need these geters and setters?, they are all public fields and that's fine, there is nothing to protect with encapsulation?

src/messaging/authority.rs
```diff
@@ -122,3 +157,160 @@ pub struct SectionSigned {
...
+}
+
+/// Verified authority.
+///
+/// Values of this type constitute a proof that the signature is valid for a particular payload.
+/// This is made possible by performing verification in all possible constructors of the type.
+///
+/// Validation is defined by the [`VerifyAuthority`] impl for `T`.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Authority<T>(T);
```
bochaco, line 180: I think this is nice, but only if we make it strictly part of our API and not a message definition? then we should probably move it to the same place where we have, or will have, the `MessageType` def and impl?

src/messaging/authority.rs
```diff
@@ -122,3 +157,160 @@ pub struct SectionSigned {
...
+}
+impl sealed::Sealed for BlsShareSigned {}
+
+impl Authority<SectionSigned> {
+    /// Try to construct verified section authority by aggregating a new share.
+    pub(crate) fn try_new(
+        aggregator: &mut SignatureAggregator,
+        share: BlsShareSigned,
+        payload: impl AsRef<[u8]>,
+    ) -> Result<Self, AggregatorError> {
```
bochaco, line 283: I like this, although we are mixing things a bit perhaps, returning an error type of a mod which doesn't belong to this mod....?....I think @Yoga07 mentioned that we are moving the aggregator to be part of messaging API?, if that's correct I guess then this will look better and all fitting in the correct self-contained module, i.e. the messaging API with the authority and verifications utilities, which include the aggregator utility....?

src/messaging/authority.rs
connec: Yup, I didn't try to move anything around yet. I wanted to see what we ended up with here in terms of the API. I'm definitely on board with separating this API stuff from the `*Signed` stuff. I could look at that in this PR, or pursue it as a follow-up.

src/messaging/authority.rs
connec: Yeah, I didn't feel good bringing in the aggregator stuff. I have been feeling that this stuff could easily leave in its own top-level module (or perhaps in routing?) since it's cross-cutting in terms of client/node/routing. A module to keep all of the signing/authority API together could be quite nice (though, I think the `*Signed` types should stay in `messaging` since they're part of the message serialization).

src/messaging/authority.rs
connec: Huh, TIL deref coercion works for fields as well as methods! So yeah, these can be removed 🎉

connec:
I've removed the getters and replaced `ClientAuthority` with `Authority<ClientSigned>`. I've not moved anything around yet. If I was to do so, I'd be tempted to do something like create a `messaging::api` mod and move all our impls and non-serialized types into it, which might be better as a separate PR?

src/messaging/authority.rs
Yoga07: > I think @Yoga07 mentioned that we are moving the aggregator to be part of messaging API

Yup, that's the plan. I've moved it out of `routing` into the `messaging` module in my wip branch. So it's now generic and any layer can use it.

bochaco: I think we need to make sure this is either part of another top level mod, or if it's part of messaging mod it has to be part of the API part, but I guess once @connec starts with the separation of messaging and API it will be more clear where this will fit in. We don't want this to be part of the messaging mod as the aggregator is not a message type but just a utility to aggregate sigs.

bochaco:
@connec it looks like this is ready to merge, just need to rebase it.

connec:
Rebased!

bochaco (approved):

PR #230, opened 2021-07-27
chore(routing): removing OtherSection struct as the SAP is already self-verifiable

bochaco:


maqi (approved):

src/routing/core/msg_handling/agreement.rs
```diff
@@ -230,9 +233,13 @@ impl Core {
             if section_auth.value.prefix.matches(&self.node.name()) {
                 let _ = self.section.update_elders(section_auth, key_sig);
             } else {
-                let _ =
-                    self.network
-                        .update_section(section_auth, Some(key_sig), self.section.chain());
+                // FIXME: we shouln't be receiving or updating a SAP for
```
maqi, line 236: may receive such during split?
I'd agree this is more like an optimisation now after other refactories.
So can consider remove it.

src/routing/core/msg_handling/agreement.rs
```diff
@@ -203,10 +203,13 @@ impl Core {
             );
         } else {
             // Other section
-
-            let _ = self
-                .network
-                .update_section(section_auth, None, self.section.chain());
+            // FIXME: we shouln't be receiving or updating a SAP for
```
maqi, line 206: may receive such during split?
I'd agree this is more like an optimisation now after other refactories.
So can consider remove it.

src/routing/network/mod.rs
```diff
@@ -142,82 +132,98 @@ impl NetworkUtils for Network {
...
+        // Check if SAP signature is valid
+        if !signed_section_auth.self_verify() {
+            trace!(
+                "Failed to update remote section knowledge, SAP signature invalid: {:?}",
+                signed_section_auth.value
+            );
+            return false;
+        }
+
+        // TODO: Let's make sure the proof chain can be trusted,
```
maqi, line 152: SecuredLinkedList itself shall provide an function like `self_verified` to carry out such check?

src/routing/network/mod.rs
```diff
@@ -142,82 +132,98 @@ impl NetworkUtils for Network {
...
-                section_auth.value
+                "Failed to update remote section knowledge, SAP's key ({:?}) doesn't match proof chain last key ({:?})",
+                 signed_section_auth.value.public_key_set.public_key(), proof_chain.last_key()
             );
             return false;
         }
 
-        if let Some(old) = self.sections.insert(info) {
-            if old.section_auth == section_auth {
+        // We currently don't keep the complete chain of remote sections (TODO??),
```
maqi, line 164: we shall never do that (keep a complete chain of remote section) ? you can remove the `TODO` I think.

src/routing/network/mod.rs
```diff
@@ -142,82 +132,98 @@ impl NetworkUtils for Network {
...
+                // We are not aware of the prefix, let's then verify
+                // it can be trusted based on our own section chain and the
+                // provided proof chain.
+                our_section_chain.check_trust(proof_chain.keys())
+            }
+        };
+
+        if !is_sap_trusted {
+            trace!(
+                "Failed to update remote section knowledge, SAP cannot be trusted: {:?}",
```
maqi, line 190: better to log the detailed failing reason, is that our chain cannot trust their chain, or their chain doesn't have the claimed SAP.

src/routing/network/mod.rs
```diff
@@ -142,82 +132,98 @@ impl NetworkUtils for Network {
...
-            if old.section_auth == section_auth {
+        // We currently don't keep the complete chain of remote sections (TODO??),
+        // **but** the SAPs of remote sections we keep were already verified by us
+        // as trusted before we store them in our local records.
+        // Thus, we just need to check our knowledge of the remote section's key
+        // is part of the proof chain received.
+        let is_sap_trusted = match self.sections.get(&signed_section_auth.value.prefix) {
+            Some(sap) if sap == &signed_section_auth => {
+                // It's the same SAP we are already aware of
                 return false;
```
maqi, line 172: we had some `false alerts` regarding this false value returned for `no-update` , maybe can consider change it to `true` ?

src/routing/network/mod.rs
```diff
@@ -142,82 +132,98 @@ impl NetworkUtils for Network {
...
+        if !is_sap_trusted {
+            trace!(
+                "Failed to update remote section knowledge, SAP cannot be trusted: {:?}",
+                signed_section_auth.value
+            );
+            return false;
         }
 
+        // We can now update our knowledge of the remote section's SAP.
+        // Note: we don't expect the the same SAP to be found in our records
```
maqi, line 197: `we don't expect the same SAP`

src/routing/network/mod.rs
bochaco: we may actually need a diff return type, I think we may need to know if it was updated or if was not needed.

src/routing/network/mod.rs
bochaco: yup

src/routing/core/msg_handling/agreement.rs
bochaco: I'm keeping it for now just to not break DKG and/or splits, once we are fully AE and no more Sync msgs, this may go away, I'm keeping it for later fix/review.

joshuef (approved):

PR #231, opened 2021-07-27
chore(routing): Remove message filtering

connec:
- 828eaf21f **chore(routing): Remove message filtering**

  The message filtering is part of a system intended to ensure
  exactly-once delivery by filtering out duplicate incoming (and
  outgoing!) messages. We now want to instead ensure our messages are
  handled idempotently, in which case at-least-once delivery is enough.
  That means we can remove the message filtering, and later remove any use
  deterministic message IDs.

- c3c6dbecb **chore: Update qp2p to get more connection retries**

  Hopefully this will make the tests a bit less flaky, though the jury's
  still out.


src/routing/core/api.rs
```diff
@@ -171,37 +171,16 @@ impl Core {
...
-                .msg_filter
-                .filter_outgoing(&wire_msg, peer.name())
-                .await
-                .is_new()
-            {
-                let _ = targets.push((*peer.name(), *peer.addr()));
-            }
-        }
-
-        if targets.is_empty() {
```
bochaco, line 201: Perhaps we still want this check if for some reason we don't find any targets for dst_location? just to prevent from then getting errors in the log saying the recipients list is empty.

connec: Hrm, fair point. I removed it on the assumption it would catch the case where all `presumed_targets` were filtered out. I'll double check `delivery_group::delivery_targets` and re-add it tomorrow if it makes sense.

connec: Yes, so `delivery_group::delivery_targets` can definitely return an empty vec, I'll restore the `is_empty` check.

connec:
Closing in favour of #298.

PR #234, opened 2021-07-27
Simple™️ serialization

connec:
- 06e116f73 **test(messaging): Add a serialization test for `WireMsgHeader`**

  This will facilitate refactoring, without breaking the format
  (unintentionally at least).

- c55923449 **refactor(messaging)!: Move version before length in `WireMsgHeader`**

  This is a step towards a (hopefully) substantial simplification of
  message serialization by using bincode. By putting the length before the
  serialized message envelope, we should be able to derive deserialization
  for the overall `(version, len, header bytes)` structure.
  
  BREAKING CHANGE: This breaks the wire message format – messages from
  earlier versions cannot be deserialized by this version, and vice-versa.

- 3b765ebc6 **refactor(messaging)!: Make header size exclusive of version and size**

  This is another step towards simplifying message serialization with
  bincode. This brings us closer to the `(size, bytes)` portion of the
  message being directly serializable as `&[u8]`, since `size` needs to be
  the size of bytes, rather than the whole container.
  
  The `WireMsgHeader::write` method still returns the overall serialized
  size, so the interface with `WireMsg` doesn't need to change.
  
  BREAKING CHANGE: This breaks the wire message format. Messages from
  earlier versions will not be readable in this version (and vice-versa).

- 200b13471 **refactor(messaging)!: Use bincode for header version and size**

  This changes how we serialise the version and header size for wire
  message headers. Previously we used fixed-size big-endian encoded
  `u16`s. We now use bincode's variable length encoding, which is another
  step towards deserializing the whole thing as bincode.
  
  bincode does support fixed-length encoding, but it would mean
  serialising the header size as `u64`, so always 8 bytes. For our
  `u16`-sized headers that would be wasting 5-7 bytes per message, vs. a
  single extra byte for larger headers (≥251 bytes) when using
  variable-length.
  
  BREAKING CHANGE: This breaks the wire message format. Messages from
  earlier versions will be unreadable from this version, and vice-versa.

- 28863e87f **refactor(messaging): Derive most of `WireMsgHeader` serialization**

  With the previous changes, it's now possible to (de)serialize the
  message header as a struct containing the version and the payload bytes
  (which would be encoded with the length as a prefix). This has been
  called `MessageSkeleton`, for want of a better name, but since it's
  private to the module it's not too important.
  
  The unchanged test shows that this is compatible with the previously
  hand-written serialization code.

- 122865a7f **refactor(messaging): Remove `u16` limit from header size**

  This probably existed in order to have a fixed-size encoding, and `u16`
  is a pretty sensible limit for a header size. However, since we can now
  support arbitrary `usize` through bincode, and since message size limits
  would be better handled earlier in the stack, we can simplify things by
  removing this limit.

### DIFF
--- a/src/messaging/serialisation/wire_msg_header.rs
+++ b/src/messaging/serialisation/wire_msg_header.rs
@@ -10,7 +10,7 @@ use crate::messaging::{DstLocation, Error, MessageId, MsgKind, Result};
 use bytes::Bytes;
 use cookie_factory::{bytes::be_u16, combinator::slice, gen, gen_simple};
 use serde::{Deserialize, Serialize};
-use std::mem::size_of;
+use std::{convert::TryFrom, mem::size_of};
 
 // Current version of the messaging protocol.
 // At this point this implementation supports only this version.
@@ -99,7 +99,7 @@ impl WireMsgHeader {
 
         // We check that at least we have the minimum number of bytes
         // for the header of any kind of message to be deserialised.
-        if length < header_size.into() {
+        if length < Self::serialized_size(header_size).into() {
             return Err(Error::FailedToParse(format!(
                 "not enough bytes received ({}) to deserialize wire message header",
                 length
@@ -107,7 +107,8 @@ impl WireMsgHeader {
         }
 
         // ...finally, we read the message envelope bytes
-        let msg_envelope_bytes = &bytes[HDR_SIZE_BYTES_END..header_size.into()];
+        let msg_envelope_bytes =
+            &bytes[HDR_SIZE_BYTES_END..Self::serialized_size(header_size).into()];
         let msg_envelope: MsgEnvelope =
             rmp_serde::from_slice(msg_envelope_bytes).map_err(|err| {
                 Error::FailedToParse(format!(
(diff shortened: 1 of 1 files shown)


src/messaging/serialisation/wire_msg_header.rs
```diff
@@ -128,43 +132,63 @@ impl WireMsgHeader {
...
+        let skeleton = HeaderSkeleton {
+            version: self.version,
+            header_bytes: &msg_envelope_vec,
+        };
+        BINCODE_OPTS
+            .serialize_into(&mut buffer, &skeleton)
+            .map_err(|err| Error::Serialisation(format!("failed to serialize header: {}", err)))?;
 
-        Ok((buf_at_payload, header_size))
+        let serialized_size = u16::try_from(buffer_len - buffer.len()).unwrap();
```
lionel1704, line 145: Just flagging the `.unwrap()` here :eyes:

connec: Yeah, I should tighten that up. Since the size limit was taken away, we can't rely on this holding (used to be that `serialized_size` *must* fit `u16`, now it's `usize`).

connec:
Some of this is now a bit egregious vs. #293, which achieves roughly the same thing far more simply and with no breakage. I'll hold on to the branch though, as I like the `MessageSkeleton` idea still.

PR #236, opened 2021-07-28
Use reg store in routing

joshuef:

### DIFF
--- a/examples/routing_minimal.rs
+++ b/examples/routing_minimal.rs
@@ -35,7 +35,9 @@
 
 use anyhow::Result;
 use futures::future::join_all;
-use safe_network::routing::{Config, Event, EventStream, Routing, TransportConfig};
+use safe_network::routing::{
+    create_test_register_store, Config, Event, EventStream, Routing, TransportConfig,
+};
 use std::{
     collections::HashSet,
     convert::TryInto,
@@ -43,18 +45,9 @@ use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr},
 };
 use structopt::StructOpt;
-use tempfile::tempdir;
 use tokio::task::JoinHandle;
 use tracing::info;
 use tracing_subscriber::EnvFilter;
-
-use safe_network::node::RegisterStorage;
-use safe_network::UsedSpace;
-use std::path::Path;
-
-use rand::{distributions::Alphanumeric, Rng};
-
-const TEST_MAX_CAPACITY: u64 = 1024 * 1024;
 
 /// Minimal example node.
 #[derive(Debug, StructOpt)]
@@ -203,16 +196,7 @@ async fn start_node(
         ..Default::default()
     };
 
-    let used_space = UsedSpace::new(TEST_MAX_CAPACITY);
-    let tmp_dir = tempdir()?;
-
-    let register: String = rand::thread_rng()
-        .sample_iter(&Alphanumeric)
-        .take(7)
-        .map(char::from)
-        .collect();
-    let storage_dir = tmp_dir.into_path().join(Path::new(&register));
-    let storage = RegisterStorage::new(&storage_dir, used_space)?;
+    let storage = create_test_register_store()?;
 
     let (node, event_stream) = Routing::new(config, storage)
         .await
--- a/examples/routing_minimal.rs
+++ b/examples/routing_minimal.rs
@@ -36,7 +36,7 @@
 use anyhow::Result;
 use futures::future::join_all;
 use safe_network::routing::{
-    create_test_register_store, Config, Event, EventStream, Routing, TransportConfig,
+    create_test_used_space_and_root_storage, Config, Event, EventStream, Routing, TransportConfig,
 };
 use std::{
     collections::HashSet,
@@ -196,9 +196,9 @@ async fn start_node(
         ..Default::default()
     };
 
-    let storage = create_test_register_store()?;
+    let (used_space, root) = create_test_used_space_and_root_storage()?;
 
-    let (node, event_stream) = Routing::new(config, storage)
+    let (node, event_stream) = Routing::new(config, used_space, root)
         .await
         .expect("Failed to instantiate a Node");
 
--- a/examples/routing_minimal.rs
+++ b/examples/routing_minimal.rs
@@ -33,7 +33,9 @@
 //! command-line option for more details.
 //!
 
+use anyhow::Result;
 use futures::future::join_all;
+use safe_network::routing::{Config, Event, EventStream, Routing, TransportConfig};
 use std::{
     collections::HashSet,
     convert::TryInto,
@@ -41,11 +43,18 @@ use std::{
     net::{IpAddr, Ipv4Addr, SocketAddr},
 };
 use structopt::StructOpt;
+use tempfile::tempdir;
 use tokio::task::JoinHandle;
 use tracing::info;
 use tracing_subscriber::EnvFilter;
 
-use safe_network::routing::{Config, Event, EventStream, Routing, TransportConfig};
+use safe_network::node::RegisterStorage;
+use safe_network::UsedSpace;
+use std::path::Path;
+
+use rand::{distributions::Alphanumeric, Rng};
+
+const TEST_MAX_CAPACITY: u64 = 1024 * 1024;
 
 /// Minimal example node.
 #[derive(Debug, StructOpt)]
@@ -97,13 +106,13 @@ struct Options {
 }
 
 #[tokio::main]
-async fn main() {
+async fn main() -> Result<()> {
     let opts = Options::from_args();
     init_log(opts.verbosity);
 
     let handles: Vec<_> = if opts.count <= 1 {
         let handle =
-            start_single_node(opts.first, opts.bootstrap_contacts, opts.ip, opts.port).await;
+            start_single_node(opts.first, opts.bootstrap_contacts, opts.ip, opts.port).await?;
         iter::once(handle).collect()
     } else {
         start_multiple_nodes(
@@ -113,9 +122,11 @@ async fn main() {
             opts.ip,
             opts.port,
         )
-        .await
+        .await?
     };
     let _ = join_all(handles).await;
+
+    Ok(())
 }
 
 // Starts a single node and block until it terminates.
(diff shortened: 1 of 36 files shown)


src/routing/core/msg_handling/mod.rs
```diff
@@ -205,8 +210,15 @@ impl Core {
                 let payload = WireMsg::serialize_msg_payload(&msg)?;
                 let data_auth = Authority::verify(data_signed, &payload)?;
 
-                self.handle_client_msg_received(msg_id, msg, user, data_auth)
-                    .await
+                // self.handle_data_msg_received(msg_id, msg, user, data_auth)
```
maqi, line 213: this line can now be removed?

src/routing/routing_api/command.rs
```diff
@@ -27,6 +28,14 @@ pub(crate) enum Command {
...
         wire_msg: WireMsg,
     },
+    /// Handle DataMsg, either directly or notify via event listener
+    HandleDataMessage {
+        msg_id: MessageId,
+        msg: DataMsg,
+        user: EndUser,
+        data_auth: Authority<DataSigned>,
+    },
+    // handle_data_msg_received(msg_id, msg, user, data_auth)
```
maqi, line 38: removed?

src/routing/routing_api/mod.rs
```diff
@@ -370,45 +374,49 @@ impl Routing {
...
-        }
-
-        let command = Command::RelayMessage(wire_msg);
-        self.dispatcher.clone().handle_commands(command).await
+    pub async fn send_message(&self, wire_msg: WireMsg) -> Result<()> {
+        self.dispatcher
+            .clone()
+            .handle_commands(Command::ParseAndSendWireMsg(wire_msg))
+            .await
+        // if let DstLocation::EndUser(EndUser { socket_id, xorname }) = wire_msg.dst_location() {
```
maqi, line 382: remvoe ?

src/routing/routing_api/tests/mod.rs
```diff
@@ -39,25 +42,29 @@ use crate::routing::{
     MIN_ADULT_AGE, MIN_AGE,
 };
 use crate::types::{Keypair, PublicKey};
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use assert_matches::assert_matches;
 use ed25519_dalek::Signer;
 use rand::rngs::OsRng;
+use rand::{distributions::Alphanumeric, Rng}; // 0.8
```
maqi, line 49: what does this `0.8` mean?

src/routing/routing_api/tests/mod.rs
```diff
@@ -310,6 +321,19 @@ async fn receive_join_request_with_resource_proof_response() -> Result<()> {
     Ok(())
 }
 
+fn create_test_register_store() -> Result<RegisterStorage> {
```
maqi, line 324: consider make this function and that TEST_MAX_CAPACITY const in a seperate util mod ?
I see they also got used in examples for two times.

src/node/mod.rs
```diff
@@ -21,6 +21,10 @@ mod network;
 mod node_api;
 mod node_ops;
 
+/// TODO: make these intrnal only once routing/node are collapsed
```
maqi, line 24: internal ?

src/node/node_api/mod.rs
```diff
@@ -70,6 +71,9 @@ pub struct Node {
     network_api: Network,
     node_info: NodeInfo,
     used_space: UsedSpace,
+    // We instantiate this even if adult. It just won't be used, but makes it easier keep once storage instance going on.
```
maqi, line 74: you mean `makes it easier by just carry the storage instance on when promoted` ?

examples/routing_minimal.rs
```diff
@@ -191,7 +202,19 @@ async fn start_node(
         transport_config,
         ..Default::default()
     };
-    let (node, event_stream) = Routing::new(config)
+
+    let used_space = UsedSpace::new(TEST_MAX_CAPACITY);
```
maqi, line 206: can this now use the new util function ?

src/routing/core/api.rs
```diff
@@ -52,6 +55,7 @@ impl Core {
             new_section,
             None,
             self.event_tx.clone(),
+            self.register_storage.clone(),
```
maqi, line 58: this shall actually using a new register_storage ?
I suppose this `new_node` shall use a new register_storage already ?

src/routing/core/msg_handling/end_user.rs
```diff
@@ -19,22 +20,49 @@ use bytes::Bytes;
...
-        self.send_event(Event::DataMsgReceived {
-            msg_id,
-            msg: Box::new(msg),
-            user,
-            data_auth,
-        })
-        .await;
-
-        Ok(vec![])
+        if let DataMsg::Process(ProcessMsg::Cmd(DataCmd::Register(register_cmd))) = msg {
```
maqi, line 30: So register_storage is only care about the DataCmd::Register ?
if so, this means just a small part of the DataMsg, then the naming of `Command::ParseAndSendWireMsg` better to be more specific ?

src/routing/core/msg_handling/end_user.rs
```diff
@@ -19,22 +20,49 @@ use bytes::Bytes;
...
+
+                    let dst = DstLocation::EndUser(user);
+
+                    // FIXME: define which signature/authority this message should really carry,
+                    // perhaps it needs to carry Node signature on a NodeMsg::QueryResponse msg type.
+                    // Giving a random sig temporarily
+                    let (msg_kind, payload) = Self::random_client_signature(&the_error_msg)?;
+                    let wire_msg = WireMsg::new_msg(MessageId::new(), payload, msg_kind, dst)?;
+
+                    let command = Command::ParseAndSendWireMsg(wire_msg);
```
maqi, line 50: you shall be able to `send` the error msg directly ? why still need a Parse ?

src/routing/core/msg_handling/end_user.rs
```diff
@@ -19,22 +20,49 @@ use bytes::Bytes;
...
+                        error: CmdError::Data(sending_error),
+                        correlation_id: msg_id,
+                    });
+
+                    let dst = DstLocation::EndUser(user);
+
+                    // FIXME: define which signature/authority this message should really carry,
+                    // perhaps it needs to carry Node signature on a NodeMsg::QueryResponse msg type.
+                    // Giving a random sig temporarily
+                    let (msg_kind, payload) = Self::random_client_signature(&the_error_msg)?;
```
maqi, line 47: there is `build_client_error_response` , can you use that instead ?

src/routing/mod.rs
```diff
@@ -71,6 +79,22 @@ pub const RECOMMENDED_SECTION_SIZE: usize = 2 * ELDER_SIZE;
 /// Number of elders per section.
 pub const ELDER_SIZE: usize = 7;
 
+const TEST_MAX_CAPACITY: u64 = 1024 * 1024;
+
+/// Create a register store for routing examples
+pub fn create_test_register_store() -> AnyhowResult<RegisterStorage> {
+    let used_space = UsedSpace::new(TEST_MAX_CAPACITY);
+    let tmp_dir = tempdir()?;
```
maqi, line 87: just to confirm that the `tempdir` will got the folder removed after the destruction of the instance.
seems the lifespan of the `temp_dir` is just within this function ?

src/routing/core/msg_handling/end_user.rs
joshuef: This is to abstract it. We use this in a couple of places so we don't have that same logic each time. 

Also puts that logic into a new thread.

src/routing/core/msg_handling/end_user.rs
joshuef: For now. To keep this PR small. 

But that send function should likely be used all over. It's basically the `send_message` API that was exposed to node

src/routing/core/api.rs
joshuef: I'm confused as to what you're asking here.

The intention is that each `Node` has one `RegisterStorage` instance, which is currently passed into routing and into node code and accessed in both.

The long term view is that there would be only one needed. But I'm doing this in steps.

src/node/mod.rs
joshuef: They do not need to be exposed but for the routing examples.

When we have node/routing collapsed into one module, they won't need to be exposed

examples/routing_minimal.rs
joshuef: ah missed this!

src/routing/core/msg_handling/end_user.rs
joshuef: that returns some Node specific things. Eventually we'll coalesce into one impl. This is just the first step on the merging of routing/node.

src/routing/mod.rs
joshuef: tempdir just returns the systems `$TMP_DIR` directory path as I understand. Not managing the folder/filesystem at all

src/node/mod.rs
maqi: sry, I mean the word of `intrnal`, is it a typo for `internal` or you mean `intranal` or something else?

src/routing/core/api.rs
maqi: Yeah, as you said `each Node has one RegisterStorage instance` .
Here is within `relocation`, where the new_node is a new_instance. 
However, seems the `RegisterStorage` is `cloned` from previous ?

src/routing/core/msg_handling/end_user.rs
maqi: Ok, then.

src/routing/mod.rs
maqi: https://docs.rs/tempfile/3.2.0/tempfile/fn.tempdir.html

it says `The tempdir function creates a directory in the file system and returns a TempDir. The directory will be automatically deleted when the TempDirs destructor is run`

src/routing/core/api.rs
joshuef: ahhh, sorry, i see. yep yep, you're right

src/routing/mod.rs
joshuef: Righto. Hmmm, I wonder if we're using this wrong throughout. Will have a look at this

joshuef: Well, actually it doesnt matter if the dir is deleted on exit of the func, as it's path remains and that will be created as needed by the internal funcs. 

it'd only be an issue if it was droppeed while that was happening but this would be before we get there

maqi (dismissed):

lionel1704 (dismissed):

lionel1704 (approved):

PR #243, opened 2021-07-29
fix: remove the backlog cache for bootstrapping node

maqi:

### DIFF
--- a/src/routing/core/bootstrap/join.rs
+++ b/src/routing/core/bootstrap/join.rs
@@ -20,7 +20,6 @@ use crate::routing::{
     messages::{NodeMsgAuthorityUtils, WireMsgUtils},
     node::Node,
     peer::PeerUtils,
-    routing_api::command::Command,
     section::SectionUtils,
     SectionAuthorityProviderUtils, FIRST_SECTION_MAX_AGE, FIRST_SECTION_MIN_AGE, MIN_ADULT_AGE,
 };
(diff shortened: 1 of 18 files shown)


src/routing/core/bootstrap/join.rs
```diff
@@ -312,45 +302,30 @@ impl<'a> Join<'a> {
                     Ok(wire_msg) => match wire_msg.msg_kind() {
                         MsgKind::DataMsg(_) | MsgKind::SectionInfoMsg => continue,
                         MsgKind::NodeBlsShareSignedMsg(_) | MsgKind::SectionSignedMsg(_) => {
-                            self.backlog_message(Command::HandleMessage { sender, wire_msg });
+                            trace!(
+                                "Previously backlogged: sender: {:?} wire_msg: {:?}",
```
joshuef, line 306: I'm not sure this message is super helpful. What are we wanting to indicate with this going forward?

maqi: this shall be very rare to happen (according to my local tests)
and in case of happening, just want to see what msg got discarded for any later on investigating.

joshuef: okay, but is "Previously backlogged" helpful info there I mean? Maybe something like "Bootstrap message discarded..."

connec (approved):

bochaco (approved):

PR #247, opened 2021-07-29
Clients multi-section and Anti-Entropy

Yoga07:


src/client/connections/messaging.rs
```diff
@@ -207,37 +222,43 @@ impl Session {
...
 
-        // We select the NUM_OF_ELDERS_SUBSET_FOR_QUERIES closest
-        // connected Elders to the data we are querying
-        let elders: Vec<SocketAddr> = self
-            .connected_elders
-            .read()
-            .await
-            .clone()
+        // Wait for 2 seconds to receive SAP
+        let _ = sleep(Duration::from_secs(2));
```
bochaco, line 230: I guess this is just temporary as part of draft?

src/messaging/section_info/mod.rs
```diff
@@ -7,16 +7,21 @@
...
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 pub enum SectionInfoMsg {
     /// SectionInfoMsg to request information about the section that matches the given name.
-    GetSectionQuery(PublicKey),
+    GetSectionQuery {
+        /// XorName for which we need SectionAuthorityProvider
+        name: XorName,
+        /// Fetching section details for Client bootstrapping
+        is_bootstrapping: bool,
```
bochaco, line 23: You don't need this, what any client would instead receive if it's not the correct section, it's a `GetSectionResponse::Redirect`, I think you don't need to change this message definition, just handle the redirect from the client till it reached the correct destination, which you will need to do because a section may not know the correct destination section but just a closest one. 
Then eventually once we have AE in place, you won't even need to send a `GetSectionQuery` message, just the actual message and the network may reply with AntiEntropyRedirect msg giving the SAP of a closer section where to retry, so same flow but without sending a `GetSectionQuery` beforehand.

src/client/connections/messaging.rs
Yoga07: Yeah, once we have AE baked into client (that's what I am on to next), we don't need to GetSection everytime, so we remove one round of messaging and remove this sleep too :)

Yoga07:
Superseded by #317

PR #249, opened 2021-07-29
refactor(msgs): rename datamsg to servicemsg

oetyng:
The msg type encompasses more than data; in fact all the exposed network services.
Therefore, the name `ServiceMsg` is a better fit.
It also collapses former `DataMsg` and `ProcessMsg` into this one `ServiceMsg`, thereby removing one layer.
Also replaces `Signed` with `Auth`, to denote authority, since that's
what it is described to be in the docs.
`Authority` is then renamed to `AuthorityProof`, since in the docs it is said 
that `Values of this type constitute a proof that the signature is valid for 
a particular payload `.

### DIFF
--- a/src/node/error.rs
+++ b/src/node/error.rs
@@ -7,7 +7,7 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::dbs;
-use crate::messaging::{data::Error as ErrorMessage, MessageId, WireMsg};
+use crate::messaging::{data::Error as ErrorMessage, MessageId};
 use crate::routing::Prefix;
 use crate::types::{convert_dt_error_to_error_message, DataAddress, PublicKey};
 use std::io;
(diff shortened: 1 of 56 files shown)


src/node/error.rs
```diff
@@ -37,12 +37,12 @@ pub enum Error {
...
-    #[error("No source message for ProcessingError")]
-    NoSourceMessageForProcessingError,
-    /// Unexpected Process msg. A ProcessingError was expected here...
-    #[error("Unexpected Process msg. A ProcessingError was expected here...")]
-    UnexpectedProcessMsg,
+    /// No source message provided for ServiceError
+    #[error("No source message for ServiceError")]
+    NoSourceMessageForServiceError,
+    /// Unexpected Process msg. A ServiceError was expected here...
+    #[error("Unexpected Process msg. A ServiceError was expected here...")]
```
connec, line 44:
```suggestion
    #[error("Unexpected service msg. A ServiceError was expected here...")]
```

connec (commented):
LGTM, with one nitpicky suggestion.

#252 may be needed for this to get through CI.

joshuef (approved):

PR #259, opened 2021-07-30
VersionHash replacing the u64 versions

grumbach:
Update versions with VersionHash instead of u64
VersionHash is needed by the new NRS and file containers in `sn_api`

### DIFF
--- a/src/url/mod.rs
+++ b/src/url/mod.rs
@@ -11,6 +11,7 @@
 
 mod errors;
 mod url_parts;
+mod version_hash;
 mod xorurl_media_types;
 
 use crate::types::register;
@@ -21,6 +22,7 @@ use std::fmt;
 use tracing::{info, trace, warn};
 use url::Url;
 use url_parts::NativeUrlParts;
+pub use version_hash::VersionHash;
 use xor_name::{XorName, XOR_NAME_LEN};
 use xorurl_media_types::{MEDIA_TYPE_CODES, MEDIA_TYPE_STR};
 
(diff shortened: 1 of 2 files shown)


src/url/version_hash.rs
```diff
@@ -0,0 +1,105 @@
...
+    #[error("Invalid hash length")]
+    InvalidHashLength,
+    #[error("Invalid encoding (must be Base32Z)")]
+    InvalidEncoding,
+}
+
+/// Version Hash corresponding to the register entry hash where the content is stored
+#[derive(Debug, PartialEq, Serialize, Deserialize, Clone, Default, Copy)]
+pub struct VersionHash {
+    register_entry_hash: EntryHash,
```
bochaco, line 31: Perhaps make this generic and not necessarily for a `Register` entry, e.g. `entry_hash`, I hope eventually we have other native content types which work also with hashes and we can use `VersionHash` in the URLs in the same way.

src/url/mod.rs
```diff
@@ -1725,8 +1729,8 @@ mod tests {
     #[test]
     fn test_safeurl_parts() -> Result<()> {
         // These two are equivalent.  ie, the xorurl is the result of nrs.to_xorurl_string()
-        let nrsurl = "safe://my.sub.domain/path/my%20dir/my%20file.txt?this=that&this=other&color=blue&v=5&name=John+Doe#somefragment";
-        let xorurl = "safe://my.sub.hnyydyiixsfrqix9aoqg97jebuzc6748uc8rykhdd5hjrtg5o4xso9jmggbqh/path/my%20dir/my%20file.txt?this=that&this=other&color=blue&v=5&name=John+Doe#somefragment";
```
bochaco, line 1729: Perhaps use some default/random hash rather than removing it from the test, I believe this test as well as `test_safeurl_to_string` and `test_safeurl_validate` are aimed at making sure we are parsing/encoding/decoding all parts without missing any part, so removing the version/hash from them wouldn't be desirable?

bochaco (changes requested):
It looks good to me, I left only a couple of minor things.

bochaco:
@grumbach I just realised we'll need a commit marked as breaking change.

bochaco (approved):

bochaco (approved):

connec (approved):

PR #269, opened 2021-08-03
refactor(client): remove EndUserRegistry and use latest qp2p which maintains connection mappings for us

lionel1704:
- when a connection is created qp2p will maintain this and it gives us
an ID using which this connection can be referenced. Nodes can respond
back to clients using this connection ID


lionel1704:
These changes will go atop #247

PR #272, opened 2021-08-03
Switch from `anyhow` to `eyre`

connec:
- 50b031240 **fix: Restore `criterion` dependency**

  This was removed in 62998935f, but is required for benchmarks (currently
  only `benches/put.rs`).

- 7e6036264 **refactor: Put routing test utils behind `testing` feature**

  With the merge of crates, we can use the same feature used for `client`
  test utils for routing's ones as well.
  
  This also removes `anyhow` from the library API (it still can't be a dev
  dependency since the `sn_node` binary also depends on it).

- e94adee3c **refactor!: Rename `testing` feature to `test-utils`**

  This is a better name for the purpose (including test utility functions
  in the build).
  
  BREAKING CHANGE: The `testing` feature has been renamed to `test-utils`.

- 0ce05dde7 **refactor!: Switch from `anyhow` to `eyre`**

  This is *almost* the smallest change to get things compiling.
  Technically, `eyre` also re-exports its `eyre!` macro as `anyhow!`, but
  I deemed it confusing to leave that in.
  
  In future we might also want to move away from `Error` to `Report` and
  from `Context` to `WrapErr` to better align with `eyre`'s idioms.
  
  BREAKING CHANGE: Functions exposed with the `test-utils` feature enabled
  now return `eyre::Report` rather than `anyhow::Error`.

- 65726b7d5 **chore: Update `bls_dkg`**

  This fully removes `anyhow` from our dependencies 🎉

- d45f78903 **feat: Use `color-eyre` in `sn_node` for pretty errors**

  This doesn't actually do anything for now since `run_node` swallows all
  errors (apart from an error if the verbosity handling code is buggy,
  which it currently isn't...).

- 03d740492 **refactor(sn_node): Make `run_node` infallible**

  This is a better reflection of the *current* logic (though not what we
  want eventually). The only way `run_node` would return an error is if
  our verbosity handling was buggy, which is a good case for `expect`.

### DIFF
--- a/src/bin/sn_node.rs
+++ b/src/bin/sn_node.rs
@@ -27,7 +27,7 @@
     unused_results
 )]
 
-use eyre::{eyre, Result};
+use eyre::{eyre, Result, WrapErr};
 use safe_network::node::{add_connection_info, set_connection_info, Config, Error, Node};
 use self_update::{cargo_crate_version, Status};
 use std::{io::Write, process};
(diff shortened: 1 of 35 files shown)


src/bin/sn_node.rs
```diff
@@ -96,13 +99,11 @@ async fn run_node() -> Result<()> {
...
-
-            EnvFilter::from_default_env().add_directive(
-                module_log_filter
-                    .parse()
-                    .map_err(|_| anyhow!("Could not parse module log filter"))?,
-            )
+            let module_filter = format!("{}={}", MODULE_NAME, level_filter)
+                .parse()
+                // parsing will fail only if `MODULE_NAME` or `LevelFilter::fmt` is broken
+                .expect("BUG: invalid module filter constructed");
```
connec, line 105: ☣️ This is here temporarily to rationalise the `run_node` function, from where it will be refactored to hopefully behave infallibly.

joshuef (dismissed):

src/bin/sn_node.rs
connec: Fixed.

joshuef (approved):

PR #282, opened 2021-08-04
refactor(url): simplify data type, impl scope

oetyng:
Before, data types were duplicate, as there was one for private, and one
for public.
Now we split out the scope into it's own enum. This makes the structure
more sane.
BREAKING CHANGE: This changes the structure of the NativeUrl.

### DIFF
--- a/src/url/mod.rs
+++ b/src/url/mod.rs
@@ -34,7 +34,7 @@ const SAFE_URL_PROTOCOL: &str = "safe://";
 const SAFE_URL_SCHEME: &str = "safe";
 const XOR_URL_VERSION_1: u64 = 0x1; // TODO: consider using 16 bits
 const XOR_URL_STR_MAX_LENGTH: usize = 44;
-const XOR_NAME_BYTES_OFFSET: usize = 4; // offset where to find the XoR name bytes
+const XOR_NAME_BYTES_OFFSET: usize = 5; // offset where to find the XoR name bytes
 const URL_VERSION_QUERY_NAME: &str = "v";
 
 /// The XOR-URL type
@@ -155,13 +155,9 @@ pub enum DataType {
     #[allow(missing_docs)]
     SafeKey = 0x00,
     #[allow(missing_docs)]
-    PublicBlob = 0x01,
+    Blob = 0x01,
     #[allow(missing_docs)]
-    PrivateBlob = 0x02,
-    #[allow(missing_docs)]
-    PublicRegister = 0x03,
-    #[allow(missing_docs)]
-    PrivateRegister = 0x04,
+    Register = 0x02,
 }
 
 impl std::fmt::Display for DataType {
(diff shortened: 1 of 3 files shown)


src/url/mod.rs
```diff
@@ -172,18 +168,42 @@ impl std::fmt::Display for DataType {
...
+            1 => Ok(Self::Blob),
+            2 => Ok(Self::Register),
             _ => Err(Error::InvalidInput("Invalid DataType code".to_string())),
         }
     }
 }
 
+/// We also encode the data scope - i.e. accessibility on the SAFE Network.
+#[derive(Debug, Copy, Clone, PartialEq, Deserialize, Serialize)]
+pub enum Scope {
```
connec, line 183: Not sure if this is fully settled, but if there's room for some bike-shedding how about `Visibility`, or `Privacy`?

src/url/mod.rs
```diff
@@ -172,18 +168,42 @@ impl std::fmt::Display for DataType {
...
+    #[allow(missing_docs)]
+    Public = 0x00,
+    #[allow(missing_docs)]
+    Private = 0x01,
+}
+
+impl std::fmt::Display for Scope {
```
connec, line 190: How come we implement `Display` for this? Also, using the `Debug` format will end up including quotes, e.g.

```rust
assert_eq!(&format!("{}", Scope::Public), "\"Public\"");
```

src/url/mod.rs
```diff
@@ -503,15 +527,24 @@ impl NativeUrl {
             content_type
         );
 
-        let data_type = match xorurl_bytes[3] {
+        let scope = match xorurl_bytes[3] {
```
connec, line 530: Maybe a future thing since it was the same before, but how come we have `Scope::from_u8` (and `DataType::from_u8`) but repeat the matching here? I don't see either function being used atm.

connec (commented):
Left a couple of comments, mostly just queries.

src/url/mod.rs
oetyng: Ah yeah, not at all settled, it's completely open for suggestions :)

src/url/mod.rs
oetyng: IDK, I just implemented with DataType had, but now that you mention it I could remove them.

src/url/mod.rs
oetyng: Hmm.... same as with `from_u8` .. just implemented what was there for `DataType`.

src/url/mod.rs
oetyng: Not sure if it's maybe used by sn_api or something like that? 
Well, this is already a breaking change, so I'll remove.

src/url/mod.rs
oetyng: Neither `Privacy::Private` nor `Privacy::Public` seems ... well entirely clean, I think.. 
`Visibility::Private` and `Visibility::Public` is better in that way - makes more sense semantically.

But `Scope::Private` and `Scope::Public` is shorter, and I feel it quite makes sense semantically as well?

src/url/mod.rs
oetyng: Neither `Privacy::Private` nor `Privacy::Public` seems ... well entirely clean, I think.. 
`Visibility::Private` and `Visibility::Public` is better in that way - makes more sense semantically.

But `Scope::Private` and `Scope::Public` is shorter, and I feel it quite makes sense semantically as well?

connec: Fair enough, `Scope` feels less obvious to me ("what's the scope of the URL") vs. `Visibility` ("what's the visibility of the URL"), but this isn't enough to block anything.

connec (approved):

src/url/mod.rs
oetyng: Yeah, I'm with you. I'm fine with `Visibility` as well.

PR #294, opened 2021-08-05
refactor(register): make it store pointers only

oetyng:
The pointers are in the form of `NativeUrl`.
This hinders the storage of arbitrary data into the registers,
and moves data storage completely to Adults.
TODO: Use proptest generated data in the tests.

### DIFF
--- a/src/client/client_api/register_apis.rs
+++ b/src/client/client_api/register_apis.rs
@@ -174,7 +174,7 @@ impl Client {
             .get(hash, None)?
             .ok_or_else(|| Error::from(crate::types::Error::NoSuchEntry))?;
 
-        Ok(entry.to_vec())
+        Ok(entry.to_owned())
     }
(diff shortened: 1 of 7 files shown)


src/client/errors.rs
```diff
@@ -129,6 +129,9 @@ pub enum Error {
     /// Database error.
     #[error("Database error:: {0}")]
     Database(#[from] crate::dbs::Error),
+    /// Failed to parse a string.
+    #[error("Failed to parse: {0}")]
+    FailedToParse(String),
```
bochaco, line 134: I think we shouldn't define `Errors` which are meant only to support tests, and by simply using anyhow::Result` or `eyre::Result` as the result from tests then there is no need to map errors.

oetyng: Sure, I agree.

PR #295, opened 2021-08-05
fix(deps): add criterion again

oetyng:
This dev-dependency was for some reason removed by 6299893,
which caused the benches to break.


connec (approved):

oetyng:
Also fixed in #272. Closing this as to not require another rerun of CI for that one.

PR #297, opened 2021-08-05
fix(routing): keep the handle to the RegisterStorage when relocating

bochaco:


bochaco:
This fix in included in PR #201

PR #315, opened 2021-08-11
Remove locking routing core on msgs

joshuef:


src/routing/routing_api/dispatcher.rs
```diff
@@ -130,12 +130,67 @@ impl Dispatcher {
 
     async fn try_handle_command(&self, command: Command) -> Result<Vec<Command>> {
         match command {
+            // Non-data node msg that requires more locking
```
oetyng, line 133: I guess it should say Data msg that require less locking.

oetyng (dismissed):

bochaco (approved):

PR #317, opened 2021-08-11
Client multi-section and AE

Yoga07:


src/client/connections/listeners.rs
```diff
@@ -222,11 +141,25 @@ impl Session {
...
+                    sap: Some(section_auth),
+                    source_message: Some(message),
+                }) => {
+                    {
+                        // Update our network knowledge
+                        let _ = network
+                            .write()
+                            .await
+                            .insert(section_auth.prefix, section_auth);
+                    }
```
bochaco, line 160: What's the enclosing scope for? I don't think you need it, any lock will be dropped at the end of the statement

Yoga07: >the lock will be dropped at the end of the statement?

Ah okay. If so, we do not need the scope. Will remove 👍

src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -329,8 +346,74 @@ impl Core {
...
+        info!("Our SAP: {:?}", our_section);
+        info!("Better SAP: {:?}", better_sap);
+        if better_sap != our_section {
+            // Update the client of the actual destination section
+            info!("We have a better matched section for the data name");
+            Some(better_sap)
+        } else {
+            // Check if client has latest knowledge of our section
+            if received_section_pk != &our_section.section_key() {
+                info!("Client is lagging on knowledge of our section, updating them");
```
maqi, line 412: show our_key and received_key

src/client/connections/messaging.rs
```diff
@@ -6,83 +6,53 @@
...
+// Number of attempts when retrying to send a message to a node
 const NUMBER_OF_RETRIES: usize = 3;
 // Number of Elders subset to send queries to
 const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 3;
 
 impl Session {
-    /// Bootstrap to the network maintaining connections to several nodes.
-    pub(crate) async fn bootstrap(&mut self, client_pk: PublicKey) -> Result<(), Error> {
+    /// Bootstrap to the network, connecting with the closest section to
+    /// the XorName of our PublicKey.
```
maqi, line 38: This bootstrap function only returns the elders that known via the first connected node from the connection config file.
Which is nothing to do with the closest check ?

src/client/connections/messaging.rs
```diff
@@ -91,28 +61,38 @@ impl Session {
     /// Send a `ServiceMsg` to the network without awaiting for a response.
     pub(crate) async fn send_cmd(
         &self,
-        dst_address: XorName,
+        data_name: XorName,
```
maqi, line 64: now the cmd is only data related?
otherwise, I think using the original name of `dst_address` will be more accurate?

src/client/connections/messaging.rs
```diff
@@ -91,28 +61,38 @@ impl Session {
...
+                (
+                    sap.elders
+                        .values()
+                        .cloned()
+                        .take(targets)
+                        .collect::<Vec<SocketAddr>>(),
+                    sap.public_key_set.public_key(),
+                )
+            })
+            .or_else(|| {
```
maqi, line 87: I think here we shall have a `closest` call, which guaranteed to return something (if return nothing, then means NotBootstrapped) ?

src/client/connections/messaging.rs
```diff
@@ -195,35 +134,54 @@ impl Session {
...
+                self.bootstrap_peer
+                    .map(|addr| (vec![addr], bls::SecretKey::random().public_key()))
+            })
+            .ok_or(Error::NotBootstrapped)?;
+
+        // We select the NUM_OF_ELDERS_SUBSET_FOR_QUERIES closest Elders we are querying
+        let chosen_elders = elders
             .into_iter()
-            .sorted_by(|(_, lhs_name), (_, rhs_name)| data_name.cmp_distance(lhs_name, rhs_name))
+            // .sorted_by(|(lhs_name, _), (rhs_name, _)| data_name.cmp_distance(lhs_name, rhs_name)) // FIXME: uncomment once we have PrefixMap read from disk
```
maqi, line 161: Mark that this has to be uncommented before merge

src/client/connections/messaging.rs
```diff
@@ -195,35 +134,54 @@ impl Session {
...
+            DstLocation::Section {
+                name: data_name,
+                section_pk,
+            },
+        )?;
+        let priority = wire_msg.msg_kind().priority();
+        let msg_bytes = wire_msg.serialize()?;
+
+        let elders_len = chosen_elders.len();
+        // if elders_len < NUM_OF_ELDERS_SUBSET_FOR_QUERIES {
```
maqi, line 178: can this block of commented out code to be removed?

src/client/connections/messaging.rs
```diff
@@ -419,86 +368,103 @@ impl Session {
...
+pub(crate) async fn rebuild_message_for_ae_resend(
+    msg_id: MessageId,
+    serialized_cmd: Bytes,
+    auth: ServiceAuth,
+    dst_address: Option<XorName>,
+    network: Arc<RwLock<PrefixMap<SectionAuthorityProvider>>>,
+) -> Option<(WireMsg, Vec<SocketAddr>)> {
+    if let Some(dst_address) = dst_address {
+        if let Some((elders, section_pk)) =
+            network
```
maqi, line 440: For the case of AntiEntropyRedirect, here we shall use the SAP that got included in the Redirect response to send the re-built msg back.
And that SAP cann't be used to update the network knowledge. As we cann't verify it.

AND, the section_pk used can be genesis_key or even a random one, but shall not be the SAP.section_pk, to allow the AE update flow got triggered properly.

src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -329,8 +346,74 @@ impl Core {
                 return Ok(vec![]);
             }
         };
+        info!("Checking for entropy at Client");
+        if let Some(return_sap) = self.check_entropy_for_client(&data_name, &received_section_pk) {
+            info!("Client AE triggered. Seems like client is out of date.");
```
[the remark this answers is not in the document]
Yoga07: Out of date in general to the network. So there is a better matching section or self section has gone through churn and has new keys.

src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -329,8 +346,74 @@ impl Core {
...
+            // AE triggered! Send back the message with SAP of actual destination section
+            let service_msg = ServiceMsg::ServiceError(ServiceError {
```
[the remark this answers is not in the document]
Yoga07: >AntiEntropyRedirect or AntiEntropyRetry

They are both NodeMsgs, so we cannot use them for sending it to client. Hence we use the ServiceError message that is specific to client.

src/client/connections/messaging.rs
Yoga07: Yep, outdated comment. Will remove 👍

src/client/connections/messaging.rs
Yoga07: We are going to have the `try_get_matching` API [here](https://github.com/maidsafe/safe_network/pull/344/files#diff-53290c5eb239638e3b4e332f3a0d1c871102d3b3ded3ec835ca2a2e9bbf40244R76) which will guaranteed return something. The or_else is for when we do not have any knowledge in our PrefixMap, we send the message to our bootstrapping peer with a random key that will give us our first SAP.

src/client/connections/messaging.rs
Yoga07: Yeps, the updated AE flow is upcoming in another PR 👍  We are accepting the given SAP to just move ahead with the tests for now. Soon we'll be re-querying using the given SAP and update our knowledge properly.

src/messaging/data/errors.rs
```diff
@@ -59,4 +59,7 @@ pub enum Error {
         "Could not generation operation id for chunk retrieval. Error was not 'DataNotFound'."
     )]
     InvalidQueryResponseErrorForOperationId,
+    /// Destination is either outdated or incorrect
+    #[error("Destination is either outdated or wrong")]
```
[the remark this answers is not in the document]
Yoga07: Yes, this is the error we return to the client along with the SAP to redirect the client.

src/routing/core/mod.rs
```diff
@@ -156,6 +158,14 @@ impl Core {
             info!("Split");
         }
 
+        if self
+            .network
```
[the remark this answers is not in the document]
Yoga07: Our own section's details are not covered in the `self.network` type, so t fix that, we need to update it whenever we go through churn.

src/routing/core/msg_handling/agreement.rs
```diff
@@ -246,6 +246,6 @@ impl Core {
             }
         }
 
-        self.update_state(snapshot).await
+        self.update_state(signed_section_auth, snapshot).await
```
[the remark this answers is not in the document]
Yoga07: It is needed to do the updated mentioned [here](https://github.com/maidsafe/safe_network/pull/317#discussion_r698413838).

src/messaging/msg_kind.rs
```diff
@@ -45,6 +38,9 @@ pub enum MsgKind {
     /// A message from an Elder node with authority of its whole section.
     // FIXME: find an example.
     SectionAuthMsg(SectionAuth),
+    /// A proxy message for testing
```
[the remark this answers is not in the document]
bochaco: I think we should use some real msg instead of having a test msg

PR #318, opened 2021-08-11
fix: avoid looping NodeErrorMsg wrapping

maqi:

### DIFF
--- a/src/routing/core/msg_handling/mod.rs
+++ b/src/routing/core/msg_handling/mod.rs
@@ -426,16 +426,35 @@ impl Core {
                 error,
                 correlation_id,
             } => {
-                self.send_event(Event::MessageReceived {
-                    msg_id,
-                    src: msg_authority.src_location(),
-                    dst: dst_location,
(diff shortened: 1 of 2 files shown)


src/routing/core/msg_handling/mod.rs
```diff
@@ -426,16 +426,26 @@ impl Core {
...
+                            src: msg_authority.src_location(),
+                            dst: dst_location,
+                            msg: Box::new(MessageReceived::NodeMsgError {
+                                error,
+                                correlation_id,
+                            }),
+                        })
+                        .await
+                    }
+                    _ => {}
```
oetyng, line 446: maybe put a wee log here when not handling the case?

bochaco (changes requested):
@maqi let's send this PR against https://github.com/maidsafe/safe_network/tree/Feat-ChunksToRouting branch so we prevent from having to resolve conflicts. That branch is already more stable than main so will soon be merged I think.

bochaco:
Superseded by PR #325

PR #325, opened 2021-08-12
fix: avoid looping NodeErrorMsg wrapping

maqi:

### DIFF
--- a/src/routing/core/msg_handling/mod.rs
+++ b/src/routing/core/msg_handling/mod.rs
@@ -186,7 +186,7 @@ impl Core {
         sender: SocketAddr,
         msg_id: MessageId,
         msg_authority: NodeMsgAuthority,
-        dst_location: DstLocation,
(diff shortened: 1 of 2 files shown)


src/node/event_mapping/node_msg.rs
```diff
@@ -76,6 +76,18 @@ fn match_node_msg(msg: MessageReceived) -> NodeDuty {
         MessageReceived::NodeCmd(NodeCmd::StorageFull { node_id, .. }) => {
             NodeDuty::IncrementFullNodeCount { node_id }
         }
+        MessageReceived::NodeMsgError {
```
joshuef, line 79: I'd suggest we remove this from node event map if we're not doing anything. 

We could log this directly in routing/core/msg_handling and remove another `send_event`

joshuef (changes requested):

joshuef (approved):

PR #341, opened 2021-08-20
fix(routing): restart DKG session with new set of candidates upon a failed one

bochaco:


src/routing/core/msg_handling/dkg.rs
```diff
@@ -93,24 +90,27 @@ impl Core {
...
-            // The DKG failure is a corrupted one due to lagging.
-            trace!(
-                "Received DKG failure agreement - restarting: {:?}",
-                elder_candidates
-            );
-
-            self.send_dkg_start(elder_candidates, slice::from_ref(sender))
-        } else {
+        let mut commands = vec![];
+        /*
```
maqi, line 94: Just remove this part of code ?

src/routing/core/msg_handling/dkg.rs
```diff
@@ -93,24 +90,27 @@ impl Core {
...
+            commands.extend(self.cast_offline_proposals(&failure_set.failed_participants)?);
         }
+        */
+
+        trace!(
+            "Received DKG failure agreement, we will restart with new candidates. Previous candidates: {:?}",
+            elder_candidates
+        );
+
+        commands.extend(self.promote_and_demote_elders_except(&failure_set.failed_participants)?);
```
maqi, line 112: previously, when a node got voted off, it triggers a new round of DKG vote as well.

also, the `exclusion` part is not implemented yet within the promote_and_demote_elders_except,
so the result will still include the failed_participant.

we may need to discuss whether such exclusion is really needed?
i.e. for the case of a new node just joined, that missed the original DkgStart, the exclusion is bad
     however for the case of a bad guy not response in DKG vote, the exclusion is needed.

src/routing/routing_api/dispatcher.rs
```diff
@@ -324,6 +324,14 @@ impl Dispatcher {
         delivery_group_size: usize,
         wire_msg: WireMsg,
     ) -> Result<Vec<Command>> {
+        match wire_msg.clone().into_message() {
```
maqi, line 327: this introduce an extra deserialization of payload, which normally we want to avoid.

is it really necessary?

maqi (commented):
Better mark this PR as WIP as it needs some further discussion and work.

src/routing/core/msg_handling/dkg.rs
bochaco: ah yeah, this is a draft code yet, note I marked the PR as draft

src/routing/core/msg_handling/dkg.rs
bochaco: the reason you see things incomplete is because this PR is a draft, so it's not ready for review yet

bochaco: > i.e. for the case of a new node just joined, that missed the original DkgStart, the exclusion is bad
I don't think so, it's already misbehaving, so it may not be too bad to already dismiss it. Moreover, perhaps in the future we want to punish/relocate it, this way we have an additional incentive for Adults to behave. If an Adult is in the state to be promoted and is not responsive to the DKG process for promotion then maybe it shouldn't be promoted...

src/routing/routing_api/dispatcher.rs
bochaco: this is temporary coed which reproduces the original issue of Adults missing the DKG start and Elders not picking up another Adult for promotion, so it's just to reproduce the problem. Again, that's why this PR is marked as draft

src/routing/core/msg_handling/dkg.rs
maqi: but as you observed, the reason that adult not responsing is because it missed the DkgStart.
And the exclusion will prevent it to be promoted. which is not ideal.
As this will cause an extra new round of DKG immediately during next churn, as this Adult still having a `higher age`

bochaco: >  the reason that adult not responsing is because it missed the DkgStart

it's the adult's problem, Elders don't care what the adult problem was, it just didn't respond to the DKG msgs then it cannot be part of it. How do you know it's a faulty adult, a rogue adult, or a sybil adult trying to do something else by disrupting the DKG rounds? you don't know, so you cannot assume anything like assuming it just missed the dkg start msg.

> As this will cause an extra new round of DKG immediately during next churn, as this Adult still having a higher age

so that's why I said we may eventually want to not only remove it form next DKG round but also punish it, perhaps reducing its age and/or relocating it.

src/routing/section/mod.rs
```diff
@@ -229,14 +229,24 @@ impl Section {
...
+    /// excluding any member matching a name in the provided `excluded_names` set.
     /// Returns a set of candidate SectionAuthorityProviders.
-    pub(super) fn promote_and_demote_elders(&self, our_name: &XorName) -> Vec<ElderCandidates> {
+    pub(super) fn promote_and_demote_elders(
+        &self,
+        our_name: &XorName,
+        excluded_names: &BTreeSet<XorName>,
+    ) -> Vec<ElderCandidates> {
         if let Some((our_elder_candidates, other_elder_candidates)) = self.try_split(our_name) {
             return vec![our_elder_candidates, other_elder_candidates];
```
maqi, line 241: when there is a split, the `excluded names` is not considered?

src/routing/section/section_peers.rs
```diff
@@ -102,14 +107,17 @@ impl SectionPeersUtils for SectionPeers {
...
     ) -> Vec<Peer> {
         elder_candidates(
             elder_size,
             current_elders,
             self.members
-                .values()
-                .filter(|info| is_active(&info.value, current_elders))
-                .filter(|info| info.value.peer.is_reachable()),
+                .iter()
+                .filter(|(name, _)| !excluded_names.contains(name))
```
maqi, line 117: put this as the last checking criteria will save some computation power in theory?

src/routing/section/mod.rs
bochaco: good point, i guess we do need to consider them to...I'll try to fix it now

src/routing/section/section_peers.rs
```diff
@@ -119,15 +128,18 @@ impl SectionPeersUtils for SectionPeers {
...
             elder_size,
             current_elders,
-            self.members.values().filter(|info| {
-                info.value.state == MembershipState::Joined
-                    && prefix.matches(info.value.peer.name())
-                    && info.value.peer.is_reachable()
-            }),
+            self.members
+                .iter()
+                .filter(|(_, info)| info.value.state == MembershipState::Joined)
```
maqi, line 138: just curious, from the performance perspective, is the previous approach more efficient?

bochaco: I just double checked and the filter applies the condition to each element before next iteration kicks in, so it's the same but cleaner?

src/routing/core/api.rs
```diff
@@ -237,10 +235,30 @@ impl Core {
     pub(crate) fn promote_and_demote_elders(&mut self) -> Result<Vec<Command>> {
```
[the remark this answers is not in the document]
bochaco: yes, I'll change it

maqi (approved):

PR #342, opened 2021-08-21
refactor(monitoring): report on every +/-10 %-points used

oetyng:
Instead of `Adults` reporting when at `50%`, and then considered full, and
`Elders` measuring the total resource usage by checking number of
full `Adults` as a share of non-full.. we simply have `Adults` send in their
number every `10%-points` increase / decrease.
An `Adult` is now considered full when at `90%` or more.
If their used space ratio drops by a full `10%-points` below a previously reported value (follwing deletion of private chunks), this is reported to `Elders` as well.

This will now give us a far better measurement of the section available storage.
Before, we could have had all `Adults` being at `49.9999... %` full (not unlikely if eventually going with uniform storage sizes), and the `Elders` would consider used space to be `0 %`. Then after a few more chunks, they would consider used space to be `100 %`.

Additionally, `Adults` would never be able to use more than `50 %` of their allocated space.. to no gain for the network (it wouldn't be a buffer, just an arbitrary `"what ever you say you have, we only accept half of it"`-rule),

Now, we will instead have a measurement which is accurate within roughly `10 %-points` on average.
Also, we will allow `Adults` to use up to `90 %` of their allocated storage, before we consider them full.


src/messaging/data/data_exchange.rs
```diff
@@ -41,10 +41,57 @@ pub struct DataExchange {
...
+
+/// The degree to which storage has been used.
+/// Expressed in values between 0-10, where each unit represents 10-percentage points.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
+pub struct StorageLevel(u8);
+
+///
+impl StorageLevel {
+    /// The maximum level (100%).
+    pub const MAX: u8 = 10;
```
iancoleman, line 60: This doesn't follow the principle of least surprise. It would seem natural for a percent to be between 0-100, so maxing at 10 is maybe embedding too much knowledge, ie we need to know it's measured as 10 chunks of 10%

src/routing/core/capacity/mod.rs
```diff
@@ -6,149 +6,128 @@
...
     }
 
-    /// Whether the adult is recorded as full
-    pub(super) async fn is_full(&self, adult: &XorName) -> bool {
-        self.adult_storage_info
-            .full_adults
-            .read()
-            .await
-            .contains(adult)
+    /// Avg usage by nodes in the section, a value between 0 and 100.
```
iancoleman, line 52: This would be between 0 and 10, since max level is 10

src/messaging/data/data_exchange.rs
oetyng: Yes. That reporting is coarse grained and happens at every level.
In contrast if using percent, with a value between 0 and 100 we are giving an impression that values between whole 10-s are important. Also there's then the other rule that we report at every modulo 10 == 0, instead of at every level..

I though about it for a second or so, and wasn't persuaded by myself to go that way.

iancoleman: Yeah I think leaving it abstract as 'level' is no worry. It also leaves the door open for other measures of stress beyond storage percent full.

oetyng: I'm not against using `0-100`, it would allow changing the granularity of reporting.
It's just that with this change now, I wanted to move gently.
It seems to me that even though not the previous solution wasn't functioning correctly (doing what was expected and needed), who ever came up with the previous way of measurement seems to have wanted some granularity. And that is preserved here.

But this needs to be built on. With this PR we just move from dysfunctional to perhaps OK-ish.

For broader resource measurement (which I believe is necessary sooner or later), I'd imagine we might have a `Resources` enum

```
Resources {
    Storage(StorageLevel), // rather fix and singular
    Cpu(CpuLevel), // probably a few moving averages,
    Memory(MemoryLevel), // somewhat a mix..?
    Bandwidth(BandwidthLevel), // probably a few moving averages,
}
```

This can then tell us things like:
- What role is this node suitable for (Adult, Elder, Archive, .. etc)?
- What is the health of the section?
- Do we need to split?
- Op cost +/-
- ...

oetyng: The problem is not that of gathering info though..
.. but to collect just as much as needed, to be able to make necessary decisions.

More info than that is a burdon in all ways.

Yoga07 (approved):

PR #344, opened 2021-08-25
Dashmap prefixmap

grumbach:
- Add PrefixMaps using DashMaps


src/routing/network/prefix_map.rs
```diff
@@ -0,0 +1,366 @@
...
+            .filter(move |p| p.key().is_extension_of(prefix))
+    }
+
+    /// Remove `prefix` and any of its ancestors if they are covered by their descendants.
+    /// For example, if `(00)` and `(01)` are both in the map, we can remove `(0)` and `()`.
+    pub(crate) fn prune(&self, mut prefix: Prefix) {
+        // TODO: can this be optimized?
+
+        loop {
+            {
```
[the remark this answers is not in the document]
grumbach: The unused functions will be used soon by @Yoga07. 
The inner scope is here for descendants (borrowing `prefix`) to be dropped before prefix is modified below with the 
`prefix = prefix.popped();`

src/routing/network/mod.rs
```diff
@@ -35,27 +36,23 @@ impl Network {
 
     /// Returns the known section that is closest to the given name, regardless of whether `name`
     /// belongs in that section or not.
-    pub(crate) fn closest(&self, name: &XorName) -> Option<&SectionAuth<SectionAuthorityProvider>> {
+    pub(crate) fn closest(&self, name: &XorName) -> Option<SectionAuth<SectionAuthorityProvider>> {
```
Yoga07, line 39: Do we need this API? This is what `get_matching` does internally right? 
If we need it more lenient we can use the `get_matching_or_opposite`.

bochaco (approved):

src/routing/network/mod.rs
Yoga07: We need this API as it differs from the mentioned ones 👍

Yoga07 (approved):

PR #349, opened 2021-08-31
fix(client): use the original payload destination when resending message

lionel1704:
with updated dest section key


joshuef:
Closed as similar flow incorporated by #348 in the end

PR #358, opened 2021-09-01
feat(client): use the network genesis key to validate SAPs received in AE messages

bochaco:


src/client/connections/listeners.rs
```diff
@@ -197,4 +163,151 @@ impl Session {
...
+    // Handle Antry-Entropy Redirect messages
+    async fn handle_ae_redirect_msg(
+        &self,
+        section_auth: SectionAuthorityProvider,
+        _section_signed: KeyedSig,
+        bounced_msg: Bytes,
+        sender: SocketAddr,
+    ) -> Result<(), Error> {
+        // TODO: Check if SAP signature is valid
+        /*let signed_section_auth = SectionAuth {
```
Yoga07, line 176: Can we uncomment this?

bochaco: yeah, I'll do it now

Yoga07 (approved):

PR #360, opened 2021-09-02
Testnet logging

connec:
- ed8d2953e **refactor: Improve error handling in `testnet` binary**

  This now uses `eyre` for errors (and `color-eyre` for reporting) and has
  more robust handling of some error cases (e.g. it will no longer ignore
  legit errors when removing the node directory or handling non-UTF8
  `NODE_COUNT` values).

- f43150b60 **chore: Update `sn_launch_tool`**

  The new version uses `tracing` for all its output, which gives us more
  consistent output. It also fixes the handling of the `--local` flag, and
  has some breaking API changes that required changing how we call it.


Cargo.toml
```diff
@@ -77,7 +77,7 @@ serde_bytes = "0.11.5"
 serde_json = "1.0.53"
 signature = "1.1.10"
 sled = "0.34.6"
-sn_launch_tool = "0.3.2"
+sn_launch_tool = { git = "https://github.com/connec/sn_launch_tool", branch = "egregious-refactor" }
```
oetyng, line 80: Just needs this updated :)

oetyng (changes requested):

Cargo.toml
connec: Good catch 🤦

oetyng (approved):

PR #364, opened 2021-09-02
fix: full adults being chunk holders as well

maqi:

### DIFF
--- a/src/routing/core/msg_handling/service_msgs.rs
+++ b/src/routing/core/msg_handling/service_msgs.rs
@@ -20,7 +20,7 @@ use crate::routing::peer::PeerUtils;
 use crate::routing::{error::Result, routing_api::command::Command, section::SectionUtils};
 use crate::types::PublicKey;
 use itertools::Itertools;
-use std::collections::BTreeSet;
+use std::{cmp::Ordering, collections::BTreeSet};
 use xor_name::XorName;
 
 impl Core {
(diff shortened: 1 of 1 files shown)


oetyng (approved):

oetyng (changes requested):
Only include full adults who would be close to the chunk.

src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -294,11 +294,38 @@ impl Core {
...
-            .collect::<BTreeSet<_>>()
+            .collect::<BTreeSet<_>>();
+        trace!(
+            "Chunk holders of {:?} are empty adults: {:?} and full adults: {:?}",
+            target,
+            candidates,
+            full_adults
+        );
+
+        // Full adults that close to the chunk, shall still be considered as candidates
```
oetyng, line 309: miniscule nitpick: `Full adults that *are* close to the`

src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -294,11 +294,38 @@ impl Core {
...
+        // to allow full loaded adult to be queried later on.
```
oetyng, line 310: > to allow full loaded adult

Maybe enough with `to allow full adult`?

oetyng (approved):

oetyng (approved):

PR #367, opened 2021-09-06
chore(ci): combine E2E and all other tests

oetyng:
- Reduces compilations to 1 per PR run, and by that num workers needed.
This should then allow more PRs to run concurrently without clogging GA.


.github/workflows/pr_tests.yml
```diff
@@ -105,3 +81,135 @@ jobs:
...
+        if: matrix.os == 'windows-latest'
+        run: rd /s /q %USERPROFILE%\.safe\node
+        shell: cmd
+
+      - name: Remove node dir and logs for next test
+        if: matrix.os != 'windows-latest'
+        run: rm -rf ~/.safe/node
+
+      - name: Run example of split and chunk check
+        if: false
```
jacderida, line 196: Should we include these 2 steps if they're not going to run? Or is that because they might be re-enabled in the future?

oetyng: Yeah, good point. I'm not sure if we prefer commenting out, keeping or deleting.
But just running 15 nodes I think is very temporary, until we get splits stabilizing, and so I've followed that, except here. To be consistent we should comment these out as well.

oetyng: ..oor IDK, this is separate. Why do you mean they are not going to run?

jacderida: The condition is set to `false`, so they'll never run as they are just now.

joshuef: this wont run atm due to `false`. 

I think this should be able to be re-enabled (just remove the `if`)
```suggestion

```

.github/workflows/pr_tests.yml
```diff
@@ -105,3 +81,135 @@ jobs:
...
         run: cargo test --release types 
+
+      - name: Build all safe_network targets
+        run: cargo build --all-targets --release --features=always-joinable,test-utils --bins --examples --tests
+
+      - run: ./target/release/testnet
+        env:
+          NODE_COUNT: 15 # TODO: change it back to 43 (this should be a split section)
+
+      # - name: extra wait for CI split
```
jacderida, line 92: Btw, in general I'm not a fan of leaving commented out code in commits. Do we want to keep this around for some kind of reference?

joshuef: naw we dont need that if it's not being used atm (that's a hangover from the e2e file i believe). kill it

oetyng: These numbers (node count, i.e. split or not split in e2e) keep changing back and forth all the time while we are not stable.

I guess it is too much of a hassle to delete and lookup/write again while we're not stable.

So, yeah, all the commented out stuff are things that would be brought back when NODE_COUNT is set to 43 again.

jacderida: Cool, it's not a huge big deal either way.

jacderida (approved):

PR #369, opened 2021-09-06
Update neighbouring section's SAP on splits

Yoga07:


src/prefix_map/mod.rs
```diff
@@ -214,6 +229,7 @@ impl NetworkPrefixMap {
     /// Get the entry at the prefix that matches `name`. In case of multiple matches, returns the
     /// one with the longest prefix. If there are no prefixes matching the given `name`, return
     /// a prefix matching the opposite to 1st bit of `name`. If the map is empty, return None.
+    #[allow(unused)]
```
bochaco, line 232: Shall we just remove it if we are not using it?

src/prefix_map/mod.rs
```diff
@@ -68,13 +68,28 @@ impl NetworkPrefixMap {
...
+    fn closest(&self, name: &XorName) -> Option<SectionAuth<SectionAuthorityProvider>> {
         self.sections
             .iter()
             .min_by(|lhs, rhs| lhs.key().cmp_distance(rhs.key(), name))
             .map(|e| e.value().clone())
     }
 
+    /// Returns the known section that is closest to the given name, regardless of whether `name`
+    /// belongs in that section or not. If there are no close matches, return a SAP from an opposite prefix.
+    pub(crate) fn closest_or_opposite(
```
maqi, line 80: not sure what's the point of `opposite`
the function says `closest` , which means as long as we know a section, this function shall return with a result.
And, in case of `no close matches`, it means we don't know a section, which won't have a result for `opposite` either?

src/routing/core/msg_handling/agreement.rs
```diff
@@ -227,20 +227,38 @@ impl Core {
...
+                    &old_chain,
+                    self.section_chain(),
+                ) {
+                    error!("Error updating neighbouring section's details on our NetworkPrefixMap: {:?}", e);
+                }
             }
         }
 
-        self.update_state(signed_section_auth, snapshot).await
+        info!("PREFIXES WE KNOW ABOUT: {:?}", self.network);
```
bochaco, line 260: I guess this msg shall be lower case

src/prefix_map/mod.rs
Yoga07: This is for scenarios where there is just one branch of sections in the PrefixMap(happens mostly at the client). Eg: The map would contain only 10, 11 and so on where prefixes start with just 1. So a `closest` match for a dst name `00111.....` would only return `None`. So as a failsafe, we need to return at least one prefix from the opposite section.

src/prefix_map/mod.rs
```diff
@@ -255,13 +271,12 @@ impl NetworkPrefixMap {
             .iter()
             .map(|e| e.value().value.elder_count())
             .sum();
-        let known = our.elder_count() + network_elders_count;
-        let total = known as f64 / network_fraction;
+        let total = network_elders_count as f64 / network_fraction;
```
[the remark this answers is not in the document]
Yoga07: Nop, `networks_elders_count` now contains elders of our own section.

src/routing/core/mod.rs
```diff
@@ -142,11 +141,7 @@ impl Core {
         }
     }
 
-    pub(crate) async fn update_state(
-        &mut self,
-        new_section_auth: SectionAuth<SectionAuthorityProvider>,
-        old: StateSnapshot,
-    ) -> Result<Vec<Command>> {
+    pub(crate) async fn update_state(&mut self, old: StateSnapshot) -> Result<Vec<Command>> {
```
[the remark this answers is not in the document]
Yoga07: We only use the old state to make comparisons and take decisions on what to do with the churn/split :)

src/prefix_map/mod.rs
maqi: then what if you have 01, 11, and `closest match` for `00111` ? returns `01` ? 
If as you explained, I think the name of the function is bit misleading, `closest` shall only mean Xor min distance among the sections to the target. Which, shall return a section as long as knows a section.

bochaco (approved):

PR #373, opened 2021-09-07
Refactor chunk handling

oetyng:
- Makes use of the refactored `self-encryption`.
- Removes private chunk owner.
- Moves private/public chunk distinction to blob level.
- Refactors some client session handling.
- Updates `blob_api` tests.


joshuef:
is there a need for both `tests::create_and_retrieve_20mb_private` and `tests::create_and_retrieve_10mb_private` to run?

src/client/connections/listeners.rs
```diff
@@ -162,16 +182,18 @@ impl Session {
...
             };
         });
+
+        Ok(session)
     }
 
     // Handle Antry-Entropy Redirect messages
     async fn handle_ae_redirect_msg(
-        &self,
+        session: Session,
```
Yoga07, line 191: Why do we want to pass in a Session? We do not have multiple sessions atm.

src/client/connections/listeners.rs
```diff
@@ -229,21 +251,23 @@ impl Session {
...
-        send_message(elders, wire_msg, self.endpoint.clone(), msg_id).await
+        send_message(elders, wire_msg, session.endpoint.clone(), msg_id).await?;
+
+        Ok(session)
     }
 
     // Handle Antry-Entropy Retry messages
     async fn handle_ae_retry_msg(
-        &self,
+        session: Session,
```
Yoga07, line 261: Same as above.

oetyng:
> is there a need for both `tests::create_and_retrieve_20mb_private` and `tests::create_and_retrieve_10mb_private` to run?

We had a 10MB and a 100MB before, which never ran on ci, so I thought we could put something there higher than 10 that does run.
Whether it needs to, meh. As soon as we have tested the general capability of handling more chunks (with min blob size that's 3 chunks, and then we go higher and higher), any additional chunks are essentially a load test.

src/messaging/data/mod.rs
```diff
@@ -41,6 +39,13 @@ use xor_name::XorName;
 /// TODO: make uniquer per requester for some operations
 pub type OperationId = String;
 
+/// Return operation Id of a chunk
+pub fn operation_id(address: &ChunkAddress) -> Result<OperationId> {
```
Yoga07, line 43: If we are batching chunks now? Should we have an OpId for an entire batch?

oetyng: the batching is not part of this PR, have a draft PR raised with that stuff

src/client/connections/listeners.rs
oetyng: It's a different pattern. Instead of calling methods on &self we're passing in a session instance.

Logic is clearer, less entagled, and at upper layers it makes it makes the session more "session-like". I.e. something that you can have many of.

src/client/connections/listeners.rs
oetyng: It's a different pattern. Instead of calling methods on &self we're passing in a session instance.

Logic is clearer, less entagled, and at upper layers it makes it makes the session more "session-like". I.e. something that you can have many of.

src/messaging/data/mod.rs
Yoga07: Yea, right. Just a tangential thought of making `operation_id` take just a XorName 👍

oetyng: Yeah, I will look into that. I think I already had a small encounter with that Q as I was doing the batching.
We'll get that discussed when I circle back there.

src/client/client_api/commands.rs
```diff
@@ -52,8 +52,8 @@ impl Client {
         // So it's no more than 2 with 7 Elders.
         // With 3 we are "guaranteed" 1 correctly functioning Elder.
         let targets = match &cmd {
-            DataCmd::Chunk(_) => 3, // stored at Adults, so only 1 correctly functioning Elder need to relay
-            DataCmd::Register(_) => 7, // only stored at Elders, all need a copy
+            DataCmd::StoreChunk(_) => 1, // stored at Adults, so only 1 correctly functioning Elder need to relay
```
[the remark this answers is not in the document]
oetyng: Comments should be added for this, it was only mentioned in the commit itself. So, this is just a temporary change now that we were looking at resource usage. I was unsure about how temporary it should be. Considering not incling it here..

src/client/connections/listeners.rs
```diff
@@ -20,120 +15,145 @@ use crate::messaging::{
...
+                let result = Self::handle_ae_redirect_msg(
+                    session,
+                    section_auth,
+                    section_signed,
+                    bounced_msg,
+                    src,
+                )
+                .await;
+                if result.is_err() {
+                    warn!("Failed to handle AE-Redirect");
```
maqi, line 93: log the result as well?

src/client/connections/listeners.rs
```diff
@@ -20,120 +15,145 @@ use crate::messaging::{
...
+                let result = Self::handle_ae_retry_msg(
+                    session,
+                    section_auth,
+                    section_signed,
+                    bounced_msg,
+                    proof_chain,
+                )
+                .await;
+                if result.is_err() {
+                    warn!("Failed to handle AE-Retry msg");
```
maqi, line 116: log the result?

src/client/connections/listeners.rs
```diff
@@ -20,120 +15,145 @@ use crate::messaging::{
...
-                        debug!("Query response (op_id is: {})", op_id);
-
                         if let Some(sender) = &queries.read().await.get(&op_id) {
                             trace!("Sending response for query w/{} via channel.", op_id);
                             let _ = sender.send(response).await;
                         } else {
-                            trace!("No channel found for operation {}", op_id);
+                            // temporarily commenting this out
+                            // if we have already received enough responses for a query,
+                            // we drop the channels and any drop further responses for that query.
```
maqi, line 153: and drop any further responses for that query ?

src/client/client_api/data/pac_man.rs
```diff
@@ -0,0 +1,134 @@
...
+    // which points to all of those additional chunks.. and so on.
+    let mut chunks = vec![];
+    let mut chunk_content = pack_secret_key(SecretKey::FirstLevel(secret_key), encryption)?;
+
+    let (address, additional_chunks) = loop {
+        let chunk = to_chunk(chunk_content, encryption)?;
+        // If data map chunk is less that 1MB return it so it can be directly sent to the network
+        if chunk.validate_size() {
+            let name = *chunk.name();
+            chunks.reverse();
```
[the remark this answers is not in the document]
oetyng: No, the reverse only happens once. And it's because we have chained all previous rounds to the end of the newest round of chunks. That means the newer chunks are at the front of the vec. So, when we reverse and push the final chunk, we have them all in order.

Example:_

```
data_chunks: [x, k, p..] // there is no order here
after first head chunk: [1, x, k, p, ..]
after second first head chunk: [2, 1, x, k, p, ..]
after last head chunk: [.., p, k, x, 1, 2, 3]
```

src/client/connections/listeners.rs
```diff
@@ -20,120 +15,145 @@ use crate::messaging::{
...
 
                     if let Ok(op_id) = response.operation_id() {
-                        debug!("Query response (op_id is: {})", op_id);
-
                         if let Some(sender) = &queries.read().await.get(&op_id) {
                             trace!("Sending response for query w/{} via channel.", op_id);
                             let _ = sender.send(response).await;
                         } else {
-                            trace!("No channel found for operation {}", op_id);
+                            // temporarily commenting this out
```
maqi, line 151: put a TODO instead?
also, no sure what's the link between the below comment and `comment out the trace log`

src/client/connections/listeners.rs
```diff
@@ -20,120 +15,145 @@ use crate::messaging::{
...
+                            // if we have already received enough responses for a query,
+                            // we drop the channels and any drop further responses for that query.
+                            // but we should not drop it immediately, but clean it up after a while
+                            // and then not log that "no channel was found" when we already had enough responses..
```
maqi, line 155: instead clean it up
`enough responses..` => enough responses.

examples/routing_stress.rs
```diff
@@ -382,7 +382,7 @@ impl Network {
     }
 
     // Returns the socket addresses to bootstrap against.
-    async fn get_bootstrap_addrs(&self) -> Vec<SocketAddr> {
+    async fn get_bootstrap_nodes(&self) -> BTreeSet<SocketAddr> {
```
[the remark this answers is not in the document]
oetyng: The reason is that further down the line we're using `BTreeSet`, so this was a way to do less conversions.
Might need another walkthrough though, to see what's most efficient.

src/client/connections/messaging.rs
```diff
@@ -10,26 +10,131 @@ use super::{QueryResult, Session};
...
+    time::sleep,
+};
 use tracing::{debug, error, trace, warn};
 use xor_name::XorName;
 
 // Number of attempts when retrying to send a message to a node
 const NUMBER_OF_RETRIES: usize = 3;
 // Number of Elders subset to send queries to
-const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 3;
+const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 1;
```
maqi, line 42: just to confirm, will this be too risky ?

src/client/connections/messaging.rs
```diff
@@ -10,26 +10,131 @@ use super::{QueryResult, Session};
...
+        );
+        debug!("QP2p config: {:?}", qp2p_config);
+
+        let bootstrap_nodes = bootstrap_nodes.iter().copied().collect_vec();
+        let (endpoint, incoming_messages, _) =
+            Endpoint::new_client(local_addr, qp2p_config.clone())?;
+        let bootstrap_peer = endpoint.connect_to_any(&bootstrap_nodes).await;
+
+        // *****************************************************
+        // FIXME: receive the network's genesis pk from the user
```
maqi, line 67: think there is a plan to read the genesis_key from the config file?

src/client/connections/messaging.rs
```diff
@@ -10,26 +10,131 @@ use super::{QueryResult, Session};
...
 // Number of attempts when retrying to send a message to a node
 const NUMBER_OF_RETRIES: usize = 3;
 // Number of Elders subset to send queries to
-const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 3;
+const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 1;
+// Number of attempts to make when trying to bootstrap to a section
+const NUM_OF_BOOTSTRAPPING_ATTEMPTS: u8 = 3;
 
 impl Session {
+    /// Acquire a session by bootstrapping to a section, maintaining connections to several nodes.
```
maqi, line 47: the implementation seems only connected to one of the bootstrap_nodes?

src/client/connections/messaging.rs
```diff
@@ -10,26 +10,131 @@ use super::{QueryResult, Session};
...
+
+        Self::spawn_message_listener_thread(session.clone(), incoming_messages).await;
+
+        Ok(session)
+    }
+
+    /// Utility function that bootstraps a client to a section. If there is a failure then it retries.
+    /// After a maximum of three attempts if the boostrap process still fails, then an error is returned.
+    #[async_recursion]
+    pub(crate) async fn attempt_bootstrap(
```
maqi, line 92: got bit confused here.
seems the `bootstrap` here only means `connected_to`
From routing/node perspective, `bootstrap` means accepted by the network, i.e. for a client, that client's pk shall matches the section it connect to. Hence there could be a step of `redirect` happens, i.e. the first node that a client connected to is not the right section the client shall be, hence being asked to connect to another section.

src/dbs/kv_store/mod.rs
```diff
@@ -147,11 +147,11 @@ impl<K: Key, V: Value + Send + Sync> KvStore<K, V> {
                 let key = value.key().to_db_key()?;
                 Ok((key, serialised_value))
             })
-            .partition(|r| r.is_err());
+            .partition(|r| r.is_ok());
 
         if !err_results.is_empty() {
             for e in err_results {
-                error!("{:?}", e);
+                println!("{:?}", e);
```
maqi, line 154: why println?

src/dbs/kv_store/mod.rs
```diff
@@ -175,11 +175,11 @@ impl<K: Key, V: Value + Send + Sync> KvStore<K, V> {
 
         match res {
             Ok(_) => {
-                info!("Writing batch succeeded!");
+                println!("Writing batch succeeded!");
                 Ok(())
             }
             Err(e) => {
-                info!("Writing batch failed!");
+                println!("Writing batch failed!");
```
maqi, line 182: why println?

src/node/network.rs
```diff
@@ -42,9 +45,9 @@ impl Network {
             first: config.is_first(),
             local_addr: config
                 .local_addr
-                .unwrap_or_else(|| SocketAddr::from((Ipv4Addr::LOCALHOST, 0))),
-            bootstrap_nodes: config.hard_coded_contacts.iter().copied().collect(),
-            transport_config: config.network_config().clone(),
+                .unwrap_or_else(|| SocketAddr::from((Ipv4Addr::LOCALHOST, rand::random()))),
```
maqi, line 48: passing 0 allows the OS to pick next available port for the SocketAddr, 
using random port could result in the port already occupied or out-of-range?
better double check it.

src/routing/core/chunk_records.rs
```diff
@@ -90,21 +67,19 @@ impl Core {
...
         &self,
         chunk: Chunk,
         msg_id: MessageId,
         auth: AuthorityProof<ServiceAuth>,
         origin: EndUser,
     ) -> Result<Vec<Command>> {
-        if let Err(error) = validate_chunk_owner(&chunk, &auth.public_key) {
-            return self.send_error(error, msg_id, origin).await;
-        }
+        trace!("Sending chunk to adults {:?}", chunk);
```
maqi, line 77: Sending chunk {:?} to adults.", chunk.name()

src/routing/core/comm.rs
```diff
@@ -470,6 +470,7 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "Failing for some yet unknown qp2p/os difference reason. To be enabled promptly!"]
```
maqi, line 473: as mentioned in the slack discussion,
you can put a sleep after the disconnection, before the send, to make the test pass in Linux.
Just put a FIXME comment above the sleep as a notice.

src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -55,17 +55,13 @@ impl Core {
                     );
                     self.section
                         .merge_chain(&signed_section_auth, proof_chain)?;
-                    self.section.merge_members(members)?;
```
maqi, line 58: why this can not be removed?
thought it is the main purpose of AEUpdate ?

src/client/client_api/blob_storage.rs
```diff
@@ -6,151 +6,56 @@
...
-    client: Client,
-    public: bool,
-}
-
-impl BlobStorage {
-    /// Create a new BlobStorage instance.
-    pub(super) fn new(client: Client, public: bool) -> Self {
-        Self { client, public }
-    }
+#[allow(unused)]
```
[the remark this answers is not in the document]
oetyng: It was actually a file that should have been removed. Removed now :+1:

src/client/client_api/commands.rs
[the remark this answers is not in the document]
oetyng: I removed this change.

src/client/connections/listeners.rs
oetyng: It's returned to upper layer, where it is logged.
Too much logging if doing it on every level.

src/client/connections/listeners.rs
oetyng: It's returned to upper layer, where it is logged.
Too much logging if doing it on every level.

src/client/connections/messaging.rs
oetyng: this is from a previous PR, it's his code, I have just moved the fn.

src/dbs/kv_store/mod.rs
oetyng: probably just a remaining thing from debugging

src/dbs/kv_store/mod.rs
oetyng: probably just a remaining thing from debugging

src/node/network.rs
oetyng: will put 0 instead

src/routing/core/msg_handling/anti_entropy.rs
oetyng: you mean, _now_ be removed?

it is put outside of the if/else, since both had the same call.. (also, a clippy demand)

src/routing/core/comm.rs
oetyng: ok, I'll put a sleep then

src/client/connections/messaging.rs
oetyng: >  for a client, that client's pk shall matches the section it connect to. 

this is not quite how it works anymore though?

A client connects to any section.

I think the `bootstrap` thing can be ditched all together.

We connect to a section, and that's it.

src/client/client_api/data/pac_man.rs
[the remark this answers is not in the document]
maqi: right, I get it.
as long as `1-2 are each placeholders for any number of chunks (with no order within them)`, I am fine with it.

src/client/connections/listeners.rs
oetyng: Instead clean up what?

src/client/connections/listeners.rs
oetyng: Will put a TODO.

The trace is only needed when we have an _identified_ case of not finding a channel when expecting one, we can log "No channel found for operation", (and then probably at warn or error level). But when we have received enough responses, we aren't really expecting a channel there, so there is no reason to log anything.
That's the link.

src/client/connections/messaging.rs
oetyng: also from a slack discussion of going with happy path for testnets now
it's not the ideal design, and lots of unfinished thinking/work going on there

src/client/connections/messaging.rs
oetyng: hmmm, initially yes, but eventually it will maintain connections to several nodes I think?

src/routing/core/comm.rs
oetyng: Is 1 second ok?

src/client/client_api/blob_apis.rs
```diff
@@ -148,199 +157,117 @@ impl Client {
...
+        reader: Client,
+        keys: impl Iterator<Item = ChunkKey>,
+    ) -> Vec<EncryptedChunk> {
+        let tasks = keys.map(|key| {
+            let reader = reader.clone();
+            task::spawn(async move {
+                for i in 1..11 {
+                    let r_num: u8 = rand::random();
+                    let sleep_ms = i * r_num as u32;
+                    tokio::time::sleep(Duration::from_millis(sleep_ms as u64)).await;
```
joshuef, line 229: what's this sleep about? 

maybe a comment here, or pull this out into a func for clarity about what's going on

src/routing/core/msg_handling/anti_entropy.rs
joshuef: > thought it is the main purpose of AEUpdate ?

also @maqi this is _in_ the AE update process

src/client/connections/messaging.rs
joshuef: Yeh, qp2p used to just give us one conn. And then we queried that to get relevant elders, and connected to _those_

src/client/client_api/blob_apis.rs
oetyng: Yeah, will put a comment. 
It's a bit wip, I feel it is not quite good yet. But when we used retries in the tests for example, on a request to read a blob, we would retry fetching 100 chunks when 1 failed. This here is a move to only retry individual failing chunks.

src/client/connections/messaging.rs
maqi: So, previously Session only contains one connection
but now seems need to hold a list of connections.
Hence the implementation needs to be refactorer as well, instead of using the previous code directly.

src/client/connections/messaging.rs
maqi: right.
however, still need to connect to multiple elders or nodes? instead of just one?

src/routing/core/comm.rs
maqi: shall be

src/client/client_api/blob_apis.rs
joshuef: >on a request to read a blob, we would retry fetching 100 chunks when 1 failed. This here is a move to only retry individual failing chunks.

We shouldn't. There's the client side recent blob cache that should stop refetching anything we're already received so we only actually fire our reqs for chunks we still didnt get back yet

joshuef:
> Our E2E doubles a bit as a load/perf smoke detector rn, though should ideally be its own thing.

Agreed.

I'd leave off 20mb for now (or maybe add it to the benches); it bogs test down a fair bit for me locally (more than 10mb)

src/client/client_api/blob_apis.rs
joshuef: Or you mean that those reqs may still be in flight, so we're retrying before stuff has had a chance to come in at all?

src/types/chunk.rs
```diff
@@ -7,116 +7,37 @@
...
 use std::u64;
 
 /// Maximum allowed size for a serialised Chunk to grow to.
 pub const MAX_CHUNK_SIZE_IN_BYTES: u64 = 1024 * 1024 + 10 * 1024;
 
-/// Private Chunk: an immutable chunk of data which can be deleted. Can only be fetched
-/// by the listed owner.
+/// Chunk, an immutable chunk of data
 #[derive(Hash, Eq, PartialEq, PartialOrd, Ord, Clone, custom_debug::Debug)]
-pub struct PrivateChunk {
```
maqi, line 21: just to confirm, so we now no longer to differetiate PrivateChunk and PublicChunk?

src/types/chunk.rs
```diff
@@ -137,99 +58,37 @@ impl PublicChunk {
 
     /// Returns size of this chunk after serialisation.
     pub fn serialised_size(&self) -> u64 {
-        serialized_size(self).unwrap_or(u64::MAX)
+        serialized_size(&self.value).unwrap_or(u64::MAX)
```
maqi, line 61: the self.value is Bytes, which means serialised_size can directly return with self.value.len() ?
this also elminates the chance of return with u64::MAX

src/client/client_api/blob_apis.rs
```diff
@@ -6,123 +6,136 @@
...
+use futures::future::join_all;
+use itertools::Itertools;
+use self_encryption::{self, ChunkKey, EncryptedChunk, SecretKey as BlobSecretKey};
+use std::time::Duration;
+use tokio::task;
+use tracing::trace;
+use xor_name::XorName;
+
+struct HeadChunk {
+    chunk: Chunk,
```
maqi, line 28: is `Chunk` supposed to have a `upper_limit_size` , meanwhile `HeadChunk` doesn't have an upper limit?

src/client/client_api/blob_apis.rs
```diff
@@ -6,123 +6,136 @@
...
-    /// and with which the data can be read.
-    /// It performs data storage as well as all necessary payment validation and checks against the client's AT2 actor.
-    ///
-    /// # Examples
-    ///
-    /// TODO: update once data types are crdt compliant
-    ///
-    pub async fn store_public_blob(&self, data: &[u8]) -> Result<ChunkAddress, Error> {
-        self.create_new_blob(data, true).await
+        let chunk = self.read_from_network(address.name(), false).await?;
```
maqi, line 84: what will happen the stored one is private, but here pass in an address of public with same name?

src/client/client_api/blob_apis.rs
```diff
@@ -148,199 +157,117 @@ impl Client {
...
+        });
 
-            // If the chunk (data map) is bigger than 1MB we need to break it down
-            if chunk.validate_size() {
-                break (data_map, chunk);
-            } else {
-                data = serialize(&chunk)?;
-            }
-        };
+        let _ = join_all(tasks)
```
maqi, line 172: just a remind that this could be a point of infinite hanging.
better to put a timeout bar and throw out some error, saying what was the error.
unless, the `send_cmd` just fire the cmd, and don't wait for any response?

maqi, line 172: just a remind that this could be a point of infinite hanging.
better to put a timeout bar and throw out some error, saying what was the error.
unless, the `send_cmd` just fire the cmd, and don't wait for any response?

src/client/client_api/blob_apis.rs
```diff
@@ -148,199 +157,117 @@ impl Client {
...
+    }
+
+    async fn get_chunks(
+        reader: Client,
+        keys: impl Iterator<Item = ChunkKey>,
+    ) -> Vec<EncryptedChunk> {
+        let tasks = keys.map(|key| {
+            let reader = reader.clone();
+            task::spawn(async move {
+                for i in 1..11 {
```
maqi, line 226: why 11 ?

src/client/client_api/blob_apis.rs
```diff
@@ -148,199 +157,117 @@ impl Client {
...
+                }
+                Err(Error::NoResponse)
+            })
+        });
 
-        match self_encryptor.delete().await {
-            Ok(_) => Ok(()),
-            Err(error) => Err(Error::SelfEncryption(error)),
-        }
+        join_all(tasks)
```
maqi, line 242: this is trickier than the case of when storing a chunk
reading from network is not guaranteed to get a response, so it may return with an error or hang there infinitely.
join_all and swallow all error may cause many trouble to the later on debug or normal use.

src/client/client_api/blob_apis.rs
```diff
@@ -404,318 +351,192 @@ mod tests {
...
 
     #[tokio::test(flavor = "multi_thread")]
-    async fn create_and_retrieve_index_based() -> Result<()> {
-        create_and_index_based_retrieve(1024).await
+    #[ignore = "too heavy for CI"]
+    async fn create_and_retrieve_80mb_private() -> Result<()> {
+        create_and_retrieve(80 * 1024 * 1024, Scope::Private).await
     }
 
-    async fn create_and_index_based_retrieve(size: usize) -> Result<()> {
```
maqi, line 666: do we still have this kind of `index_based_retrieve` test?
and
why we no longer needed such kind of client get api function?

src/client/client_api/blob_apis.rs
```diff
@@ -404,318 +351,192 @@ mod tests {
...
-    }
-
-    #[allow(clippy::match_wild_err_arm)]
-    async fn gen_data_then_create_and_retrieve(size: usize, public: bool) -> Result<()> {
-        let raw_data = generate_random_vector(size);
+        // the larger the file, the longer we have to wait before we start querying
+        let delay = usize::max(1, size / 2_000_000);
 
-        let client = create_test_client(Some(BLOB_TEST_QUERY_TIMEOUT)).await?;
+        println!("Delay: {} s", delay);
```
maqi, line 534: remove?

src/types/chunk.rs
oetyng: Yes, that's right.

src/types/chunk.rs
oetyng: Good catch, leftover code from before it was a `Bytes` type.

src/client/client_api/blob_apis.rs
oetyng: There is only one namespace for chunks. There are just chunks.

A Private Blob is one where the chunks containing secret key are encrypted. 
If you have a Public Blob producing a head chunk which has same name and content as a head chunk of a Private Blob, well that is just a collision.
Just as likely and unlikely as any collision within the xorspace of chunks.

src/client/client_api/blob_apis.rs
oetyng: afaik it just fires the cmd and doesn't wait for response

src/client/client_api/blob_apis.rs
oetyng: We have timeouts further in there, so there will be no hanging here.

Swallowing the errors is just a way to compact them into one. There will be an error when we compare number of chunks required and the number we actually got.

src/client/client_api/blob_apis.rs
oetyng: See `seek_in_data()`
The same thing.

src/client/client_api/blob_apis.rs
oetyng: It's just a number of retries
This piece of code was commented by Josh as well, more details here: https://github.com/maidsafe/safe_network/pull/373#discussion_r704340053

src/client/client_api/blob_apis.rs
oetyng: afaik it just fires the cmd and doesn't wait for response

src/client/client_api/blob_apis.rs
oetyng: No. A `HeadChunk` is just a client side convenience wrapper packing a chunk (with `upper_limit_size`) and a BlobAddress - effectively indicating if the chunk content is encrypted or not (i.e. private or public).

src/client/connections/messaging.rs
oetyng: It does, eventually. It starts with one, and then continues with the rest as it gets the required details.

src/client/client_api/blob_apis.rs
oetyng: This code is now removed since blob cache was handling this situation. Will bring that work over to its own PR instead.

joshuef (approved):

PR #376, opened 2021-09-07
feat: start using the genesis key from clients and joining nodes

bochaco:


src/client/connections/messaging.rs
```diff
@@ -67,18 +67,18 @@ impl Session {
 
         // *****************************************************
         // FIXME: receive the network's genesis pk from the user
-        let genesis_pk = bls::SecretKey::random().public_key();
+        let genesis_key = bls::SecretKey::random().public_key();
```
connec, line 70: Would we have to fix this for actual users to connect, e.g. for a public testnet?

src/node/config_handler.rs
```diff
@@ -382,26 +387,32 @@ impl Config {
...
-    let bootstrap_nodes = if let Ok(mut bootstrap_nodes) = read_conn_info_from_file().await {
+    let info = if let Ok((genesis_key_hex, mut bootstrap_nodes)) = read_conn_info_from_file().await
+    {
         let _ = bootstrap_nodes.insert(contact);
-        bootstrap_nodes
+        (genesis_key_hex, bootstrap_nodes)
     } else {
-        vec![contact].into_iter().collect()
+        warn!(
+            "Connection info file being created at {} but no genesis key is being set.",
```
connec, line 405: Should we either:

- Take a `genesis_key` arg and error if it doesn't match what's in the file (conflicting networks?).
- Error if there's no file.
- Early `return Ok(())`.

Instead of writing a broken file? Even the last one seems fine, and doesn't need any new errors etc.

src/node/node_api/mod.rs
```diff
@@ -137,6 +137,11 @@ impl Node {
         self.network_api.our_prefix().await
     }
 
+    /// Returns the network;s genesis key.
```
connec, line 140: Nitpick:

```suggestion
    /// Returns the network's genesis key.
```

src/routing/core/msg_handling/join.rs
```diff
@@ -94,16 +94,19 @@ impl Core {
...
-                    "Ignoring JoinRequest from {} - first-section node having wrong age {:?}",
-                    peer,
-                    peer.age(),
-                );
-                return Ok(vec![]);
+                let node_msg = SystemMsg::JoinResponse(Box::new(JoinResponse::Retry(
+                    self.section.authority_provider().clone(),
+                )));
+                trace!("New node in first section should join with age greater than MIN_ADULT_AGE. Sending {:?} to {}", node_msg, peer);
+                return Ok(vec![self.send_direct_message(
```
connec, line 105: Do we need to change this? Seems fine generally, but I'd assumed the idea is that only a broken/malicious node would try to join with and invalid age, and so the recipient should just ignore them (otherwise they can be spammed and have to do lots of signing etc. for their responses).

connec (changes requested):
Left a couple of small queries and suggestions.

src/node/config_handler.rs
bochaco: I was hesitating what to do here, I was thinking that perhaps we go for returning an Error, the user of this api needs to make sure it's calling this after it set the genesis key, it can even handle the error and call `set_connection_info` before retrying to call this function...?

src/client/connections/messaging.rs
bochaco: oops...nice catch! I'll be taking care of this

src/routing/core/msg_handling/join.rs
bochaco: The issue is that we have this special thing about age with first section, so a joining node doesn't know up front if it's a first section or not, so it needs a `Retry` or any response with the current prefix so it can generate a name and age accordingly. Prevent spam attacks always come up to my head when I'm coding msg handlers, but it sounds like even we can reduce the effect by ignoring some type of msgs (which I agree we should do), there is always an easy way to spam a node, e.g. in this case I can spam it by sending JoinReqs with all valid data/content after I'm part of a section, this will be rejected only when hitting the verification in L68 further down....so....

connec: Good points! Makes sense to me.

connec (dismissed):

connec (approved):

joshuef (dismissed):

connec (approved):

PR #389, opened 2021-09-09
fix: add missing await for liveness tracking.

joshuef:
Some log message tweaks too


grumbach (approved):

src/routing/core/chunk_records.rs
```diff
@@ -128,7 +132,8 @@ impl Core {
         for target in targets {
             let _ = self
```
bochaco, line 133: I think we don't need the `let _ =` since that function doesn't return a value

bochaco (changes requested):

bochaco (dismissed):

bochaco (approved):

PR #391, opened 2021-09-09
Simplify node address config

connec:
- 7e2643254 **chore!: remove unused `node::Config::bootstrap_cache_dir` field**

  This was used to configure the 'bootstrap cache' in qp2p, which no
  longer exists. Apart from tests and config merging, the fields is no
  longer used.
  
  BREAKING CHANGE: The `bootstrap_cache_dir` field of `node::Config` has
  been removed.

- 280b1fd31 **refactor!: simplify node address config and improve validation**

  Node address configuration has been simplified to just two CLI options:
  
  - `--local-addr` to set the local address that will be bound. This still
    defaults to `0.0.0.0:0` (any IP, random port).
  - `--public-addr` to set the public address that will be written in
    connection info files/logs. If unspecified, this will be resolved by
    qp2p by querying from a contact, if any, or falling back to the
    resolved local address once bound.
  
  In particular, `--first` is now a flag, and no longer takes an address
  value.
  
  Validation of address parameters has also changed:
  
  - Public addresses must be concrete (e.g. not `0.0.0.0` or port `0`) and
    non-loopback (e.g. not `127.0.0.1`).
  - If the local address uses a loopback IP (e.g. `127.0.0.1`), a public
    address cannot be specified (since the node will not be public).
  - If `--first` is set, then either a public address, or a local address
    with a concrete IP, must be specified.
  
  Ultimately this should make the address configuration easier to
  understand, and make it harder to start an unreachable node.
  
  BREAKING CHANGE: At the API level, the `node::Config::first` field has
  changed from `Option<SocketAddr>` to `boolean`. This also affects the
  `sn_node` CLI, where `--first` is no longer allowed a value.
  Additionally, the semantics and validation of `--local-addr` and
  `--public-addr` have changed, however this should only affect
  nonsensical combinations.


joshuef (approved):
nice. much clearer

joshuef (approved):

PR #406, opened 2021-09-13
Back pressure

oetyng:
Implements the `Backpressure` module which is used to apply a
retry-policy per node, based on their reported cpu load over time.

***

This uses the cpu load avg of `one`, `five` and `fifteen` minutes (as gotten by the `sysinfo` crate that we already use).

The idea is to have nodes check their cpu load (if a minimum amount of time has passed since last), when taking in requests, and if it is above a few thresholds (currently hard coded) it will return an error with the cpu load in it.

Receiving node can then map this to a retry policy which it stores for the socketaddr.

***

Step by step:

1. On incoming msgs, the node checks whether it is under strain.
2. If it is, it will use the `SystemMsg::Backpressure` variant to tell msg producers to slow down.
3. Included in the msg is the `LoadReport`, which contains the cpu load avg during a few time spans.
4. At receival of the backpressure msg at the original msg producer, the comms to the strained node are regulated proportionally to its reported cpu load.


src/messaging/system/mod.rs
```diff
@@ -158,3 +162,15 @@ pub enum SystemMsg {
         correlation_id: MessageId,
     },
 }
+
+/// Average cpu load to be sent over the wire.
+/// The values represent percentages, e.g. 12.234.., 21.721.., etc.
```
connec, line 167: Have you confirmed if this is true? I'm suspicious, since "load average" typically means CPU queue depth on Linux systems.

Looking at [the implementation of `sysinfo`'s `load_average`](https://docs.rs/sysinfo/0.20.3/src/sysinfo/linux/system.rs.html#474-493) it suggests the info is taken straight from `/proc/loadavg` on Linux:

> The first three fields in this file are load average figures giving the number of jobs in the run queue

On Mac it will use [`getloadavg`](https://man7.org/linux/man-pages/man3/getloadavg.3.html):

> The getloadavg() function returns the number of processes in the system run queue averaged over various periods of time.

It seems Windows doesn't have a high level API, so sysinfo has its own fairly involved [implementation](https://github.com/GuillaumeGomez/sysinfo/blob/master/src/windows/processor.rs), but it seems to be reading from `\System\Processor Queue Length`:

> The Processor Queue Length is the number of threads that are ready but currently unable to run on the processor due to another active thread

src/routing/core/msg_handling/mod.rs
```diff
@@ -316,6 +336,11 @@ impl Core {
                 )
                 .await
             }
+            SystemMsg::BackPressure(load_report) => {
+                trace!("Handling msg: BackPressure from {}", sender);
+                self.comm.regulate(sender, load_report).await;
```
joshuef, line 341:
```suggestion
                // TODO: Factor in med/long term backpressure into general node liveness calculations
                self.comm.regulate(sender, load_report).await;
```

src/messaging/system/mod.rs
oetyng: Yes I have confirmed it to be true. /joke :D

Your lookups show it is queue length.

No biggie, just another evaluation of the value then. (Or find crate that measures cpu load).

connec: Sorry, I was more cautious incase I was misreading something 😄 

I think getting a value normalised into 0-1 would be best for making balancing decisions. CPU utilisation would be easiest, or I think you could do 'load average average' (e.g. load average / num cores – referred to as "normalized load average" by some articles). This lets you test the general rule that queue depth should not be > number of cores. So if normalized load average is > 1 the node would have some queued work, close to one would be approaching capacity, etc.

connec: [This](https://blog.appsignal.com/2018/03/28/understanding-system-load-and-load-averages.html) article has a nice summary:

> # Reading load averages
>
> Since it’s an average of the absolute number of processes, it can seem difficult to determine what a proper load average is, and when to be alarmed. In general, since each of your CPU cores can handle one process at a time, the system isn’t overloaded until it goes over 1.0 per logical processors. On a dual-core system (without hyper-threading), that’d be 2.0.
>
> Usually, it’s fine if the load average is above 1.0 per core in the last minute mark, but elevated load in the five or fifteen-minute averages could indicate a problem.

oetyng: Ah I missed this. Yes this is what I ended up doing :+1:

Yoga07 (approved):

PR #426, opened 2021-09-15
feat(client): introduce caching of AE resends at client

Yoga07:


src/client/connections/listeners.rs
```diff
@@ -279,7 +279,61 @@ impl Session {
...
+                        other
+                    );
+                    return Ok(session);
+                }
+            };
+
+        let (targets, dst_address_of_bounced_msg) = match &service_msg {
+            ServiceMsg::Cmd(cmd) => {
+                match &cmd {
+                    DataCmd::StoreChunk(_) => (3, cmd.dst_name()), // stored at Adults, so only 1 correctly functioning Elder need to relay
```
oetyng, line 307: As a reminder: We probably should make those changes mentioned where these values were initially introduced.
Will be good as a follow up.

oetyng (approved):

oetyng:
Just some clippy to run.

PR #438, opened 2021-09-16
test(ae): dont pass rand xorname to entropy check

oetyng:


Yoga07:
Superseded by https://github.com/maidsafe/safe_network/pull/440

PR #458, opened 2021-09-18
feat(data): distinguish between spot and blob

oetyng:
A `Blob` is at least 3072 bytes, as that is min self encryptable bytes.
A `Spot` is smaller than that, and will be encrypted with the client
encryption keys.
Both of these are type safe in that the construction of them will
validate their size.

### DIFF
--- a/src/client/client_api/blob_apis.rs
+++ b/src/client/client_api/blob_apis.rs
@@ -6,7 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{data::get_data_chunks, Client};
+use super::data::{to_chunk, Blob, BlobAddress, Spot, SpotAddress};
+use super::{data::encrypt_blob, Client};
 use crate::messaging::data::{DataCmd, DataQuery, QueryResponse};
 use crate::types::{Chunk, ChunkAddress, Encryption};
 use crate::{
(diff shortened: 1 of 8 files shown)


src/client/client_api/blob_apis.rs
```diff
@@ -136,12 +104,42 @@ impl Client {
...
+        encrypt_blob(blob.bytes(), owner.as_ref())
+    }
+
+    /// Encrypts a small piece of t(d)ata (spot) and returns the resulting address and the chunk.
+    /// Does not store anything to the network.
+    pub fn encrypt_spot(&self, spot: Spot, scope: Scope) -> Result<(SpotAddress, Chunk)> {
+        let encryption = encryption(scope, self.public_key());
+        let chunk = to_chunk(spot.bytes(), encryption.as_ref())?;
+        if chunk.value().len() >= self_encryption::MIN_ENCRYPTABLE_BYTES {
+            return Err(Error::Generic("You might need to pad the spot contents, as the encryption of it made it slightly too big".to_string()));
```
joshuef, line 120: you might need to pay (aka add more bytes?) as the encryption made it _too big_?

I'm confused why more bytes gets us something smaller here. Maybe just the msg needs a tweak though

src/client/client_api/blob_apis.rs
```diff
@@ -136,12 +104,42 @@ impl Client {
...
+    pub fn encrypt_blob(&self, blob: Blob, scope: Scope) -> Result<(BlobAddress, Vec<Chunk>)> {
         let owner = encryption(scope, self.public_key());
-        let (head_address, all_chunks) = get_data_chunks(data, owner.as_ref())?;
+        encrypt_blob(blob.bytes(), owner.as_ref())
+    }
+
+    /// Encrypts a small piece of t(d)ata (spot) and returns the resulting address and the chunk.
+    /// Does not store anything to the network.
+    pub fn encrypt_spot(&self, spot: Spot, scope: Scope) -> Result<(SpotAddress, Chunk)> {
+        let encryption = encryption(scope, self.public_key());
```
joshuef, line 117: When I read this line, I think this is always going to encrypt.

I code w/ rust_analyzer, I can see it's an option. But skimming it's not always clear (it took me a bit to see that `None` here is how we _dont_ encrypt public scope stuff.

Anyway. I'm basically wondering if there's a name tweak here that can help make it clear (and some docs on the `encryption` func)

Perhaps it just being `optional_encryption` or something? Not really, sure but just wanted to highlight something that was tripping me up here

src/client/client_api/blob_apis.rs
oetyng: If encrypting it made it slightly too big, say it was `3071` and become `3072` bytes (idk if that can happen but I suppose so), then unless they pad it they won't be able to store it.
So if they pad it with a byte, they can store it as a `Blob`.

src/client/client_api/blob_apis.rs
oetyng: Ah, this was simply from copy passing `encrypt_blob`, which will in fact always (self)encrypt it.

But with `Spots` that is only if it's private, so another name definitely needed.

joshuef (approved):

PR #460, opened 2021-09-20
fix: avoid adding root_dir into used_space

maqi:

### DIFF
--- a/src/dbs/kv_store/used_space.rs
+++ b/src/dbs/kv_store/used_space.rs
@@ -10,6 +10,7 @@ use super::Error;
 use async_recursion::async_recursion;
 use dashmap::DashSet;
 use futures::future::join_all;
+use std::collections::HashSet;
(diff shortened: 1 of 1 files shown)


src/dbs/kv_store/used_space.rs
```diff
@@ -31,7 +32,38 @@ impl UsedSpace {
...
+        if self.dirs.iter().any(|dir| {
+            let cur_dir_str = if let Some(str) = dir.to_str() {
+                str
+            } else {
+                return false;
+            };
+            ancestor_strs.contains(cur_dir_str)
+        }) {
+            // Ancestor exists
+            trace!("Path {:?} is being descendant of existing entry.", dir);
```
joshuef, line 56:
```suggestion
            trace!("Path {:?} is a descendant of an existing entry, it will not be added for storage calcuation checks.", dir);
```

joshuef (approved):

PR #464, opened 2021-09-20
feat: accept empty url scheme

grumbach:
- accept empty url scheme in the SAFE Url parser
- free `cli` and `api` from the `"safe://"` prepending shenanigans

### DIFF
--- a/src/url/url_parts.rs
+++ b/src/url/url_parts.rs
@@ -94,7 +94,7 @@ impl UrlParts {
 
         // Validate the url scheme is 'safe' if it is not empty
         let scheme = parsing_url.scheme().to_string();
-        if scheme != "" && scheme != URL_SCHEME {
+        if !scheme.is_empty() && scheme != URL_SCHEME {
             let msg = format!("invalid scheme: '{}'. expected: '{}'", scheme, URL_SCHEME);
             return Err(Error::InvalidXorUrl(msg));
         }
--- a/src/url/url_parts.rs
+++ b/src/url/url_parts.rs
@@ -92,9 +92,9 @@ impl UrlParts {
             Error::InvalidXorUrl(msg)
         })?;
 
-        // Validate the url scheme is 'safe'
+        // Validate the url scheme is 'safe' if it is not empty
         let scheme = parsing_url.scheme().to_string();
-        if scheme != URL_SCHEME {
+        if scheme != "" && scheme != URL_SCHEME {
             let msg = format!("invalid scheme: '{}'. expected: '{}'", scheme, URL_SCHEME);
             return Err(Error::InvalidXorUrl(msg));
         }


bochaco (approved):

dan-da:
What is the rationale for that?    to me, a URL without a scheme is very clearly an invalid URL.

If you want to represent a "URL-without-scheme" I would suggest a new data struct for that purpose.

At the least, I think this PR needs more discussion.  What is the motivation/objective?

bochaco:
Basically, this is a SAFE Url API and utilities, thus, if the provided string doesn't specify a scheme it shall be assumed to be safe://.

dan-da:
For the record, I do not (yet) agree with the intent of this change and I do not think it should have been merged without further discussion/concensus given my previous objection.

A few points:

1. I'm not so sure this code is working as intended.  See (4).

2. As intended, this is not only an internal optimization/cleanup.  It is a public API change, and a URL is an entity that has well defined standards and meaning outside of SN.  As such, I think we need to consider the change carefully.

3. If the input string has no scheme, how do we distinguish between a relative url and an absolute url?   Let's say:  Url::from("mystuff").      Does mystuff represent an NRS name, or does it represent path "mystuff" in a relative URL?

For similar reasons, Url::url only allows absolute URLs.  See [docs](https://docs.rs/url/2.2.2/url/#base-url) as well as this [issue](https://github.com/servo/rust-url/issues/487).   If I write:  ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7bd5668c55093619bca81e27940c1409))

```
    let r = Url::parse("mystuff");
    println!("url: {}", r.unwrap());
```
Output is:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: RelativeUrlWithoutBase', src/main.rs:7:27
```

Go developers also [take this position](https://github.com/golang/go/issues/12585), quote:

> Not a bug. url.Parse parses URLs. That's not a URL. A URL starts with a scheme.

**There may be other consequences I/we do not see yet.**  Url::url follows a standard, and I don't see that we should deviate from it for trivial reasons.

4. Is this code even working as intended?   It does this:
```
        let parsing_url = RustUrl::parse(url).map_err(...)?;
￼        let scheme = parsing_url.scheme().to_string();
        if !scheme.is_empty() && scheme != URL_SCHEME {
```

As far as I can tell, RustUrl::parse() will also return an error if input url does not have a scheme.   So your is_empty() check never gets reached.

5.  has the code been tested? I don't see any test cases for it.   But I do see this *unchanged* test case.

```
        let result = Url::from_url("withoutscheme");
        verify_expected_result!(result, Err(Error::InvalidXorUrl(err)) if err.contains("relative URL without a base"))?;
```

So if that is still passing, then I think it backs up my assertion in (4).

6. Similar/same objective of cleaning up internal code could be achieved with a helper struct that prepends "safe://" or possibly by having a special "::from_schemeless()" fn with well defined behavior for prepending safe://.   That would not affect the existing from_nrsurl, from xorurl, etc.

bochaco:
>  I do not think it should have been merged without further discussion/concensus given my previous objection

It hasn't been merged, it was closed due to further discussions that took place.

dan-da:
my mistake then.   I'll close it again.  thx.

PR #470, opened 2021-09-21
ci: ignore `safe_network` version in cache key

connec:
- bd1110075 **ci: use a single `paths-filter` step**

  `dorny/paths-filter` allows multiple filters to be specified with
  different keys. We can use this, rather than multiple steps, to avoid
  the overhead of additional checkouts and step setup/teardown.

- 2a4f5544c **ci: ignore `safe_network` version in cache key**

  This should allow us to reuse compiled dependencies across
  `safe_network` releases. This could be quite impactful since every
  merged PR generates a new release, and PR are tested merged (meaning
  `Cargo.lock` is likely to update).

### DIFF
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -84,6 +84,15 @@ jobs:
           toolchain: stable
           override: true
 
+      # Remove safe_network version from Cargo.lock to maintain cache across versions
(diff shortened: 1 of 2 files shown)


connec:
Gonna close this, since the results were pretty mixed anyway, and the CI workflows have changed enough that rebasing would be a major pain.

PR #485, opened 2021-09-23
test: retry register delete while it still exists

joshuef:
We should attempt to delete the register again if it still exists, otherwise a dropped message can cause the `can_delete_private_register` test to fail


src/client/client_api/register_apis.rs
```diff
@@ -547,6 +547,9 @@ mod tests {
         let mut res = client.get_register(address).await;
         while res.is_ok() {
             tokio::time::sleep(tokio::time::Duration::from_millis(200)).await;
+            client.delete_register(address).await?;
+            // attempt to delete register again (perhaps a message was dropped)
+            tokio::time::sleep(tokio::time::Duration::from_millis(400)).await;
```
connec, line 552: Is it necessary to wait twice? Since it's to the same address, it seems reasonable to:

```rust
while res.is_ok() {
    client.delete_register(address).await?;
    tokio::time::sleep(tokio::time::Duration::from_millis(400)).await;
    res = client.get_register(address).await;
}
```

Particularly since `client.get_register` has a built-in timeout.

connec (dismissed):

src/client/client_api/register_apis.rs
joshuef: fair

connec (approved):
Bellissimo 😗👌

connec (approved):

PR #488, opened 2021-09-24
feat(routing): introduce probing messages

Yoga07:


src/routing/core/msg_handling/mod.rs
```diff
@@ -327,6 +328,10 @@ impl Core {
                 )
                 .await
             }
+            SystemMsg::ProbeMessage(_dst) => {
```
joshuef, line 331: @Yoga07 I'd perhaps rename this, as it's a bit confusing what this is just from the name.

Perhaps `AE-ProbeForNewKeys` ? And we send along a last known key there? (so as not to get full genesis->now update every time?

Yoga07: Yea, we attach the latest key we know for that message's section, so we'll get back only a truncated chain, but enough to verify the incoming deets.

maqi (approved):

PR #489, opened 2021-09-24
Feat enable network log asserts

joshuef:
Idea here is basically to provide some helper for grepping the testnet logs for certain things.

### DIFF
--- a/.github/workflows/pr_tests.yml
+++ b/.github/workflows/pr_tests.yml
@@ -196,6 +196,16 @@ jobs:
       - name: Extra wait on CI for split organisation
         run: sleep 60
 
+      - name: Run network asserts one by one
+        shell: bash
+        # FIXME: windows has an issue w/ locking of log files
+        if: matrix.os != 'windows-latest'
+        # here test-threads=1 is important so we dont pollute log counts by running tests in parallel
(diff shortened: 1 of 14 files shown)


.github/workflows/pr_tests.yml
```diff
@@ -208,23 +208,29 @@ jobs:
...
+      #     killall sn_node
+      #     rm -rf ~/.safe/node
+      #   shell: bash
+
+      # - name: Kill the current network (not needed for next test)
+      #   if: matrix.os == 'windows-latest'
+      #   run: |
+      #     taskkill /IM sn_node.exe /F
+      #     rd /s /q %USERPROFILE%\.safe\node
+      #   shell: cmd
```
connec, line 233: How come these are commented out?

connec (commented):
LGTM, one Q about some commented out stuff in the test workflow (also clippy ain't happy and some tests are failing, can't easily see which/why as the logs are too long).

connec:
In fact, MacOS is failing a new assertion: https://github.com/maidsafe/safe_network/pull/489/checks?check_run_id=3723267210

.github/workflows/pr_tests.yml
joshuef: Ah, cos the next test is commented out still also. I'll leave that out actually to tidyup the pr

joshuef:
seems like there's a bug and we're still resending more on some AE instances than we'd want 🕵️

PR #491, opened 2021-09-24
chore: Update `qp2p`

connec:
- 5dab62225 **chore: update `qp2p`**

  The new version removes the `Endpoint::try_send_*` methods, and the
  `Incoming*` streams yield `Connection`s rather than `SocketAddr`. Simple
  workarounds exist in both cases.


lionel1704 (approved):
Looking good! :100:

lionel1704 (approved):

PR #505, opened 2021-09-30
fix(client): send a message to probe the network initially

Yoga07:
This triggers an AE flow for the client during creation to get a basic knowledge(at least one valid SAP) of the network before it begins to serve requests from its user.


joshuef (approved):
For a simple solution this is fine. 

Long term we want the client to be able to send proper AE-Probe messages though, so perhaps that's a good next target?

joshuef (approved):

PR #507, opened 2021-09-30
feat: add spot entry to datatype enum

jacderida:
BREAKING CHANGE:
This is to facilitate the new way that files are stored on the network, which was updated to support changes in self encryption.

Self encryption doesn't allow encryption of files less than 3072 bytes in size, so a new type of data was created to support storage of smaller files. This data type is known as a 'spot'.

We need to distinguish between blobs and spots when retrieving files, and the solution is for the DataType on XorUrls to be marked as the Spot type when they are saved (which will be done in the API).

The `encode_blob` and `encode_spot` functions are both introduced for convenience. These accept a BlobAddress and a SpotAddress, respectively. Added a couple of extra tests for checking that `encode_spot` was setting fields correctly.

### DIFF
--- a/examples/client_blob.rs
+++ b/examples/client_blob.rs
@@ -37,12 +37,7 @@ async fn main() -> Result<()> {
     println!("Storing data.. ({} bytes)", blob.bytes().len());
 
     let address = client.write_blob_to_network(blob, Scope::Public).await?;
-    let xorurl = Url::encode_blob(
-        *address.name(),
-        Scope::Public,
-        ContentType::Raw,
-        DEFAULT_XORURL_BASE,
-    )?;
+    let xorurl = Url::encode_blob(address, ContentType::Raw, DEFAULT_XORURL_BASE)?;
     println!("Blob stored at xorurl: {}", xorurl);
 
     let delay = 5;
--- a/examples/network_split.rs
+++ b/examples/network_split.rs
@@ -237,12 +237,7 @@ async fn upload_data() -> Result<(BlobAddress, [u8; 32])> {
     println!("Storing data w/ hash {:?}", output);
 
     let address = client.write_blob_to_network(blob, Scope::Public).await?;
-    let xorurl = Url::encode_blob(
-        *address.name(),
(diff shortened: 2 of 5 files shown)


src/types/register/mod.rs
```diff
@@ -1165,6 +1165,7 @@ mod tests {
         let url = Url::encode_blob(
             xor_name,
             Scope::Public,
+            DataType::Blob,
```
oetyng, line 1168: `XorName`, `Scope` and `DataType` params can be replaced with `BlobAddress`
Then add `encode_spot`, taking `SpotAddress`.

jacderida: Thanks for the feedback. I amended the commit so that we now have 2 different functions.

joshuef (dismissed):

examples/client_blob.rs
```diff
@@ -38,8 +40,7 @@ async fn main() -> Result<()> {
 
     let address = client.write_blob_to_network(blob, Scope::Public).await?;
     let xorurl = Url::encode_blob(
-        *address.name(),
-        Scope::Public,
+        BlobAddress::Public(*address.name()),
```
oetyng, line 43: You can pass the address directly from the write fn.

examples/network_split.rs
```diff
@@ -238,8 +238,7 @@ async fn upload_data() -> Result<(BlobAddress, [u8; 32])> {
 
     let address = client.write_blob_to_network(blob, Scope::Public).await?;
     let xorurl = Url::encode_blob(
-        *address.name(),
-        address.scope(),
+        BlobAddress::Public(*address.name()),
```
oetyng, line 241: Same here, can pass address directly.

jacderida: I think you were suggesting doing this?
```
BlobAddress::Public(client.write_blob_to_network(blob, Scope::Public).await?),
```

However, in both cases, the `address` variable gets used later in the code. If you're only around intermittently, I think I'll leave this one for now.

PR #522, opened 2021-10-05
fix: send proposal to proper dst_location

maqi:


src/routing/core/messaging.rs
```diff
@@ -73,10 +73,14 @@ impl Core {
             sig_share,
         };
         let section_pk = *self.section.chain().last_key();
+        let dst_name = self
```
joshuef, line 76: can you add a note as to why this is necessary? it isn't immediately obvious to me why we're doing this here.

joshuef (dismissed):

Yoga07 (approved):

PR #527, opened 2021-10-05
draft: trace logs to detect routing internal commands handling

bochaco:


src/routing/routing_api/tests/mod.rs
```diff
@@ -1078,9 +1078,9 @@ async fn get_internal_commands(
 async fn relocation_of_non_elder() -> Result<()> {
     relocation(RelocatedPeerRole::NonElder).await
 }
-
+*/
 const THRESHOLD: usize = supermajority(ELDER_SIZE) - 1;
-
+/*
```
joshuef, line 1083: should we just remove these comments?

PR #538, opened 2021-10-06
Fix remove ae msgs loop

bochaco:


joshuef (approved):

joshuef (dismissed):

src/routing/core/messaging.rs
```diff
@@ -94,6 +94,42 @@ impl Core {
     // ------------------------------------------------------------------------------------------------------------
     // ------------------------------------------------------------------------------------------------------------
 
+    /// Generate message to update a peer with our current section chain
```
maqi, line 97: Better changed the comment to:
Generate AntiEntropyUpdate message to update a peer with proof_chain, and members_info if required.

src/routing/core/msg_handling/mod.rs
```diff
@@ -143,11 +136,11 @@ impl Core {
                                     cmds.push(ae_command);
                                     return Ok(cmds);
                                 }
+
+                                info!("Entropy check passed. Handling verified msg {:?}", msg_id);
```
maqi, line 140: better put a similar comment within the other branch (Line 118). something like:
 info!("Entropy check skipped. Handling message {:?} directly", msg_id);

src/routing/core/msg_handling/mod.rs
```diff
@@ -475,61 +475,56 @@ impl Core {
...
-                        if !self
-                            .section
-                            .chain()
-                            .has_key(&sig_share.public_key_set.public_key())
-                        {
-                            let cmd =
-                                self.handle_untrusted_message(sender, node_msg, msg_authority)?;
-                            return Ok(vec![cmd]);
-                        }
+                    // TODO: should be able to remove the sig_share from the Propose msg
```
maqi, line 516: better to put some word in to explain why can remove, for example:
as the sig_share can be carried within the msg_kind header.

bochaco: sounds good, I'm working on PR #546 for this already.

PR #549, opened 2021-10-08
fix(chunks): do not double-encrypt in packing step

oetyng:
Encryption is an option which was passed in when wrapping
bytes in `Chunk`, also when the bytes were self encrypted..
So, since no need to encrypt what is self-encrypted, we pass in
`None` for those cases
Also now propagates any error when packing data map.

### DIFF
--- a/src/client/client_api/data/pac_man.rs
+++ b/src/client/client_api/data/pac_man.rs
@@ -93,7 +93,7 @@ pub(crate) fn pack(
                 .collect();
             if expected_total > chunks.len() {
                 // as we flatten above, we need to check outcome here
-                return Err(Error::Generic("Not all data was chunked!".to_string()));
+                return Err(Error::NotAllDataWasChunked(expected_total, chunks.len()));
             }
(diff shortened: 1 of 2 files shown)


src/client/client_api/data/pac_man.rs
```diff
@@ -78,23 +84,34 @@ pub(crate) fn pack(
...
                 .par_iter()
-                .map(|c| to_chunk(c.content.clone(), encryption))
+                .map(|c| to_chunk(c.content.clone(), no_encryption)) // no need to encrypt what is self-encrypted
                 .flatten()
                 .chain(chunks)
                 .collect();
-            chunk_content = pack_data_map(DataMapLevel::Additional(data_map), encryption)?;
+            if expected_total > chunks.len() {
+                // as we flatten above, we need to check outcome here
+                return Err(Error::Generic("Not all data was chunked!".to_string()));
```
joshuef, line 96: Can we add a specific error case for this?

joshuef (approved):

PR #619, opened 2021-10-20
Chore logs inspector use one regex

joshuef:


bochaco (dismissed):

src/bin/log_cmds_inspector.rs
```diff
@@ -273,21 +380,33 @@ fn inspect_log_files(args: &CmdArgs) -> Result<BTreeMap<String, Vec<String>>> {
...
+    }) {
+        let logfile = logfile.display().to_string();
+
         let root_cmd_id = get_root_cmd_id(cmd_id);
-        if commands
-            .started
+
+        if started
             .get(&root_cmd_id)
             .map(|subcmds| subcmds.cmds_logs.get(cmd_id))
```
connec, line 411: Not your change, but this looks buggy or redundant. The `is_none` applies to the outer `Option` returned by `starts.get(&root_cmd_id)` – the `Option` returned by `map` is not tested. I suspect it should be `and_then` instead?

src/testnet_grep/mod.rs
```diff
@@ -135,3 +150,73 @@ pub(crate) fn search_testnet(pattern: &LogMarker) -> Result<Matches, Error> {
...
+            if !dent.file_type().is_file() {
+                continue;
+            }
+
+            Searcher::new().search_path(
+                &matcher,
+                dent.path(),
+                UTF8(|_lnum, _line| {
+                    count += 1;
+                    // matches.push((lnum, line.to_string(), dent.path().to_path_buf()));
```
connec, line 214: Leftover comment?

src/bin/log_cmds_inspector.rs
joshuef: aha. yeh and_then applies to both, true

PR #656, opened 2021-10-28
fix: nrs get with versions, nrs_map always returned

grumbach:
Couple of NRS fixes and API changes for `nrs_get`
- now manages versions
- `nrs_get` returns the nrs_map as well
- the `nrs_map` in `SafeData` is not an option anymore
- fixed a data race possibility in the nrs resolver handler


joshuef:
@grumbach I think this is a breaking change, no?

PR #662, opened 2021-10-28
Cache PrefixMap for Client and read from disk on restart

Yoga07:


sn/src/client/connections/messaging.rs
```diff
@@ -59,6 +63,24 @@ impl Session {
 
         let endpoint = Endpoint::new_client(local_addr, qp2p_config)?;
 
+        let mut root_dir = dirs_next::home_dir()
```
joshuef, line 66: let's split this out into a separate func to call here

joshuef (dismissed):

PR #667, opened 2021-10-30
fix: ae blob tests dont have standard wait

joshuef:


sn/src/routing/core/msg_handling/agreement.rs
```diff
@@ -273,28 +275,51 @@ impl Core {
...
+                    )?;
+
+                    if !network_updated {
+                        warn!(">>>>> Section Chain updated, but network not!??!?!?!!!!!");
+                    }
+
+                // }
             } else {
+
+                let mut section_chain_clone = self.section_chain().await.clone();
```
Yoga07, line 300: We should take a copy of the chain before we begin updating. Otherwise, if our prefix gets updated first, the head of the section chain would be our section's and we'd be inserting neighboring section chain head atop ours.

So it's ideal to use old_chain here from L#263.

sn/src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -364,6 +383,12 @@ impl Core {
                 let section_signed = section_signed_auth.sig;
                 let bounced_msg = original_bytes;
 
+                // if proof_chain.last_key() != &signed_section_auth.value.public_key_set.public_key() {
+
+                // assert!(proof_chain.last_key(), );
+
+                trace!(">>>>>  proof chain has our sign key?: {:?}", proof_chain.has_key(&section_auth.public_key_set.public_key()));
```
Yoga07, line 390: Do we need these logs and the commented-out codes?

sn/src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -43,6 +43,13 @@ impl Core {
             sig: section_signed,
         };
 
+        // FIXME: perhaps we should update our section chain
+        // only upon successful DKG round we participated on??
+        self.section
+            .merge_chain(&signed_section_auth, proof_chain.clone())
```
Yoga07, line 49: Would be better to have a check if incoming SAP is our's/

sn/src/routing/core/msg_handling/agreement.rs
joshuef: aye sound. im actually pulling in gab's tweak now which needs that too

sn/src/routing/core/msg_handling/agreement.rs
```diff
@@ -271,29 +276,55 @@ impl Core {
...
+                        section_auth
+                    );
+                }
             } else {
-                // FIXME: we shouldn't be updating our section info since it
-                // doesn't match our prefix???
-                if let Err(err) = self.section.chain.write().await.insert(
+                let mut section_chain_clone = self.section_chain().await.clone();
+
+                if let Err(err) = section_chain_clone.insert(
```
Yoga07, line 299: We can remove this insert part now since it's being covered below.

sn/src/prefix_map/mod.rs
```diff
@@ -129,17 +129,28 @@ impl NetworkPrefixMap {
...
 
-        self.verify_with_chain_and_update(
+        let res = self.verify_with_chain_and_update(
             signed_section_auth,
             proof_chain,
             &SecuredLinkedList::new(section_key),
-        )
+        );
+
+        for section in self.sections.iter() {
```
maqi, line 148:
```rust
for prefix in self.sections.keys() {
    trace!("Known prefix: {:?}", prefix);
}
```

sn/src/routing/core/msg_handling/agreement.rs
```diff
@@ -271,29 +276,45 @@ impl Core {
...
-                    .update_elders(section_auth.clone(), key_sig)
+                    .update_elders(section_auth.clone(), key_sig.clone())
                     .await;
+
+                let proof_chain = self.section_chain().await;
+
+                let network_updated = self.network.verify_with_chain_and_update(
+                    signed_section_auth.clone(),
+                    &proof_chain,
+                    &proof_chain,
```
maqi, line 287: passing the same chain twice?
bit weird here.
even an empty chain for one of it will be better? if it does being the case.

sn/src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -43,6 +43,16 @@ impl Core {
             sig: section_signed,
         };
 
+        let our_prefix = self.section().prefix().await;
+
+        // FIXME: perhaps we should update our section chain
+        // only upon successful DKG round we participated on??
+        if section_auth.prefix.is_compatible(&our_prefix) {
```
maqi, line 50: in case we are now in prefix(0) or prefix(1) , and the incoming section_auth is an outdated prefix(),
which this causing any issue?

sn/src/routing/core/msg_handling/anti_entropy.rs
```diff
@@ -109,11 +113,21 @@ impl Core {
...
         );
 
+        let signed_section_auth = SectionAuth {
+            value: section_auth.clone(),
+            sig: section_signed,
+        };
+
+        // If it's out section, update our section details.
+        if self.section.prefix().await == section_auth.prefix {
+            debug!("AE-Retry received from our section, chain was updated so updating our section info");
```
maqi, line 123: the comment says `chain was updated so updating our section info`
but the following line of code seems just updating the chain?

sn/src/prefix_map/mod.rs
joshuef: this doesnt work for a dashmap sadly :(

sn/src/prefix_map/mod.rs
```diff
@@ -219,9 +230,9 @@ impl NetworkPrefixMap {
         }
 
         // Check the SAP's key is the last key of the proof chain
-        if proof_chain.last_key() != &signed_section_auth.value.public_key_set.public_key() {
+        if !proof_chain.has_key(&signed_section_auth.value.public_key_set.public_key()) {
```
bochaco, line 233: Do we need to be more laxed? i thought requiring the proof chain to end with the key that belong to the SAP would be easier to reason

sn/src/routing/core/msg_handling/agreement.rs
```diff
@@ -258,42 +263,67 @@ impl Core {
...
         let snapshot = self.state_snapshot().await;
-        let old_chain = self.section.chain().await;
         let node_name = self.node.read().await.name();
+        let old_chain = self.section_chain().await.clone();
 
         for (section_auth, key_sig) in updates {
             let prefix = section_auth.value.prefix;
             info!("New SAP agreed for {:?}: {:?}", prefix, section_auth);
 
+            let our_prefix = self.section().prefix().await;
```
bochaco, line 273: I think this is a problem, we need to get our prefix before the start of the loop otherwise we will be switching to check other prefix once we update our SAP in a previous iteration of the loop

Yoga07: We just use this logging below at L280. So I guess that's okay? We are keeping a copy before we update with the new results, so I think this is okay, but yea, taking a copy before would work better.

sn/src/prefix_map/mod.rs
joshuef: it shouldn't matter, I think?

It's that or updating the error message to be be more accurate. But merging an old chain should not be possible. 

If it comes from an old key it's still trustworthy, right? just slow?

sn/src/routing/core/msg_handling/agreement.rs
bochaco: ah, true, then it's not a problem

sn/src/prefix_map/mod.rs
bochaco: this check is not affected by how old the proof chain or SAP is, it's checking that the proof chain's last key is the key in the SAP, weather they are old or not, it's just about if we accept a proof chain which has more blocks than those needed

Yoga07 (approved):

PR #668, opened 2021-11-01
fix(routing): provide valid proof chain to update prefix map during elders agreement handling

bochaco:


joshuef:
closing as this is now in main

PR #669, opened 2021-11-01
Refactoring section and network knowledge handling

bochaco:


bochaco:
This will be worked on PR #670

PR #670, opened 2021-11-02
Feat unified and centralised network knowledge in routing

bochaco:


bochaco:
These commits are now part of PR #673

PR #675, opened 2021-11-03
Reduce redundant space used by chunks

connec:
- 41b7e53c6 **refactor(sn): remove redundant `K` type parameter from `KvStore`**

  The `Value` trait already included an associated type for the key type,
  and this can be referenced directly in signatures without the need for
  an additional generic type parameter.

- 75cca06aa **fix(sn): make `KvStore::store` perform atomic immutable writes**

  `KvStore` is the database abstraction we use to persist chunks. This is
  itself backed by `sled`. `sled` uses a log as one of the underlying
  storage primitives, which means that all inserts (even overwrites) will
  consume additional space (until the log is compacted, which may happen
  irregularly).
  
  However, chunks themselves are write-once which means that we would only
  ever need to write a value once. As such, we can use `compare_and_swap`
  to write the value only if it's not currently set. Since chunks are
  content-addressable, we furthermore don't need to do anything if the
  value did already exist and wasn't written, since there is only one
  possible value for a given key.

- e613e3cdc **chore(sn): remove unused `KvStore::store_batch`**

  Although this was commented to indicate it will be used again soon, it
  is trivial to recover code from git history, but it may be non-trivial
  to maintain or fix unused code. That's the case for trying to fix used
  space handling in `store_batch`, since `sled::Tree::apply_batch` cannot
  perform compare-and-swap operations. As such, the whole operation
  probably needs a rethink, and since it's not currently used we can defer
  that until it's needed.

- 2823174c5 **chore(sn): remove unnecessary `allow(unused)`**

- 0860efc0b **refactor(sn): add `KvStore::flush` to avoid waiting in tests**

  Some `KvStore` tests verify that used space adjusts appropriately when
  data is written (also when data is removed, but those tests are disabled
  since it's not predictable when sled will compact its log). We measure
  used space by measuring the size of the database files.
  
  sled flushes to disk asynchronously by default, and the tests were
  accounting for this by looping until the size changed. This could lead
  to hung tests in the case of bugs with used space calculation, which
  would be harder to debug than a concrete failure. As such, a test-only
  `flush` method has been added to `KvStore`, which is now used by the
  tests to force a flush to disk.

### DIFF
--- a/sn/src/dbs/kv_store/mod.rs
+++ b/sn/src/dbs/kv_store/mod.rs
@@ -174,6 +174,14 @@ impl<V: Value + Send + Sync> KvStore<V> {
             .collect();
         Ok(keys)
     }
+
+    // We should avoid manually flushing in real code, since it can take a lot of time - tuning
+    // `flush_every_ms` would be preferrable. This is useful in tests, however.
+    #[cfg(test)]
+    pub(crate) async fn flush(&self) -> Result<()> {
+        let _bytes_flushed = self.db.flush_async().await?;
+        Ok(())
+    }
 }
 
 pub(crate) fn convert<T: DeserializeOwned>(key: sled::IVec) -> Result<T> {
--- a/sn/src/dbs/kv_store/mod.rs
+++ b/sn/src/dbs/kv_store/mod.rs
@@ -130,68 +130,6 @@ impl<V: Value + Send + Sync> KvStore<V> {
         Ok(())
     }
 
-    /// Stores a batch of values.
-    ///
-    /// If there is not enough storage space available, returns `Error::NotEnoughSpace`.  In case of
-    /// an IO error, it returns `Error::Io`.
-    ///
-    /// If a value with the same id already exists, it will be overwritten.
-    #[allow(unused)] // this will soon be used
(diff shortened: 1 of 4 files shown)


Yoga07 (dismissed):

sn/src/dbs/kv_store/mod.rs
```diff
@@ -97,96 +95,31 @@ impl<K: Key, V: Value + Send + Sync> KvStore<K, V> {
...
+                .await
+        {
             return Err(Error::NotEnoughSpace);
         }
 
-        let res = self.db.apply_batch(batch);
+        // Atomically write the value if it's new - this prevents multiple concurrent writes from
+        // consuming extra space (since sled is backed by a log). We throw away the value otherwise,
+        // since current use-cases do not require updates.
+        let _res = self
```
Yoga07, line 118: We could maybe log the result as before in the `match` below? Just to be wary that we aren't ignoring an error.

connec: Yeah, I swithered on what logging to include. I think it's a good point to log it/match on it though, to make sure we don't lose the `?` in a refactor.

Yoga07 (approved):

PR #678, opened 2021-11-04
Fix NRS failures

grumbach:
- fixed NRS register Failures (by going down to lower level register that don't use the resolver+NRS)
- along with fixing the last NRS failing tests
- NRS tests are all green ✅ now 🎉


joshuef:
Let's get those working tests running in a CI workflow (they need a live network, right? so probs the `runa all tests` flow

joshuef (changes requested):
lets get the green tests running on CI 💪

joshuef (approved):

PR #691, opened 2021-11-08
chore(sn_api): revert switch from `time` to `chrono`

jacderida:
Revert "chore(sn_api): switch from `chrono` to `time`".

This reverts commit 442c1bdd3f9c22424aa9642d71d4016c868b0b58.

This change breaks publishing for the `sn_api` crate, which seems to insist on bringing in some other version of the crate.

The change was made because RUSTSEC-2020-0159 (security advisory item) was being flagged by`cargo-deny`. It seems to be quite acceptable to ignore this issue, so it can be added to an ignore list. This advisory item was added to the ignore list in the QA repo.

I tested the dry run publish for `sn_api` and it now fails on a different error, related to a type from `safe_network` being missing. This would be expected as a new version of `safe_network` would need to be published.


joshuef:
closed as should not be an issue after #694

PR #695, opened 2021-11-09
fix: refactor the Join process regarding used_recipients

maqi:

### DIFF
--- a/sn/src/routing/core/bootstrap/join.rs
+++ b/sn/src/routing/core/bootstrap/join.rs
@@ -6,10 +6,10 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::read_prefix_map_from_disk;
+use super::{read_prefix_map_from_disk, UsedRecipientSaps};
 use crate::messaging::{
     system::{JoinRejectionReason, JoinRequest, JoinResponse, ResourceProofResponse, SystemMsg},
-    DstLocation, MessageType, MsgKind, NodeAuth, SectionAuthorityProvider, WireMsg,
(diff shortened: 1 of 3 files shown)


sn/src/routing/core/bootstrap/join.rs
```diff
@@ -204,8 +195,42 @@ impl<'a> Join<'a> {
                         section_auth
                     );
 
+                    if prefix.bit_count() < self.prefix.bit_count()
+                        || section_auth.elders.len() < expected_elders
```
bochaco, line 199: I think we should consider using the PrefixMap instead of this, the PrefixMap can check the received SAP is not only valid but a new one. Although for this we need the proof chain with the SAP, which I think we are actually missing in our Retry message so we can trust the new SAP based on the genesis key and/or previous SAPs we are aware of.

maqi: Well, PrefixMap is currently only available to `joined` node. A bootstrapping node doesn't have such struct. Sure we can add it in, but I'd prefer it to be tackled in separate PR (though, I'd argue whether it's really necessary).
Yeah, having Retry message containing a `chain` will be the best. The same that I'd suggest to have it within the same PR of using PrefixMap (if necessary).

bochaco: It is necessary because we cannot otherwise trust the new SAPs we receiev in JoinRetry/Redirect. The PrefixMap already does all the validations to be able to trust new SAPs based on the proof chain, so it is necessary unless we duplicate the SAP and proof chain validation code we have in the prefix map.

bochaco (dismissed):

bochaco (approved):

PR #697, opened 2021-11-10
Join-flow connection management

connec:
- 83d94eba9 **feat(sn): support `--json-logs` in `testnet`**

  This makes it easier to enable JSON logs for local testnets.

- 53f47101b **chore(sn): add more tracing around connections**

  This is mostly peppering `.in_current_span` to some spawned futures.
  Additionally the traces for `ConnectionOpened` and `ConnectionClosed`
  now use fields for their context, since it makes JSON processing easier.

- be2478154 **feat: add a resources dir for `jq` filters**

  Using `jq` to process JSON logs could be a great tool for debugging
  network activity. Particularly if we expand the use of fields with log
  markers. This new directory is somewhere we can put useful filters for
  reuse.
  
  The first filter is one to filter for `ConnectionOpened` and
  `ConnectionClosed` log markers, and output only the timestamp,
  connection ID, and simplified spantrace for each.

- dc4aa2fcc **refactor(sn)!: remove `Copy` from `Peer`**

  The plan is to put a `Connection` into `Peer`, which is not `Copy`. As
  such, `Peer` itself will be unable to derive `Copy`. Rather than pollute
  that change, it's going in a separate commit.
  
  BREAKING CHANGE: `routing::Peer` is no longer `Copy`.

- c0b86a362 **refactor(sn): pass `Connection` to `join_network`**

  When bootstrapping to a network, we immediately connect to a peer in
  `Comm::bootstrap`. Before this commit we would essentially throw this
  away. We now return it, and pass it on into `join_network` so that the
  initial join request reuses the connection.
  
  To make the APIs work nicely, and support more connection reuse in
  future, `Peer` now has a `connection: Option<Connection>` field, and
  this is used by `Comm::send` to send over the existing connection if one
  is set.
  
  For a two node network, this only really saves a single connection,
  since the first join request will get a retry response, and we lose the
  connection handle when this happens. This may be improved in future, if
  the gains seem worthwhile. Another caveat is that `Peer::connection` may
  be `Some`, but the contained connection may be lose. This is not
  currently handled by `Comm::send` (the send will simply fail).

- 50db6dc28 **refactor(sn): maintain connections through join redirect/retry**

  This is not a beautiful implementation, but essentially when we get the
  SAP from a `JoinRetry` or `JoinRedirect` we will update the peers with
  any connections we currently hold to existing peers.

- 125df6880 **refactor(sn): support reusing connections from incoming messages**

  This replaces the `SocketAddr` in `ConnectionEvent` with a new
  `UnknownPeer` struct, which always has an address and may have a
  `Connection` (currently, the `Connection` would only be `None` in tests,
  but that might not always be the case). An `UnknownPeer` can be
  'upgraded' to `Peer` by `identify`ing it with a name.
  
  This is currently only used in the join flow.

- 77031726e **refactor(sn): rename `routing::peer` to `routing::peers`**

  This is a more general name for the module, making it appropriate to put
  things besides the `Peer` structs (e.g. `ElderCandidates`...).

- 491d75a28 **refactor(sn)!: move `ElderCandidates` from `messaging` to `routing`**

  This is going the same way as `Peer`, and for similar reasons. Only one
  message type was using `ElderCandidates` (`DkgStart`) and the two fields
  can just as easily be inlined. The `ElderCandidatesUtils` trait has been
  dropped in favour of an inherent impl.
  
  The move to routing should allow us to add some connection-related
  properties, to facilitate better connection reuse.
  
  BREAKING CHANGE: `messaging::system::SystemMsg::DkgStart` no longer has
  an `elder_candidates` field, but rather separate `prefix` and `elders`
  fields. `messaging::system::ElderCandidates` has moved to
  `routing::ElderCandidates`.
  `routing::network_knowledge::section_authority_provider::ElderCandidateUtils`
  has been removed (the methods are replaced by an inherent impl, but they
  are now `pub(crate)`).

- a2f8f860e **refactor(sn): use `UnknownPeer` in `HandleMessage` command**

  By passing this on, we're able to reuse the incoming connection handle
  when replying. In particular, this will get the connection handle into
  `handle_valid_msg`, though we don't yet use it in many call paths
  (notably not DKG, which accounts for most messages during section
  formation).


connec:
I've drafted this for now in favour of #699 which is a much easier win, albeit less idealistic.

PR #708, opened 2021-11-11
Fix section startup

joshuef:


sn/src/routing/network_knowledge/mod.rs
```diff
@@ -191,6 +197,44 @@ impl NetworkKnowledge {
...
+                    provided_sap,
+                    section_key
+                );
+            }
+        } else {
+            trace!("Doesn't have cached SAP or chain for {:?}", section_key);
+        }
+        if switched {
+            *self.cache_chain.write().await = None;
+            *self.cache_sap.write().await = None;
```
bochaco, line 233: Perhaps we can move these two assignments above L217 where we set this flag to `true`

bochaco (approved):

PR #710, opened 2021-11-11
Further separation of messaging types from routing state

connec:
- ba7b2c4a4 **refactor(sn)!: move `SectionPeers` from `messaging` to `routing`**

  `SectionPeers` was only used in the `SystemMsg::AntiEntropyUpdate`
  message. Rather than simply inlining the `DashMap`, it's been replaced
  by `BTreeSet<SectionAuth<NodeState>>` since the map keys are redundant
  with the `name` field of `NodeState` (and `BTreeSet` specifically for
  deterministic ordering).
  
  With the move, it's also become `pub(crate)`, so from the perspective of
  the public API this type has been removed.
  
  BREAKING CHANGE: `messaging::system::SectionPeers` has been removed. The
  `messaging::system::SystemMsg::AntiEntropyUpdate::members` field's type
  has changed to `Option<BTreeSet<SectionAuth<NodeState>>>`.

- a4e857f31 **refactor(sn): duplicate `NodeState` in `routing`**

  This is part of the move towards passing connection handles through
  network knowledge. Whilst it would be possible to put a connection
  handle in the `messaging` struct, and just skip it when (de)serialising,
  this seems less clean than having a separate struct.

- 26bf06e40 **refactor(sn)!: duplicate `SectionAuthorityProvider` in `routing`**

  This is part of the move towards passing connection handles through
  network knowledge. Whilst it would be possible to put a connection
  handle in the `messaging` struct, and just skip it when (de)serialising,
  this seems less clean than having a separate struct.
  
  BREAKING CHANGE: `routing::Routing::our_section_auth` is now private.
  `routing::Routing::matching_section` now returns
  `routing::SectionAuthorityProvider`, though this struct itself is
  private and so cannot be named. `routing::Routing::matching_section`
  will hopefully become private in future, so you should avoid using it.

- 44a5157d9 **refactor(sn): make `NodeState` not `Copy`**

  At some point we might stick a connection handle in there, which would
  not be `Copy`, so unimplementing it now may save some changed lines in
  future.


joshuef (changes requested):
see `into` vs `from` comment. If we can do that easily it helps readability a heap IMO

joshuef (approved):

PR #733, opened 2021-11-17
chore: fail query if we cannot send to more than 1 elder

joshuef:

### DIFF
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -8,7 +8,7 @@
 
 use super::Session;
 use crate::client::connections::messaging::NUM_OF_ELDERS_SUBSET_FOR_QUERIES;
-use crate::client::{connections::messaging::send_message, Error};
+use crate::client::{connections::messaging::send_message, Error, Result};
 use crate::messaging::data::DataCmd;
 use crate::messaging::{
     data::{CmdError, ServiceMsg},
@@ -52,14 +52,12 @@ impl Session {
                             error!("Error while processing incoming message: {:?}. Listening for next message...", err);
                         }
                     },
-                    Err(Error::Generic(_)) => {
-                        // TODO: FIX error type
+                    Err(_) => {
+                        // TODO: Can we recover here?
                         info!("IncomingMessages listener has closed.");
                         break;
                     }
-                    Err(err) => {
-                        error!("Error while getting incoming message: {:?}. Listening for next message...", err);
-                    }
+
                 }
             }
 
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -49,13 +49,26 @@ impl Session {
                     Ok(msg) => match Self::handle_msg(msg, src, session.clone()).await {
                         Ok(()) => {},
                         Err(err) => {
-                            error!("Error while processing incoming message: {:?}. Listening for next message...", err);
+                            error!("Error while handling incoming message: {:?}. Listening for next message...", err);
                         }
                     },
-                    Err(_) => {
-                        // TODO: Can we recover here?
-                        info!("IncomingMessages listener has closed.");
-                        break;
+                    Err(error) => {
+                        match error {
+                            Error::Generic(_) => {
+                                // TODO: Can we recover here?
+                                info!("IncomingMessages listener has closed.");
+                                break;
+                            }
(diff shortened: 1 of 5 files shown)


sn/src/client/connections/listeners.rs
```diff
@@ -52,14 +52,12 @@ impl Session {
                             error!("Error while processing incoming message: {:?}. Listening for next message...", err);
                         }
                     },
-                    Err(Error::Generic(_)) => {
-                        // TODO: FIX error type
+                    Err(_) => {
```
connec, line 55: I don't think this changes anything, in the case of `ConnectionLoss`, and makes it more frail otherwise. Currently, when `ConnectionIncoming` yields a `ConnectionLoss` error:

1. It will be returned from `listen_for_incoming_message` (propagated by `?`).
2. It'll hit the `Err(err)` case, and go around the loop again.
3. The next time `incoming_messages.next()` is called, it will return `None`, since `IncomingMessages` will return `None` after connection loss (if not, that's a bug in `qp2p`).
4. On `None`, `listen_for_incoming_message` yields `Error::Generic` and the loop breaks.

The `Error::Generic` stuff does seem a bit convoluted, but we should still filter it to `ConnectionLoss` only imo, or we will throw away connections even if we receive, e.g., a message we can't deserialize.

sn/src/client/connections/messaging.rs
```diff
@@ -298,6 +299,16 @@ impl Session {
             }
         }
 
+        let send_failures = *failed_sends.lock().await;
+        if send_failures >= 2 {
```
connec, line 303: I think this could be replaced with `if discarded_responses >= 2 { ... }`. More generally, I think `failed_sends` is redundant, since counting the errors after `join_all` would get the same result with fewer shenanigans?

Also, I thought we could tolerate 2 misbehaving elders, so should this be `> 2`?

connec: Ah, I always forget `task::JoinHandle` yields panics in the outer result. I guess replacing `failed_sends` would need:

```rust
for result in results {
    match result {
        Err(_) => discarded_responses += 1,
        Ok(Err(_)) => failed_sends += 1,
        _ => {}
    }
}
```

e.g., so a bit extraneous for this PR.

sn/src/client/connections/listeners.rs
joshuef: Aye it wont propogate. But we should shut down the loop if it's useless I think. I'll fix that error detection there so e dont trip up on other things though as you say 👍

sn/src/client/connections/messaging.rs
joshuef: > if discarded_responses >= 2 { ... }

Hmm, we normally send to 3 elders for queries. We _could_ just send one... but it seemed like perhaps its worth a do over if 2/3 are failing? Dunnnoo

connec: Yeah, that makes sense. I guess this falls under the `TODO` comment above about the approach to validation.

sn/src/client/connections/listeners.rs
```diff
@@ -49,17 +49,28 @@ impl Session {
...
-                    Err(Error::Generic(_)) => {
-                        // TODO: FIX error type
-                        info!("IncomingMessages listener has closed.");
-                        break;
-                    }
-                    Err(err) => {
-                        error!("Error while getting incoming message: {:?}. Listening for next message...", err);
+                    Err(error) => {
+                        match error {
+                            Error::Generic(_) => {
```
connec, line 57: Should this not be `Error::IncomingMessages` now, after the change to `listen_for_incoming_message`?

sn/src/client/connections/listeners.rs
```diff
@@ -49,17 +49,28 @@ impl Session {
...
+                                // TODO: Can we recover here?
+                                info!("IncomingMessages listener has closed.");
+                                break;
+                            }
+                            Error::QuicP2p(qp2p_err) => {
```
connec, line 62: We could filter for connection loss only with:

```rust
Error::QuicP2p(qp2p::RpcError::Recv(qp2p::RecvError::ConnectionLost(_))) => ...
```

That said, I don't think it would be particularly harmful to close the listener for all `qp2p` errors, just marginally less efficient in some situations (but those would be situations where we or the sender is doing something odd with the streams, like closing them too soon or whatever).

connec (approved):

PR #734, opened 2021-11-17
Chore further node code cleanup

joshuef:


sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -495,6 +495,19 @@ impl Core {
             // plugging in msg handlers.
             SystemMsg::NodeCmd(node_cmd) => {
                 match node_cmd {
+                    NodeCmd::RecordStorageLevel { node_id, level, .. } => {
```
bochaco, line 498: Shall we merge the matching arms with the parent ones? i.e.:
```
SystemMsg::NodeCmd(NodeCmd::RecordStorageLevel { node_id, level, .. }) => { ... },
SystemMsg::NodeCmd(NodeCmd::ReceiveExistingData { metadata }) => { ... },
...
```
Also, I guess the above TODO shall be removed?

bochaco: ....ah i see we still send events, so perhaps keep the TODO

bochaco (approved):

PR #756, opened 2021-11-23
refactor(dkg): replace dkg message backlog with message exchange

lionel1704:
when a node has not received enough sig. shares to start processing a
SystemMsg::DkgStart message a node can request the sender for the
DkgSessionInfo which includes the aggregated section signature so the
dkg session can be initialized

### DIFF
--- a/sn/src/messaging/system/mod.rs
+++ b/sn/src/messaging/system/mod.rs
@@ -33,6 +33,9 @@ use std::{
 };
 use xor_name::{Prefix, XorName};
 
+use super::authority::SectionAuth as SectionAuthProof;
+use super::AuthorityProof;
+
 #[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
 #[allow(clippy::large_enum_variant)]
 /// Message sent over the among nodes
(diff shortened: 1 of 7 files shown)


sn/src/messaging/system/mod.rs
```diff
@@ -108,6 +111,28 @@ pub enum SystemMsg {
...
+        message: DkgMessage,
+    },
+    /// DKG session info along with section authority
+    DkgSessionInfo {
+        /// The identifier of the DKG session to start.
+        session_id: DkgSessionId,
+        /// The section prefix. It matches all the members' names.
+        prefix: Prefix,
+        /// The section's complete set of elders as a map from their name to their socket address.
+        elders: BTreeMap<XorName, SocketAddr>,
```
bochaco, line 128: Are these necessarily Elders or just participants of the DKG round/session?

bochaco: ...I jut took a look at `DkgStart` message and we also have a field name `elders`, it sounds like we should rename them in both cases to something like `participants` ?

joshuef:
Closing as this is in DkgImprovement branch now

PR #757, opened 2021-11-23
refactor(dkg): replace dkg message backlog with message exchange

joshuef:
when a node has not received enough sig. shares to start processing a
SystemMsg::DkgStart message a node can request the sender for the
DkgSessionInfo which includes the aggregated section signature so the
dkg session can be initialized

### DIFF
--- a/sn/src/messaging/system/mod.rs
+++ b/sn/src/messaging/system/mod.rs
@@ -33,6 +33,9 @@ use std::{
 };
 use xor_name::{Prefix, XorName};
 
+use super::authority::SectionAuth as SectionAuthProof;
+use super::AuthorityProof;
+
 #[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
 #[allow(clippy::large_enum_variant)]
 /// Message sent over the among nodes
@@ -107,6 +110,24 @@ pub enum SystemMsg {
         prefix: Prefix,
         /// The section's complete set of elders as a map from their name to their socket address.
         elders: BTreeMap<XorName, SocketAddr>,
+    },
+    /// Message sent when a DKG session has not started
+    DkgSessionUnknown {
+        /// The identifier of the DKG session this message is for.
+        session_id: DkgSessionId,
+    },
+    /// DKG session info along with section authority
+    DkgSessionInfo {
+        /// The identifier of the DKG session to start.
+        session_id: DkgSessionId,
+        /// The section prefix. It matches all the members' names.
+        prefix: Prefix,
+        /// The section's complete set of elders as a map from their name to their socket address.
+        elders: BTreeMap<XorName, SocketAddr>,
+        /// Section authority for the DKG start message
+        section_auth: AuthorityProof<SectionAuthProof>,
+        /// Messages processed in the session so far
+        message_cache: Vec<DkgMessage>,
     },
     /// Message exchanged for DKG process.
     DkgMessage {
--- a/sn/src/messaging/system/mod.rs
+++ b/sn/src/messaging/system/mod.rs
@@ -115,6 +115,8 @@ pub enum SystemMsg {
     DkgSessionUnknown {
         /// The identifier of the DKG session this message is for.
         session_id: DkgSessionId,
+        /// DKG message that came in
+        message: DkgMessage,
     },
     /// DKG session info along with section authority
     DkgSessionInfo {
@@ -128,6 +130,8 @@ pub enum SystemMsg {
         section_auth: AuthorityProof<SectionAuthProof>,
         /// Messages processed in the session so far
         message_cache: Vec<DkgMessage>,
+        /// The original DKG message
+        message: DkgMessage,
     },
     /// Message exchanged for DKG process.
     DkgMessage {
--- a/sn/src/routing/core/msg_handling/mod.rs
+++ b/sn/src/routing/core/msg_handling/mod.rs
@@ -633,7 +633,10 @@ impl Core {
                 )
                 .await
             }
-            SystemMsg::DkgSessionUnknown { session_id } => {
+            SystemMsg::DkgSessionUnknown {
+                session_id,
+                message,
+            } => {
                 if let Some(session_info) = self.dkg_sessions.read().await.get(&session_id).cloned()
                 {
                     let DkgSessionInfo {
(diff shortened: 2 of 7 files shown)


sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -621,6 +633,74 @@ impl Core {
...
+                    elders: elders.clone(),
+                })?;
+                let auth = section_auth.into_inner();
+                if self
+                    .network_knowledge
+                    .has_chain_key(&auth.sig.public_key)
+                    .await
+                {
+                    if let Err(err) = AuthorityProof::verify(auth, payload) {
+                        error!("Error verifying signature for DkgSessionInfo: {:?}", err)
```
maqi, line 690: in case of error verification, shall return instead of continue with the handle_dkg_start?

sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -621,6 +633,74 @@ impl Core {
...
+                // Reconstruct the original DKG start message and verify the section signature
+                let payload = WireMsg::serialize_msg_payload(&SystemMsg::DkgStart {
+                    session_id,
+                    prefix,
+                    elders: elders.clone(),
+                })?;
+                let auth = section_auth.into_inner();
+                if self
+                    .network_knowledge
+                    .has_chain_key(&auth.sig.public_key)
```
maqi, line 686: may worth to check the auth.sig.public_key is the last of the chain to avoid start up an old one?

sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -621,6 +633,74 @@ impl Core {
...
+                    .await
+                {
+                    if let Err(err) = AuthorityProof::verify(auth, payload) {
+                        error!("Error verifying signature for DkgSessionInfo: {:?}", err)
+                    } else {
+                        trace!("DkgSessionInfo signature verified");
+                    }
+                } else {
+                    warn!("Cannot verify DkgSessionInfo. Unknown key!");
+                }
```
maqi, line 696: the `section_auth` needs to be inserted into `self.dkg_sessions`

sn/src/messaging/system/mod.rs
```diff
@@ -108,6 +111,24 @@ pub enum SystemMsg {
         /// The section's complete set of elders as a map from their name to their socket address.
         elders: BTreeMap<XorName, SocketAddr>,
     },
+    /// Message sent when a DKG session has not started
+    DkgSessionUnknown {
+        /// The identifier of the DKG session this message is for.
+        session_id: DkgSessionId,
```
maqi, line 117: better to contain the DkgMessage that caused this

sn/src/messaging/system/mod.rs
```diff
@@ -108,6 +111,24 @@ pub enum SystemMsg {
...
+        /// The identifier of the DKG session to start.
+        session_id: DkgSessionId,
+        /// The section prefix. It matches all the members' names.
+        prefix: Prefix,
+        /// The section's complete set of elders as a map from their name to their socket address.
+        elders: BTreeMap<XorName, SocketAddr>,
+        /// Section authority for the DKG start message
+        section_auth: AuthorityProof<SectionAuthProof>,
+        /// Messages processed in the session so far
+        message_cache: Vec<DkgMessage>,
```
maqi, line 130: better have a seperate filed to contain the original DkgMessage that triggered the DkgSessionUnknown to be sent out (which shall be contained inside it)

sn/src/routing/dkg/voter.rs
```diff
@@ -183,20 +162,22 @@ impl DkgVoter {
...
-            trace!("Pushing to backlog {:?} - {:?}", session_id, message);
-            self.backlog.write().await.push(*session_id, message);
+            trace!(
+                "Sending DkgSessionUnknown {{ {:?} }} to {}",
+                &session_id,
+                &sender
+            );
+            commands.push(Command::PrepareNodeMsgToSend {
+                msg: SystemMsg::DkgSessionUnknown {
+                    session_id: *session_id,
```
maqi, line 174: better put the DkgMessage inside it

sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -621,6 +633,74 @@ impl Core {
...
+            }
+            SystemMsg::DkgSessionInfo {
+                session_id,
+                prefix,
+                elders,
+                message_cache,
+                section_auth,
+            } => {
+                let msg = message_cache.last().unwrap().clone();
```
maqi, line 675: This actually shall be the original DkgMessage triggered the DkgSessionUnknown .

sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -621,6 +633,74 @@ impl Core {
...
+                    .has_chain_key(&auth.sig.public_key)
+                    .await
+                {
+                    if let Err(err) = AuthorityProof::verify(auth, payload) {
+                        error!("Error verifying signature for DkgSessionInfo: {:?}", err)
+                    } else {
+                        trace!("DkgSessionInfo signature verified");
+                    }
+                } else {
+                    warn!("Cannot verify DkgSessionInfo. Unknown key!");
```
maqi, line 695: then shall return instead of continue with the handling?

sn/src/routing/dkg/voter.rs
```diff
@@ -225,16 +206,24 @@ impl DkgVoter {
...
     ) -> Result<Vec<Command>> {
         if let Some(mut session) = self.sessions.get_mut(&session_id) {
             session.handle_dkg_history(node, session_id, message_history, section_pk)
         } else {
-            for message in message_history {
-                trace!("Pushing to backlog {:?} - {:?}", session_id, message);
-                self.backlog.write().await.push(session_id, message);
-            }
-            Ok(vec![])
+            trace!(
```
maqi, line 215: better raise a warning, as handling of message_history shall always coupled with a session already got created.
if not, this will only happen when trying to handle an outdated response?

maqi (commented):
also worth to check that at https://github.com/maidsafe/safe_network/blob/main/sn/src/routing/core/messaging.rs#L453
the name shall be from the prefix

PR #774, opened 2021-11-24
Remove connection pooling

connec:
- 9ccb02c0a **refactor: add input filename to connection tracing filter**

  This makes it easier to see where each connection is opened.

- 7271932a6 **refactor(sn): reuse more `Peer`s in DKG sessions**

  We previously started to check our knowledge of section members for
  existing `Peer`s, however in some cases it seems that elders are not
  present in `members` (perhaps deliberately?). To ensure maximum reuse,
  we now also check the SAP for reusable connections.

- 95303cc57 **refactor(sn): remove connection pooling**

  This completely removes `ConnectedPeers`. Prior changes ensure that
  connection usage remains fairly nominal (though still higher than
  optimal), and that clients are still reachable without reconnection.

- 0877cc1b9 **fix(sn): buffer possible DKG messages received when joining**

  When an elder decides to approve a node's join request, it also decides
  whether the new node should become an elder. If so, it sends the join
  approval and the DKG start message concurrently. This introduces a race
  for which message the joining node will see first. If it's the DKG start
  message, it would previously just be dropped. Eventually either another
  node would join and trigger a successful round of DKG, or the node would
  be voted offline again due to being unresponsive from the perspective of
  the elder.
  
  This commit adds an (unbounded) buffer to the `Join` flow that collects
  any `NodeBlsShareAuth` messages that arrive before the join flow
  terminates. This is passed up to the caller, who replays these messages
  to the `ConnectionEvent` sender so that they are picked up by the full
  routing message handling, once it's established.
  
  This is not an ideal or complete solution (e.g. it won't help if the DKG
  start message is lost). Ideally the elder would not assumed "exactly
  once" delivery of DKG start messages and would retry if it suspects
  message loss.

- 194f41aa5 **refactor(sn): merge sender connection into SAP when joining**

  This sets up joining nodes to reuse the connection they already have
  with the approving elder.

- ba288153e **refactor(sn): reuse system message sender connections**

  Whenever we receive a valid system message, we now merge the sender's
  connection into our network knowledge, if applicable.
  
  An example of when this is useful is when a node uses their SAP to cold-
  call an elder who they have not yet spoken to. Without this change,
  there would be additional connections since the elder would reply over a
  new connection. With this change in place, they will instead be able to
  reply over the same sending connection.


connec:
Superseded by #782.

PR #780, opened 2021-11-25
Replace DKG message backlog with AE message exchange

lionel1704:
- refactor(dkg): replace dkg message backlog with message exchange
- fix(dkg)!: include the original message when requesting DkgSessionInfo


joshuef:
merged via the dkg improvement branch

PR #781, opened 2021-11-25
fixes

maqi:

### DIFF
--- a/sn/src/routing/core/bootstrap/join.rs
+++ b/sn/src/routing/core/bootstrap/join.rs
@@ -31,7 +31,7 @@ use backoff::{backoff::Backoff, ExponentialBackoff};
 use bls::PublicKey as BlsPublicKey;
 use futures::future;
 use resource_proof::ResourceProof;
-use tokio::time::{sleep, Duration as TokioDuration};
+use tokio::time::sleep;
 use tokio::{sync::mpsc, time::Duration};
 use tracing::Instrument;
 use xor_name::Prefix;
@@ -165,9 +165,8 @@ impl<'a> Join<'a> {
                     return Err(Error::TryJoinLater);
                 }
                 JoinResponse::Rejected(JoinRejectionReason::DKGUnderway) => {
-                    error!("The Section we are trying to join is going through DKG, retrying in a minute");
-                    sleep(TokioDuration::from_secs(20)).await;
-                    self.send_join_requests(join_request.clone(), &recipients, section_key, false)
+                    error!("The Section we are trying to join is going through DKG, backing off and retrying again");
+                    self.send_join_requests(join_request.clone(), &recipients, section_key, true)
                         .await?;
                     continue;
                 }
@@ -192,7 +191,7 @@ impl<'a> Join<'a> {
                     }
 
                     trace!(
-                        ">>>>> 100 >>>>> This node has been approved to join the network at {:?}!",
+                        "This node has been approved to join the network at {:?}!",
                         section_auth.prefix,
                     );
 
@@ -238,6 +237,8 @@ impl<'a> Join<'a> {
                                 resource_proof_response: None,
                                 aggregated: Some(auth),
                             };
+
+                            // Resend the JoinRequest now that we have collected enough ApprovalShares from the Elders
                             self.send_join_requests(join_req, &[sender], section_key, true)
                                 .await?;
                             continue;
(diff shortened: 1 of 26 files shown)


sn/src/routing/core/msg_handling/join.rs
```diff
@@ -95,17 +95,22 @@ impl Core {
...
-                self.send_direct_message(peer, node_msg, our_section_key)
-                    .await?,
-            ]);
+                trace!("Sending {:?} to {}", node_msg, peer);
+                return Ok(vec![
+                    self.send_direct_message(peer, node_msg, our_section_key)
+                        .await?,
+                ]);
+            } else {
+                trace!("is_dkg_underway flag is true, however dkg_voter doesn't have ongoing dkg");
```
bochaco, line 111: Can we not have the flag to avoid needing to sync them up? just query the dkg_voter from wherever we need to know if there is a DKG underway?

sn/src/routing/core/bootstrap/join.rs
```diff
@@ -237,9 +237,18 @@ impl<'a> Join<'a> {
...
+                            {
+                                signed_sap.value.elders().cloned().collect()
+                            } else {
+                                warn!("cannot find recipients to send aggregated JoinApproval");
+                                continue;
+                            };
+                            trace!("Sending aggregated JoinRequest to {:?}", recipients);
                             // Resend the JoinRequest now that we have collected enough ApprovalShares from the Elders
-                            self.send_join_requests(join_req, &[sender], section_key, true)
+                            self.send_join_requests(join_req, &recipients, section_key, false)
```
Yoga07, line 251: `backoff` argument needs to be set to `true` here as this is also one of the steps in the Join process.

sn/src/routing/core/bootstrap/join.rs
```diff
@@ -303,10 +312,19 @@ impl<'a> Join<'a> {
...
-                            &Prefix::default().range_inclusive(),
-                            expected_age,
-                        );
+                        // The expected_age is a sequence of 98, 96, 94, 92, ...
+                        // The expected_prefix is deduced from it.
+                        let mut cur_age = expected_age / 2;
+                        let mut expected_prefix = Prefix::default();
+                        while cur_age > 0 {
+                            let remain_age = cur_age / 2;
+                            expected_prefix = expected_prefix.pushed(cur_age != (remain_age * 2));
```
bochaco, line 321: If I understand this correctly, I think something like the following instead is better, it's difficult to understand otherwise:
```
let push_prefix_0 = cur_age % 2 == 0;
expected_prefix = expected_prefix.pushed(push_prefix_0);
```

sn/src/routing/core/msg_handling/join.rs
joshuef: I'm hoping we can get rid of this check entirely w/ the priority PR

sn/src/routing/core/bootstrap/join.rs
```diff
@@ -237,9 +237,18 @@ impl<'a> Join<'a> {
                                 resource_proof_response: None,
                                 aggregated: Some(auth),
                             };
-
+                            let name = self.node.name();
+                            let recipients: Vec<Peer> = if let Some(signed_sap) =
+                                self.prefix_map.closest_or_opposite(&name, None)
```
Yoga07, line 242: Not sure if we can use `closest_or_opposite` here. Theoretically, we won't have a problem as we've already contacted a section using the prefix, but just in case, it'd be better to assert that the fetched SAP PK verifies the aggregated message and then proceed.

sn/src/routing/core/bootstrap/join.rs
maqi: This for the Request that we have aggregated sucessfully, hence I don't think a backoff is really necessary?

sn/src/routing/core/bootstrap/join.rs
bochaco: Also, the expected age is supposedly always an even number, so this would be always pushing the same prefix....?

sn/src/routing/core/bootstrap/join.rs
Yoga07: Backoff would just be a more solid guarantee but yea, not backing off would also work as real-world scenarios it might take a relatively larger time for all 7 elders to communicate back and then aggregate the share. We _should_ be okay then 👍

sn/src/routing/core/msg_handling/join.rs
```diff
@@ -108,20 +131,48 @@ impl Core {
...
+            let elders = self.network_knowledge.elders().await;
+            section_members = self.network_knowledge.active_members().await.len();
+            // Forces the joining node to be younger than the youngest elder in genesis section
+            // avoiding unnecessary churn.
+
+            // Check if `elder_count()` Elders are already present
+            if elders.len() == elder_count() {
+                // Check if the joining node is younger than the youngest elder and older than
+                // MIN_AGE in the first section to avoid unnecessary churn during genesis.
+                let is_age_valid =
```
bochaco, line 150: I'm trying to understand if we changed the whole logic of how to check the valid age, but we are returning a boolean which tells if the age is invalid (check L160), and we also use it as that afterwards, so renaming it to `is_valid` is confusing as I'm not sure what we want from this flag now?

PR #789, opened 2021-11-26
Prioritise messages

joshuef:


sn/src/client/connections/messaging.rs
```diff
@@ -235,7 +235,8 @@ impl Session {
         };
         let msg_kind = MsgKind::ServiceMsg(auth);
         let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, dst_location)?;
-        let priority = wire_msg.msg_kind().priority();
+        // TODO: prevent this clone
+        let priority = wire_msg.clone().into_message()?.priority();
```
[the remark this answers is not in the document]
joshuef: It's at the client, so I'm not sure we need to worry about that (nor what the way around that would be for now?)

sn/src/routing/error.rs
```diff
@@ -22,6 +22,8 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
 pub enum Error {
+    #[error("Permit semaphore was closed. We cannot continue")]
+    SemaphoreClosed,
```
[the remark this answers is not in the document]
joshuef: I think we need it as if a semaphore has closed, the ndoe will cease to function. (it should basically never be hit, but if it is, it needs to bubble all the way up to command error handling.

sn/src/client/connections/messaging.rs
```diff
@@ -467,7 +467,8 @@ pub(super) async fn send_message(
     endpoint: Endpoint,
     msg_id: MessageId,
 ) -> Result<(), Error> {
-    let priority = wire_msg.msg_kind().priority();
+    let priority = wire_msg.clone().into_message()?.priority();
```
maqi, line 470: wondering if we can have a `priority` field within the `MsgKind` to avoid this extra `into_message` (which incur a deserialisation?) ?

sn/src/routing/routing_api/dispatcher.rs
```diff
@@ -44,26 +80,308 @@ impl Dispatcher {
...
+
+        let mut loop_count = 0;
+        while semaphore.available_permits() != count {
+            loop_count += 1;
+
+            if loop_count > 500 {
+                return Err(Error::CouldNotGetPermitInTime);
+            }
+
+            time::sleep(Duration::from_millis(500)).await;
```
maqi, line 117: 0.5s could be too long

maqi: especially the totally timeout for this loop wait is 250s

sn/src/routing/routing_api/dispatcher.rs
```diff
@@ -44,26 +80,308 @@ impl Dispatcher {
...
+                        .await
+                        .map_err(|_| Error::SemaphoreClosed),
+                )
+            }
+            AE_MSG_PRIORITY => {
+                trace!(
+                    "{:?} Awaiting DKG Completion before continuing with Join Msg",
+                    cmd_id
+                );
+                self.wait_for_priority_commands_to_finish(
```
maqi, line 198: better merge all these `wait_for_priority_commands_to_finish` calls into one function that:
calls correspodent waits based on the priority input

sn/src/routing/routing_api/dispatcher.rs
```diff
@@ -44,26 +80,308 @@ impl Dispatcher {
             core,
             cancel_timer_tx,
             cancel_timer_rx,
+            join_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
+            ae_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
+            infra_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
+            dkg_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
+            service_msg_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
+            node_data_permits: Arc::new(Semaphore::new(SEMAPHORE_COUNT)),
```
maqi, line 88: I am wondering that maybe we shall use different values for different catagory.
For example, Join/infra/Service may be 50 is enough. 100 for DKG might be OK. Better to have 200 for AE/NodeData ?

sn/src/messaging/serialisation/mod.rs
```diff
@@ -11,6 +11,14 @@ mod wire_msg_header;
 
 use xor_name::XorName;
 
+// highest prio as we can't do anything until we've joined
+pub(crate) const JOIN_RESPONSE_PRIORITY: i32 = 4;
+pub(crate) const DKG_MSG_PRIORITY: i32 = 3;
+pub(crate) const AE_MSG_PRIORITY: i32 = 2;
+pub(crate) const INFRASTRUCTURE_MSG_PRIORITY: i32 = 1;
+pub(crate) const NODE_DATA_MSG_PRIORITY: i32 = 0;
+pub(crate) const SERVICE_MSG_PRIORITY: i32 = -2;
```
Yoga07, line 20: Can we make these as enums and derive the priority from that? Eg: https://github.com/maidsafe/safe_network/pull/879/commits/a3053444321885d6e4e19fc1e920e43098123c7b#diff-d1322d94eedb660509242868fe2d1b0105651e01c6f4cbc319c0415b47023b14R32

sn/src/routing/routing_api/command.rs
```diff
@@ -106,6 +107,21 @@ pub(crate) enum Command {
...
 }
 
+impl Command {
+    /// Return the commands priority, higher being higher prio
+    pub(crate) fn priority(&self) -> Result<i32> {
+        let prio = match self {
+            Self::HandleMessage { wire_msg, .. } => wire_msg.into_message()?.priority(),
+            Self::HandleDkgOutcome { .. } | Self::HandleDkgFailure(_) => 3,
+            Self::HandleNewEldersAgreement { .. } => 2,
+            Self::HandleAgreement { .. } | Self::HandleSystemMessage { .. } => 1,
```
Yoga07: Maybe we can use the constants declared [here](https://github.com/maidsafe/safe_network/pull/789/commits/56e290fa99621e068f2598b8dc8e19f57b56c264#diff-7d31b982908d9bb02809f689fe05b137116a5ac46abfb11363023eb53a64dbf5R18)?

sn/src/routing/routing_api/dispatcher.rs
joshuef: Could be. I haven't seen any need of this as yet. I haven't seen us hit a queue of more than ~8 sempahores in any category aside from `service`.

 Happy to hear some reasoning for changing this.

sn/src/routing/routing_api/dispatcher.rs
joshuef: how long would you suggest?

sn/src/client/connections/messaging.rs
joshuef: We had it there. @lionel1704 moved it here because it's not always going to be known in there, I believe.

sn/src/routing/routing_api/dispatcher.rs
joshuef: If we're not sure, we can always look to change this in later PRs (and indeed make them non-magic)

sn/src/routing/routing_api/dispatcher.rs
joshuef: I've reduced to 50ms. (though worth noting this is all non-reachable without enabling the prioritisation `feat` atm, so we can tweak that later)

sn/src/messaging/serialisation/mod.rs
joshuef: Not cleanly as they have values...

sn/src/bin/testnet.rs
```diff
@@ -93,6 +93,10 @@ async fn main() -> Result<()> {
         let mut args = vec!["build", "--release"];
 
         // Keep features consistent to avoid recompiling when possible
+        if cfg!(feature = "unstable-command-prioritisation") {
+            args.push("--features");
+            args.push("unstable-command-prioritisation");
```
Yoga07, line 98: Follow up w/the above comment

sn/src/routing/core/comm.rs
```diff
@@ -131,7 +131,7 @@ impl Comm {
             wire_msg.set_dst_xorname(name);
 
             let bytes = wire_msg.serialize()?;
-            let priority = wire_msg.msg_kind().priority();
+            let priority = wire_msg.into_message()?.priority();
```
Yoga07, line 134: Maybe we can add a TODO here to find a way around deserialization later? Just to be cautious of the deserialization load every single time we send a message.

Yoga07: Or we can also pass on the priority from where we create the message maybe.

sn/Cargo.toml
```diff
@@ -37,6 +37,7 @@ required-features = ["test-utils"]
 [features]
 default = []
 always-joinable = []
+unstable-command-prioritisation = []
```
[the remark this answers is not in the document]
joshuef: We still need it to enable the `wait` portion. So current impl _will not wait_ for existing higher prio tasks. They all just go as long as there's a semaphore for that priority...  _unless_ it's a `ServiceMsg`, in which case it's dropped.

sn/src/messaging/serialisation/mod.rs
```diff
@@ -11,6 +11,14 @@ mod wire_msg_header;
 
 use xor_name::XorName;
 
+// highest prio as we can't do anything until we've joined
+pub(crate) const JOIN_RESPONSE_PRIORITY: i32 = 4;
```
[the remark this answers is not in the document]
joshuef: This is our node _receiving_ a join response. It cannot run DKG until joined. I don't think this is an issue...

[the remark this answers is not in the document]
Yoga07: JoinResponses are handled at https://github.com/maidsafe/safe_network/blob/4f600e179bfbf6ac018876cca6f7fc193f5b5f1e/sn/src/node/routing/core/bootstrap/join.rs#L164 This is a separate flow that is handled even before we set up our `Routing::Core`. So our priorities do not come into play here.

Yoga07 (dismissed):

Yoga07 (dismissed):

PR #792, opened 2021-11-29
fix: avoid use outdated keyshare status

maqi:

### DIFF
--- a/sn/src/routing/network_knowledge/mod.rs
+++ b/sn/src/routing/network_knowledge/mod.rs
@@ -254,37 +254,6 @@ impl NetworkKnowledge {
         }
     }
 
-    pub(super) async fn switch_to_sap(
-        &self,
-        signed_sap: SectionAuth<SectionAuthorityProvider>,
-        our_name: &XorName,
-    ) -> Result<()> {
-        let provided_sap = signed_sap.value.clone();
-        // We try to update our SAP and own chain only if we were flagged to,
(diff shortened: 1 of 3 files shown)


bochaco (approved):

sn/src/routing/network_knowledge/mod.rs
```diff
@@ -289,11 +289,22 @@ impl NetworkKnowledge {
...
                     .join(proof_chain.clone())?;
 
+                let switch_to_new_sap = (!self.is_elder(our_name).await
+                    && !provided_sap.contains_elder(our_name))
+                    || section_keys_provider
+                        .key_share(&signed_sap.section_key())
+                        .await
+                        .is_ok();
+                trace!(
+                    "update_knowledge_if_valid switch_to_new_sap {:?}",
```
bochaco, line 299: Perhaps this can be improved with something more human readable? e.g. "Are we switching to new SAP? {}" ?

sn/src/routing/network_knowledge/mod.rs
```diff
@@ -289,11 +289,22 @@ impl NetworkKnowledge {
...
+                    switch_to_new_sap
+                );
+
                 // We try to update our SAP and own chain only if we were flagged to,
```
bochaco, line 303: This is slightly outdated with this PR, "if we were flagged to," is not applicable anymore

bochaco (changes requested):

PR #794, opened 2021-11-29
Nrs on multimap

grumbach:
- NRS with multimaps instead or registers, saving both network space and ops
- restore multimap remove and tombstones
- manage register fork resolution and add informative errors

### DIFF
--- a/sn_api/src/app/nrs/mod.rs
+++ b/sn_api/src/app/nrs/mod.rs
@@ -9,14 +9,15 @@
 
 mod nrs_map;
 
-pub use crate::url::{ContentType, VersionHash};
-pub use nrs_map::NrsMap;
+pub use crate::url::{ContentType, VersionHash, DataType};
+pub use nrs_map::{NrsMap};
+
+use nrs_map::{parse_out_subnames, SubName};
 
 use crate::{app::Safe, register::EntryHash, Error, Result, Url};
 
-use bytes::{Buf, Bytes};
-use log::{debug, info, warn};
-use std::collections::BTreeSet;
+use log::{debug, info};
+use std::collections::{BTreeSet, BTreeMap};
 use std::str;
 
 // Type tag to use for the NrsMapContainer stored on Register
@@ -35,9 +36,8 @@ impl Safe {
     ///            Public Name
     /// ```
     /// Registers the given NRS top name on the network.
-    /// Returns the versioned NRS Url (containing a VersionHash)
-    /// `safe://{top_name}?v={version_hash}`
-    /// Note that this NRS Url is not linked yet to anything. You just registered the topname here.
+    /// Returns the NRS Url: `safe://{top_name}
+    /// Note that this NRS Url is not linked to anything yet. You just registered the topname here.
     /// You can now associate public_names (with that topname) to links using `nrs_associate` or `nrs_add`
     pub async fn nrs_create(&self, top_name: &str, dry_run: bool) -> Result<Url> {
         info!("Creating an NRS map for: {}", top_name);
@@ -56,23 +56,14 @@ impl Safe {
             ));
         }
 
-        // Create and store NrsMap for top_name in a Public Blob
-        let nrs_map = NrsMap::default();
-        let nrs_map_xorurl = self.store_nrs_map(&nrs_map).await?;
-
-        // Create a new Register and reference the NRS map's blob in the Register
-        let reg_xorurl = self
-            .register_create(Some(nrs_xorname), NRS_MAP_TYPE_TAG, false)
-            .await?;
-        let reg_entry = nrs_map_xorurl.as_bytes().to_vec();
-        // Note that we can use the higher level register API here
-        // because reg_xorurl is not an NRS url
-        // (high level register API uses resolution that uses NRS!)
-        let entry_hash = self
-            .register_write(&reg_xorurl, reg_entry, BTreeSet::new())
+        // Create a new empty Multimap
+        let _ = self
+            .multimap_create(Some(nrs_xorname), NRS_MAP_TYPE_TAG, false)
             .await?;
 
-        Ok(get_versioned_nrs_url(url_str, entry_hash)?)
+        let mut url = Url::from_url(&url_str)?;
+        url.set_content_type(ContentType::NrsMapContainer)?;
+        Ok(url)
     }
 
     /// # Associates a public name to a link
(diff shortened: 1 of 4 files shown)


sn_api/src/app/nrs/mod.rs
```diff
@@ -230,82 +224,80 @@ impl Safe {
...
-            .store_public_bytes(Bytes::from(serialised_nrs_map), None, false)
-            .await?;
+        // check for duplicate entries for a single key (subname)
+        let set_len = raw_set.len();
+        let raw_map: BTreeMap<Vec<u8>, Vec<u8>> = raw_set.clone().into_iter().collect();
+        if raw_map.len() != set_len {
+            let set_after_map: BTreeSet<(Vec<u8>, Vec<u8>)> = raw_map.into_iter().collect();
+            let dups: Vec<(Vec<u8>, Vec<u8>)> =
+                raw_set.difference(&set_after_map).cloned().collect();
+            return Err(Error::ConflictingNrsEntries("Found multiple entries for the same name, this happens when 2 clients write concurrently to the same NRS mapping. It can be fixed by simply associating a new link to the conflicting names.".to_string(), dups));
```
bochaco, line 287: Do we want to return an error? I think this is a valid scenario the user just needs to decide what to do about, but it's not an error. So perhaps the API shall just return a vec of values which would contain only one element when there is no forks, or more elements otherwise, the user can check that and fix the fork when it detects more than one element....

bochaco: ...also, there can be more than two concurrently written values.

grumbach: Are you suggesting to change the NrsMap (a `BTreeMap` that maps subnames to Urls) into a multimap (map that allows multiple values for the same key)? Does that make sense for Nrs? How would the resolution choose between multiple entries? 
```
"subname" -> url1
"subname" -> url2
```
I think that it is a normal scenario for registers to fork, but in the case of NRS, having multiple values for a single subname should be an error. Although it is very easy to fix now (just `nrs_associate("subname", url3)`). 
What do you think?

grumbach: Indeed, `dups` is not a great variable name for conflicting entries. How about simply `conflicting_entries`?

bochaco: ...right, i see your point, at the same time a user querying (let's say from the browser) an NRS name by visiting a site shall be shown either one of the sites or the list of conflicting sites and allow him/her to choose? ...
...If the user visits such a website, I guess it wouldn't be fair to just show an error saying "sorry we cannot show you the website because there are two urls resolved", but actually it would be correct to show at least one of them without being able to resolve them (because that's just a visitor and not the NRS name owner)....
...is that really an Error scenario or just bad luck for the user/s when concurrent updates has been written to a site/nrs-name while querying it...?.....

grumbach: Yes agreed, the user shouldn't see the error and browsers must have a way to recover from this scenario. 
Maybe we could include the `NrsMap` in the `ConflictingNrsEntries` error so that apps and browsers can (if they choose to treat this as a normal case) fully recover from it as you suggested?
My fear with changing the API to accept multiple conflicting entries is that is will pollute the entire resolution code with this and one will eventually have to manage it anyway (consider error or make a choice). 
By including the `NrsMap` we can have the best of both worlds, either cases return an `NrsMap`, the API stays intact and browsers/apps can freely choose to treat conflicts as errors or not. 
What do you think?

bochaco: > could include the NrsMap in the ConflictingNrsEntries 
why not just return it as an Ok result. This is also some of the things we need to embrace, this is not DNS and this is not any other type of database, cloud file storage, this is a decentralised network with CRDT as its native data types, trying to hide or masquerade it is more confusing and misleading than simply exposing the nature of it in the APIs. 

The main issue is we will never know what the user/application that is behind the API is willing to do or how to deal with concurrent values, we cannot even assure concurrent values are conflicts or unwanted scenario, it sounds like not even in the context of NRS. E.g. I have a safepedia site where I allow anyone to edit a page and I (the owner) decide to merge them after reviewing changes once a week. What do you think I'd expect to happen to visitors querying the website when there are 5 forks of the page? fir st answer is "you don't know, because I've created that site and the semantic or the UX is defined/decided by me, not by the creator of the APIs", second answer could be "show the page before the forks", still you are making a decision which may not be correct, so what should the browser do? for sure it cannot assume that's an error, because it depends on what the data is and which app is writing it, so for the browser consuming the API, why should see concurrent values returned as Error(...) ?

....just random thoughts, I'd also like to create an NRS abstraction which resolves ass much as it can for users, but not sure if this can be resolved without knowing the context of the app owning the data.....

joshuef: > will pollute the entire resolution

It may help to think of the overall resolution of a url is an app atop Nrs resolution (or that as a function inside of resolution). If NRS returns everything relevant, the overall resolver can act on this (or feed back on it).

I wouldn't call this pollution, so much as just exposing the underlying uncertainty of the data. 

The resolver could have some default behaviour (go with longest fork chain... go with fork initiated by owner... wahtever) which smoothes out the overall UX, and perhaps offer feedback on alternatives at that layer

grumbach: I totally concur with your browser examples. I think that what you are proposing involves changing the `NrsMap` type to a type that can support concurrent values. This will require changes and decisions in the resolver code that would require taking core design decisions. 
Although I agree register forks on their own are a completely normal case, register forks in the NRS scope are hard to reproduce and there is no way using the regular NRS API to create one. The only way is doing concurrent writes and hoping a race condition creates a fork (but that's not normal usage). That's why I think it's more an error than a Ok even though I agree it should be recoverable from. 
This PR is about optimising NRS. But it doesn't change the functionality. It keeps the one to one mapping but enhances the fork errors to make them recoverable from. 
I think the change you are suggesting is of greater scope, as it involves changing the design of the resolver and the features of NRS. I agree that the browser/app side should be able to choose their behaviour on their own. I suggest we move this to discourse, I think that's an important question and probably out of scope for this PR.

bochaco: > The only way is doing concurrent writes and hoping a race condition creates a fork (but that's not normal usage). That's why I think it's more an error than a Ok even though I agree it should be recoverable from.

This holds true only if you are using these set of APIs, but he user could be writing NRS maps using any other API which allows fork to be created by the user on purpose. This is why we can never know if this was an accident and not a normal usage. Our fetch API will still need to handle them properly.

PR #796, opened 2021-11-30
Fix dont fail on join sig agg

joshuef:

### DIFF
--- a/sn/src/routing/core/bootstrap/join.rs
+++ b/sn/src/routing/core/bootstrap/join.rs
@@ -223,6 +223,26 @@ impl<'a> Join<'a> {
                             node_state_serialized
                         };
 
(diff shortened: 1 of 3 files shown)


sn/src/routing/core/bootstrap/join.rs
```diff
@@ -258,7 +278,13 @@ impl<'a> Join<'a> {
                             continue;
                         }
                         Err(AggregatorError::NotEnoughShares) => continue,
-                        _ => return Err(Error::InvalidSignatureShare),
+                        error => {
+                            warn!(
+                                "Error received as part of signature aggregation during join: {:?}",
+                                error
+                            );
+                            continue;
```
Yoga07, line 286: Should we maybe start over here too? By sending the JoinRequest again?

Yoga07 (approved):

PR #804, opened 2021-12-01
Remove DkgUnderway, Aggregate join sigs per key

joshuef:

### DIFF
--- a/sn/src/messaging/system/join.rs
+++ b/sn/src/messaging/system/join.rs
@@ -102,6 +102,4 @@ pub enum JoinRejectionReason {
     JoinsDisallowed,
     /// The requesting node is not externally reachable
     NodeNotReachable(SocketAddr),
-    /// Backoff as the section is undergoing DKG
-    DkgUnderway,
 }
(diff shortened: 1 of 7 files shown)


sn/src/routing/core/msg_handling/join.rs
```diff
@@ -94,20 +94,6 @@ impl Core {
             ]);
         }
 
-        if self.dkg_voter.has_dkg_underway() {
```
maqi, line 97: not sure about removing this toally.
As I understand, we shall change to: accept JoinRequest with aggregated OnlineApproval even has ongoing dkg.

joshuef: I don't think we need it for anything else? To my mind it serves to prevent issues at signature aggregation which we solve otherwise in this PR.

sn/src/messaging/system/join.rs
```diff
@@ -103,5 +103,5 @@ pub enum JoinRejectionReason {
     /// The requesting node is not externally reachable
     NodeNotReachable(SocketAddr),
     /// Backoff as the section is undergoing DKG
-    DKGUnderway,
+    DkgUnderway,
```
bochaco, line 106: Shouldn't this be removed?

bochaco (approved):

PR #807, opened 2021-12-01
refactor(sn_api): refactoring files API tests and re-enabling them to run in CI

bochaco:


grumbach (approved):
Got a personal preference for `recursive_resolve_url`, but not blocking. LGTM

PR #809, opened 2021-12-02
Fixes to DKG Anti-entropy

lionel1704:
- discard DKG session info for older sessions
- skip DkgStart and DkgRetry messages for expired sessions


sn/src/routing/core/msg_handling/dkg.rs
```diff
@@ -33,6 +33,11 @@ impl Core {
         prefix: Prefix,
         elders: BTreeMap<XorName, SocketAddr>,
     ) -> Result<Vec<Command>> {
+        let current_generation = self.network_knowledge.chain_len().await;
+        if session_id.generation < current_generation - 1 {
```
maqi, line 37: is it shall be `if session_id.generation < current_generation ` ?
same to the other checking places.

joshuef (approved):

PR #812, opened 2021-12-02
fix(dkg): accept DKGOutcome only if elected elders are active members

Yoga07:


sn/src/routing/core/msg_handling/dkg.rs
```diff
@@ -231,13 +231,29 @@ impl Core {
...
     ) -> Result<Vec<Command>> {
+        // Check if the elected Elders are still `active members` in our section.
+        for elder in sap.elders_vec().iter() {
+            if !self
+                .network_knowledge
+                .active_members()
+                .await
+                .contains(elder)
+            {
+                error!("Dropping a DKGOutcome as one of the participants is not an active member in our section");
```
maqi, line 242: we could being lagged to receive a signed Online of an active member.
So, in this case, we cannot just drop the outcome.
instead, I think we shall `keep it but not switch to it`
then on each signed Online received, check whether it's now OK to switch to it.
It shall only get dropped, when there is new DKG outcome or SAP is available

Yoga07:
Will be covered by the upcoming BRB Membership refactor.

PR #819, opened 2021-12-06
Register op batch, dry runs

grumbach:
- implement basic layout for register operation batching as a WriteAheadLog (`sn` and `sn_api`'s low level register API)
- implement (or fix broken/unimplemented) dry runs for:
    - NRS
    - FileContainers
    - Registers (high level register API)
    - Multimap

### DIFF
--- a/sn/src/client/client_api/register_apis.rs
+++ b/sn/src/client/client_api/register_apis.rs
@@ -16,18 +16,58 @@ use crate::types::{
     },
     PublicKey, RegisterAddress as Address,
 };
+use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet};
 use xor_name::XorName;
+
+/// Register Operation Batch
+///
+/// Batches up register write operation before publishing them up to the network.
+/// Can also be used as a way to implement dry runs:
+/// nothing is uploaded to the network as long as the batch is not published.
+/// Batches can be republished without duplication risks thanks to the CRDT nature of registers.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct RegisterOpBatch {
+    batch: Vec<DataCmd>,
+}
+
+impl RegisterOpBatch {
+    /// Creates a new RegisterOpBatch from a register data cmd
+    pub(crate) fn new(cmd: DataCmd) -> RegisterOpBatch {
+        RegisterOpBatch { batch: vec![cmd] }
+    }
+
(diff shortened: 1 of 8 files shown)


sn/src/client/client_api/register_apis.rs
```diff
@@ -16,19 +16,70 @@ use crate::types::{
...
 impl Client {
     //----------------------
     // Write Operations
     //---------------------
 
+    /// Publish all register mutation operations in a WAL to the network
+    /// Incrementing the WAL index as successful writes are sent out. Stops at the first error.
+    /// Starts publishing from the index when called again with the same WAL.
+    #[instrument(skip(self), level = "debug")]
+    pub async fn publish_register_ops(&self, wal: &mut RegisterWriteAheadLog) -> Result<(), Error> {
```
bochaco, line 64: Do we really need the `RegisterWriteAheadLog` ? why don't we let the caller to clean it up, remove successful commands form it when retrying instead of having to keep an index in that struct. I think we should just have a Vec<DataCmd> as arg here, and let the caller/user of this API do the rest.

grumbach: Yeah I was wondering about that index, maybe it's not necessary, it even requires a mut. It could well be const, since we're dealing with CRDT ops, we could even send them all again without a problem. 

Although we're indeed dealing with a `Vec<DataCmd>` I think we need to hide the underlying sn structs with an opaque type at the API level. That's why I created that struct with a private vec field. We could add iterators or other methods to it so that users can iterate through the operations (still opaque type) to send them until an error occurs. 
I use to call that struct a RegisterOpBatch, then we called it a WriteAheadLog and added an index. 
Should I remove that index? Or rename that struct?
What do you think?

bochaco: I think we don't need an struct, or any wrapper as of yet, the Vec already has an iterator and the user can keep/discard/retry the commands that failed as wished, I think at least till we really expand our APIs for batch ops utilities we don't need it.

grumbach: Ok then, if we're gonna abstract it away at some point (with expanded batching API), let's do this.

bochaco: we definitely will as soon as we need to, expanding before we really need it would just add maintenance costs for us, specially once community start using it we'll need more effort to provide support for things that are not strictly necessary, that's how i view it at least

bochaco (approved):

PR #820, opened 2021-12-07
fix: complete the relocation flow

maqi:

### DIFF
--- a/sn/src/routing/core/bootstrap/relocate.rs
+++ b/sn/src/routing/core/bootstrap/relocate.rs
@@ -15,11 +15,10 @@ use crate::messaging::{
     AuthorityProof, DstLocation, SectionAuth, WireMsg,
 };
 use crate::routing::{
-    dkg::SectionAuthUtils,
     ed25519,
     error::{Error, Result},
     messages::WireMsgUtils,
-    network_knowledge::{NetworkKnowledge, SectionAuthorityProvider},
+    network_knowledge::SectionAuthorityProvider,
     node::Node,
     relocation::RelocatePayloadUtils,
     routing_api::command::Command,
@@ -32,7 +31,7 @@ use xor_name::{Prefix, XorName};
 
 /// Re-join as a relocated node.
 pub(crate) struct JoiningAsRelocated {
-    node: Node,
+    pub(crate) node: Node,
     genesis_key: BlsPublicKey,
     dst_section_key: BlsPublicKey,
     relocate_details: RelocateDetails,
@@ -51,12 +50,14 @@ impl JoiningAsRelocated {
         node_msg: SystemMsg,
         section_auth: AuthorityProof<SectionAuth>,
     ) -> Result<Self> {
-        let dst_section_key = relocate_details.dst_key;
-
+        // First JoinAsRelocatedRequest doesn't contain RelocatePayload,
+        // which triggers a Response being sent back anyway.
+        // Setting dst_section_key to correct one could cause the check to fail
+        // when handling such response.
         Ok(Self {
             node,
             genesis_key,
-            dst_section_key,
+            dst_section_key: genesis_key,
             relocate_details,
             node_msg,
             node_msg_auth: section_auth,
@@ -83,7 +84,7 @@ impl JoiningAsRelocated {
         // we will then use to generate the relocation payload and send the
         // `JoinAsRelocatedRequest` again with it.
         // TODO: include the section Prefix in the RelocationDetails so we save one request.
-        self.build_join_request_cmd(&recipients)
+        self.build_join_request_cmd(&recipients, dst_xorname)
     }
 
     // Handles a `JoinAsRelocatedResponse`, if it's a:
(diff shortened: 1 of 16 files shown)


sn/src/routing/core/bootstrap/relocate.rs
```diff
@@ -51,12 +50,14 @@ impl JoiningAsRelocated {
...
-
+        // First JoinAsRelocatedRequest doesn't contain RelocatePayload,
+        // which triggers a Response being sent back anyway.
+        // Setting dst_section_key to correct one could cause the check to fail
+        // when handling such response.
         Ok(Self {
             node,
             genesis_key,
-            dst_section_key,
+            dst_section_key: genesis_key,
```
joshuef, line 60: This feels wrong to me. We're looking to trigger an AE Response?

joshuef: What do we need from it we can't just return with the normal request?

sn/src/routing/core/msg_handling/mod.rs
```diff
@@ -411,12 +416,146 @@ impl Core {
...
+                                        signed_sap,
+                                        Some(prefix_map),
+                                    )?;
+                                    let _ = new_network_knowledge
+                                        .merge_members(SectionPeers::from_msg(members));
+
+                                    // TODO: confirm whether carry out the switch immediately here
+                                    //       or still using the Command pattern.
+                                    //       As the sending of the JoinRequest as notification
+                                    //       may require the `node` to be switched to new already.
```
joshuef, line 491: > sending of the JoinRequest as notification
It's not clear what we mean here...

Is there not a final `join request approved` received from our new section that'd trigger a switch?

maqi: This is to be consistent with the new Join flow, i.e. the joining node becomes the only aggregator, existing elders only sending ApprovalShare.
Hence here, once aggregation completes, the `switch` is triggered. There is no more final Approval received from elders to trigger the switch.

sn/src/routing/core/bootstrap/relocate.rs
maqi: This is the first JoiningAsRelocatedRequest sent from the relocated node to the target section, which doesn't contain the RelocatePayload
For the JoiningAsRelocatedResponse received from the target section, the relocated node needs to generate RelocatePayload correspondently and resend JoiningAsRelocatedRequest with the payload. Only with this second request, shall the dst_section_key being the correct one. 
When the dst_section_key is being the correct one at the beginning, the check at https://github.com/maidsafe/safe_network/blob/main/sn/src/routing/core/bootstrap/relocate.rs#L146 will be failed when received the JoinAsRelocatedResponse for the first request.

joshuef: Aye, so my Q is perhaps better phrased as:
> the relocated node needs to generate RelocatePayload correspondently 
What stops us from doing this first time round?

sn/src/routing/core/msg_handling/mod.rs
joshuef: at the mo we have `ReceivedJoinApproved` from the standard bootstrap flow, which (perhaps needlessly) is sent from the node after _that_ received the aggregated joinapproval: https://github.com/maidsafe/safe_network/blob/main/sn/src/routing/core/bootstrap/join.rs#L171-L178

sn/src/routing/core/bootstrap/relocate.rs
maqi: >What stops us from doing this first time round?

The prefix of the target section we known could be incorrect or even unknown, say the relocated node is restarted.
Hence, the first Request is served as a probe to encure our knowledge of the target section is the latest one.

joshuef: Hmm, if it's unknown, we could provide `Prefix()`, which should reject/retry no?

I'm not sure why we'd purposefully prvide the `genesis pk` here if we don't need to? We _may_ be correct and can avoid one messaging loop, no?

sn/src/routing/core/msg_handling/mod.rs
maqi: right, I thought that extra Approval is not necessary.
But if need to follow the Join flow, I can modify the flow to add this in

sn/src/routing/core/bootstrap/relocate.rs
maqi: The extra one messaging loop can only be avoided when we know the target section's latest SAP already.
As I understand, the current impl is just trying to convenge two flows of `known` and `unknown` into one for a consitency.

I could refactor the checking criterias to make the first shot being correct. However, this may incur some extra flow branches.

sn/src/routing/core/msg_handling/mod.rs
joshuef: Okay, so no need for that as network knowledge differs in these two flows

sn/src/routing/core/bootstrap/relocate.rs
joshuef: Okay no I see where you're coming from now. It's not that we send to, but expect geensis, and it isn't what we get so then then we can carry on there. 👍

joshuef (approved):

PR #822, opened 2021-12-08
Various Updates for NRS tests and CLI Commands

jacderida:
This PR probably turned out to be a bit bigger than it should be. Should have probably submitted each of the command updates individually, but here it is.

- ea7ba8242 **feat(sn_cli)!: `nrs create` only creates topnames**

  BREAKING CHANGE: this command should now be used to create new topnames, but *not* a subname and a
  new topname at the same time. The user is advised on this by way of suggestions in error messages.

  The user experience of this command was also improved in the following ways:
  * Validate URLs supplied with the `--link` argument *before* creating the topname. This is to ensure
    the user doesn't receive an error message after the topname was already created, which can
    potentially be quite confusing.
  * Suggest supplying valid XorUrls for the `--link` argument.
  * Suggest that links to versionable content must have their version hash specified on the `--link`
    argument.
  * Suggest the use of the `add` command with the `--create-top-name` argument to create a subname and
    a topname at the same time, or use the `create` command to first create the topname, then use the
    `add` command to create a subname.
  * Suggest trying a different name if the chosen topname is already registered.
  * The grammar of a couple of error messages in the API were updated for improved readability.

  Some behaviour was removed:
  * Since dry run support is still in progress for NRS, the `--dry-run` code was removed.
  * The `--direct` argument was removed, since it only really applies to subnames, which you can no
    longer supply with the `create` command. This is a BREAKING CHANGE.

  The initial intention was to fix up the failing NRS tests for the CLI. These failures were due
  to the fact that the API for NRS had changed, which was resulting in the commands behaving a bit
  differently than they used to. The main difference being the fact that the NRS API doesn't allow the
  creation of subnames along with a topname.

  I took the opportunity to introduce new tests that should cover all the scenarios for the `create`
  command.

- 587b65ba5 **feat(sn_cli): refined ux and behaviour for `nrs add`**

  The behaviour for the command is mostly retained, but the user experience has been improved in the
  following ways:

  * The documentation for the command and the arguments were elaborated to provide more information.
  * Text was added to the command output when the `--default` argument is used, to indicate that a
    link was created to the topname.
  * Suggest that links to versionable content must have their version hash specified on the `--link`
    argument.

  BREAKING CHANGES:
  * The `--name` argument has been renamed to `--public-name`.
  * The `--direct` flag argument was removed.

  Tests were provided for all of the scenarios and code paths explored in the command and there was
  some refactoring done to share some code between the `create` and `add` commands. The calls to
  `nrs_add` and `nrs_associate` were wrapped in functions, due to the slightly verbose error handling
  required for providing suggestions to the user.

  Finally, the dry run functionality was removed since this currently isn't available for NRS. We can
  add it back in when it is.

- 8dc24d829 **tests: add coverage for `nrs remove` command**

  There are no behavioural changes in the command, but there was a little bit of code refactoring and
  some improved user experience by way of error messages and suggestions.

  We need to come back and revisit some odd behaviour with this command, related to removing topnames
  and attempting to remove subnames that don't exist.

- dafbcb08d **feat(sn_cli): rename nrs `create` to `register`**

  BREAKING CHANGE: this command now has a different name.

  We decided it was worth having increased distinction between subname and topname management, where
  `register` and `unregister` will be for topname management, and `add` and `remove` will be for
  subname management. Behaviourally, everything is the same.

  Arguments and string references were replaced with the relevant verbs.

- 0b02aef2c **tests: remove previous nrs tests**

  To me, these tests were lacking a clear purpose and they didn't map directly onto any of the
  commands. They should now be made obsolete by the new tests, which map to specific commands and
  scenarios.
  
- 16077ebc1 **chore: address feedback from pr**

  * Remove use of unwrap on topname parsing and put the duplicated code in a function
  * Refer to 'url' rather than 'XorUrl'  

Here are the test results:
```
running 19 tests
test nrs_add_should_add_a_subname_and_a_new_topname ... ok
test nrs_add_should_add_a_subname_and_behave_idempotently_for_existing_topname ... ok
test nrs_add_should_add_a_subname_and_set_it_as_the_default_for_the_topname ... ok
test nrs_add_should_add_a_subname_to_immutable_content ... ok
test nrs_add_should_add_a_subname_to_versioned_content ... ok
test nrs_add_should_return_an_error_if_an_invalid_link_is_specified ... ok
test nrs_add_should_return_an_error_if_link_to_versioned_content_has_no_version ... ok
test nrs_add_should_update_an_existing_subname ... ok
test nrs_add_with_default_should_return_an_error_if_link_to_versioned_content_has_no_version ... ok
test nrs_add_with_register_top_name_should_return_an_error_if_link_to_versioned_content_has_no_version ... ok
test nrs_register_should_register_a_topname ... ok
test nrs_register_should_register_a_topname_with_a_versioned_content_link ... ok
test nrs_register_should_register_a_topname_with_an_immutable_content_link ... ok
test nrs_register_should_return_an_error_if_a_subname_is_specified ... ok
test nrs_register_should_return_an_error_if_an_invalid_link_is_specified ... ok
test nrs_register_should_return_an_error_if_link_to_versioned_content_has_no_version ... ok
test nrs_register_should_return_an_error_if_the_topname_already_exists ... ok
test nrs_remove_should_remove_a_subname ... ok
test nrs_remove_should_return_an_error_for_a_non_existent_topname ... ok

test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 891.18s

```

They all run consistently without random failures, though they are quite slow.


sn_cli/src/subcommands/nrs.rs
```diff
@@ -63,105 +66,261 @@ pub async fn nrs_commander(
...
+    }
+}
+
+/// Determine if the link is a valid XorUrl *before* creating the topname.
+///
+/// Otherwise the user receives an error even though the topname was actually registerd, which is a
+/// potentially confusing experience: they may think the topname wasn't registerd.
+fn validate_target_link(link: &str) -> Result<()> {
+    Url::from_url(link)
+        .wrap_err("The supplied link was not a valid XorUrl.")
```
grumbach, line 246: The link can be any safe Url (xor or nrs), not necessarily a xorurl.

sn_cli/src/subcommands/nrs.rs
```diff
@@ -63,105 +66,261 @@ pub async fn nrs_commander(
...
+    }
+    let version = url
+        .content_version()
+        .ok_or_else(|| eyre!("Content version not set for returned NRS Url"))?
+        .to_string();
+    summary_header.push_str(&format!("Now at version {}. ", version));
+
+    if default {
+        let mut parts = name.split('.');
+        let topname = parts.next_back().unwrap();
```
grumbach, line 180: Although its probably safe because of the above, do we really want to unwrap?

bochaco: no, we don't use `unwrap` anywhere, we always handle Results

sn_cli/src/subcommands/nrs.rs
jacderida: Ah ok, thanks! Will change this.

sn_cli/src/subcommands/nrs.rs
jacderida: Ah, I think I meant to come back and change this, but obviously have forgotten. Thanks, will get it sorted.

sn_cli/src/subcommands/nrs.rs
jacderida: This was addressed now.

sn_cli/src/subcommands/nrs.rs
jacderida: Removed the unwrap and put the duplicated code in a function.

grumbach (dismissed):

joshuef (approved):

joshuef (approved):

PR #838, opened 2021-12-13
ci: run a droplet testnet daily on main

joshuef:
This should run a live WAN testnet, across many droplets whenever we push to the given branch

### DIFF
--- a/.github/workflows/daily_droplet_run.yml
+++ b/.github/workflows/daily_droplet_run.yml
@@ -1,9 +1,10 @@
 name: 'sn: Run Client Tests against DO Testnet'
 
 on:
-  push:
-    branches:
-      - staging
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '30 2 * * *'
+
 
 env:
   CARGO_INCREMENTAL: '0'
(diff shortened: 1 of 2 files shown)


.github/workflows/daily_droplet_run.yml
```diff
@@ -0,0 +1,111 @@
...
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_DEFAULT_REGION: 'eu-west-2'
+
+jobs:
+  launch-testnet:
+    name: Launch Digital Ocean testnet
+    runs-on: ubuntu-latest
+    steps:
+      - name: Launch testnet
+        uses: joshuef/sn_testnet_action@ImproveAction
```
lionel1704, line 29: Perhaps we should PR the improvement to maidsafe/sn_testnet_action?

.github/workflows/daily_droplet_run.yml
```diff
@@ -0,0 +1,111 @@
+name: 'sn: Run Client Tests against DO Testnet'
+
+on:
+  schedule:
```
lionel1704, line 4: All branches?

.github/workflows/daily_droplet_run.yml
joshuef: ah, that should be in. that needs updated thanks @lionel1704 👍

.github/workflows/daily_droplet_run.yml
joshuef: ah, i'd understood `schedule` defaulted to the default branch. Lemme double check docs

joshuef: Aye, according to this it's whatever is `default`: https://stackoverflow.com/questions/58798886/github-actions-schedule-operation-on-branch

lionel1704 (approved):

PR #839, opened 2021-12-13
DO NOT MERGE

maqi:

### DIFF
--- a/sn/src/bin/sn_node.rs
+++ b/sn/src/bin/sn_node.rs
@@ -118,7 +118,9 @@ async fn run_node() -> Result<()> {
         if config.json_logs {
             builder.json().init();
         } else {
-            builder.event_format(LogFormatter::default()).init();
+            builder
+            // .event_format(LogFormatter::default())
+            .init();
         }
 
         Some(guard)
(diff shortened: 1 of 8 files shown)


.github/workflows/sn_pr_tests.yml
```diff
@@ -235,10 +235,16 @@ jobs:
         run: ./target/release/log_cmds_inspector $HOME/.safe/node/local-test-network
         if: steps.section-startup.outcome == 'success' &&  matrix.os != 'windows-latest'
 
-      - name: Upload Node Logs to AWS
+      - name: Upload Node Logs to AWS for Windowns
```
Yoga07, line 238: I think this task will run on all platforms? But will error out on non-windows since %userprofile% can't be found.

sn/src/bin/sn_node.rs
```diff
@@ -118,7 +118,9 @@ async fn run_node() -> Result<()> {
         if config.json_logs {
             builder.json().init();
         } else {
-            builder.event_format(LogFormatter::default()).init();
+            builder
+            // .event_format(LogFormatter::default())
```
Yoga07, line 122: Why are we commenting this out?

maqi: it's just I feel the other logging format is more convenient.
Not part of the CI logging issue.

.github/workflows/sn_pr_tests.yml
maqi: the uploading task won't got carried out. The later on OS checking criteria will prevent it.
There might be a better way to divert tasks depend on OS?

PR #840, opened 2021-12-13
refactor(prefix_map): clients and node read/write to a global prefix_map

Yoga07:


bochaco (commented):
How about using the genesis key rather than client's pk to name the prefixmap's file or folder? eventually you'd be using different and several client keys to interact with the network, some keys will be used for signing some data, while other keys for transfers, etc., it'll be up to the user and more likely many keys will be used. Plus, a prefix map differs from another really just due to its genesis key, if they share the genesis key then any new prefix can be merged/saved

Yoga07:
> How about using the genesis key rather than client's pk to name the prefixmap's file or folder? eventually you'd be using different and several client keys to interact with the network, some keys will be used for signing some data, while other keys for transfers, etc., it'll be up to the user and more likely many keys will be used. Plus, a prefix map differs from another really just due to its genesis key, if they share the genesis key then any new prefix can be saved onto it....

We discussed this a lot on slack. Let's take this over there too :)

joshuef:
The outcome of that discussion was something like `~/.safe/prefixmaps/<genesiskey>`, right? Which could be read/written by any client/node?

Yoga07:
Yes, a global folder for PrefixMaps in which every PrefixMap will be stored in a separate file and their names suffixed with 
 their genesis key i.e. `prefixmap-<GenesisKey>`. And so every node/client will be using that folder to find the required PrefixMap and read/write to it accordingly.

joshuef (dismissed):

PR #850, opened 2021-12-14
feat(client): add a Chunks cache on client API to prevent retrieving an already retrieved Chunk

bochaco:


sn/src/client/client_api/blob_apis.rs
```diff
@@ -110,6 +108,17 @@ impl Client {
 
     #[instrument(skip(self), level = "trace")]
     pub(crate) async fn get_chunk(&self, name: &XorName) -> Result<Chunk> {
+        // first check it's not already in our Chunks' cache
+        if let Some(chunk) = self
+            .chunks_cache
+            .write()
```
Yoga07, line 114: Do we need `write` access here?

sn/src/client/client_api/mod.rs
```diff
@@ -33,10 +33,13 @@ use tokio::{
     time::Duration,
 };
 use tracing::{debug, info};
+use uluru::LRUCache;
 use xor_name::XorName;
 
 pub use register_apis::RegisterWriteAheadLog;
 
+type ChunksCache = LRUCache<Chunk, 50>;
```
Yoga07, line 41: Better to make the size a const?

sn/src/client/client_api/blob_apis.rs
bochaco: yes, it's because LRU cache marks the item when you read it putting it to the top of the lit of most used/read items, remember this is Least Recently Used cache.

sn/src/client/client_api/blob_apis.rs
```diff
@@ -191,15 +202,9 @@ impl Client {
...
-        let query = DataQuery::GetChunk(ChunkAddress(*address.name()));
-        match self.send_query(query).await {
-            Ok(QueryResult {
-                response: QueryResponse::GetChunk(Ok(_)),
-                ..
-            }) => Ok(address),
-            Ok(QueryResult { response, .. }) => Err(Error::UnexpectedQueryResponse(response)),
-            Err(err) => Err(err),
-        }
+        let _ = self.get_chunk(address.name()).await?;
```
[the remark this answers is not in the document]
bochaco: i wouldn't think so, that's what the retry will take care of....

PR #851, opened 2021-12-14
chore(release): sn_cli-v0.39.0

jacderida:
This release is being performed 'manually' so that the first sn_cli crate can be published to crates.io. From that point forward, `smart-release` should hopefully be able to pick up the changes in the CLI.

The changelog entry here is a manual one that's been done from memory, so it may well be missing certain things. Moving forward, the changelog will be generated automatically.


jacderida:
I've discovered that unfortunately the CLI is referencing another crate with a git branch. This will need to be replaced first.

PR #861, opened 2021-12-16
chore(client): removing exponential backoff when retrying queries

bochaco:


sn/src/client/utils/test_utils/mod.rs
```diff
@@ -44,6 +45,25 @@ where
     }
 }
 
+fn retry<R, E, Fn, Fut>(
```
Yoga07, line 48: Just curious as to why this was moved to another module?

sn/src/client/client_api/queries.rs
```diff
@@ -52,55 +50,41 @@ impl Client {
         let serialised_query = WireMsg::serialize_msg_payload(&msg)?;
         let signature = self.keypair.sign(&serialised_query);
 
-        let starting_query_timeout = self.query_timeout.div_f32(MAX_RETRY_COUNT);
-        trace!(
-            "Setting up query retry, initial interval is: {:?}",
-            starting_query_timeout
-        );
+        let attempt_timeout = self.query_timeout.div_f32(MAX_RETRY_COUNT + 1.0);
```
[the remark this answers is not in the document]
bochaco: it was just to be strict on the "retry" meaning, first attempt wouldn't be a retry

sn/src/client/utils/test_utils/mod.rs
bochaco: because it's only used within that module now that we don't use it from `send_query_with_retry_count`

bochaco: ...and we may not even needed in that other module either, if we are happy with the logic/pattern of this PR

Yoga07 (approved):

PR #872, opened 2021-12-20
fix(blob): respond only once to client for every chunk query

Yoga07:


sn/src/routing/core/msg_handling/service_msgs.rs
```diff
@@ -257,16 +257,20 @@ impl Core {
             correlation_id,
         };
 
-        let origin = if let Some(origin) = self.pending_chunk_queries.get(&user.0).await {
+        let origin = if let Some(origin) = self.pending_chunk_queries.remove(&user.0).await {
             origin
         } else {
             warn!(
-                "Dropping chunk query response from {} for unknown origin: {}",
+                "Dropping chunk query response from {}. We might have already responded to them or \
```
joshuef, line 264: I think we could be a bit clearer in the log.

`from` is the adult that responded or client that requested?

Yoga07: It's a query response so it is from the Adult :)

bochaco (approved):

PR #878, opened 2021-12-22
fix(api): when in dry-run was still requiring a connection by some APIs

bochaco:
This also improves error reported by files container api. Example output when uploading a file is uploaded but verification fails:
```
FilesContainer created at: "safe://hyryyryynzxkzkmpixzurnfyqziabtgpqc9bp6zbsdixmrf1h9paiaz9fc9myeuy?v=hgip1akd6y1eg5cnzxjfqayrb7yyyobuttozu3uwcoggpcnoq1aho"
E  README.md  <Content may have been correctly stored on the network, but verification failed: safe://hy8oyeyybd55gmbsw43438bu1zmn5xk1yqtiicdj6hce7t5rsu41om5ps6ecy> 
```

### DIFF
--- a/sn_api/src/app/files/file_system.rs
+++ b/sn_api/src/app/files/file_system.rs
@@ -11,6 +11,7 @@ use super::{metadata::get_metadata, ProcessedFiles};
 use crate::{app::consts::*, Error, Result, Safe, XorUrl};
 use bytes::Bytes;
 use log::info;
+use safe_network::client::Error as ClientError;
 use std::{collections::BTreeMap, fs, path::Path};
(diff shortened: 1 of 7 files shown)


jacderida:
Just an optional suggestion: in the place where this error is handled in the CLI, you could use `color_eyre` to add a suggestion as to what the user might do in the scenario where the verification has failed.

bochaco:
> Just an optional suggestion: in the place where this error is handled in the CLI, you could use `color_eyre` to add a suggestion as to what the user might do in the scenario where the verification has failed.

sounds good, let me give that a try....

jacderida (approved):

PR #883, opened 2021-12-27
fix: resolve some failing app tests

maqi:

### DIFF
--- a/sn_api/src/app/files/mod.rs
+++ b/sn_api/src/app/files/mod.rs
@@ -2664,6 +2664,10 @@ mod tests {
         // let's add another file but with the same name
         let data = Bytes::from("9876543210");
         let other_file_xorurl = retry_loop!(safe.store_public_bytes(data.clone(), None, false));
+
+        // We give the network a moment to make sure updated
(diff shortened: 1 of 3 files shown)


sn_api/src/app/files/mod.rs
```diff
@@ -2664,6 +2664,10 @@ mod tests {
         // let's add another file but with the same name
         let data = Bytes::from("9876543210");
         let other_file_xorurl = retry_loop!(safe.store_public_bytes(data.clone(), None, false));
+
+        // We give the network a moment to make sure updated
+        std::thread::sleep(std::time::Duration::from_millis(1000));
+
         let (version2, new_processed_files, new_files_map) = retry_loop!(safe.files_container_add(
```
joshuef, line 2671: If this condition needs to return somethig specific, this is where we'd use the `retry_loop_for_pattern`. Which would keep trying until the pattern succeeds. 

I think that's more reliable than just a wait here (and will mean we can kill the above `retry_loop` too). 

I'd suggest looking to use that over an arbitrary duration in these tests

joshuef (changes requested):

PR #884, opened 2021-12-27
feat(api): change files APIs to accept std::Path for path args rather than only &str

bochaco:
- Changed the `files_container_create` API to now create just an empty FilesContainer
- Adding a new `files_container_create_from` API which creates a FilesContainer with files uploaded from a local folder.
- All path args passed to files container APIs can either be &str, std::Path, or std::PathBuf

BREAKING CHANGE: changes to the files public API.

### DIFF
--- a/sn_cli/README-symlinks.md
+++ b/sn_cli/README-symlinks.md
@@ -166,7 +166,7 @@ The JSON for a FileContainer with a single symlink looks like:
 
 A parameter `follow_links` was added to the following public APIs:
 
-* Safe::files_container_create()
+* Safe::files_container_create_from()
 * Safe::files_container_sync()
 * Safe::files_container_add()
 * Safe::files_map_sync()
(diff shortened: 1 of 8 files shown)


sn_cli/README-symlinks.md
```diff
@@ -166,7 +166,7 @@ The JSON for a FileContainer with a single symlink looks like:
 
 A parameter `follow_links` was added to the following public APIs:
 
-* Safe::files_container_create()
+* Safe::files_container_create_from()
```
Yoga07, line 169: Should we add one for `files_container_create()` since that is still a public API to?

bochaco: I don't think I understand, the `files_container_create` doesn't have a `follow_links` arg since it doesn'd upload any local files but just creates an empty container. What do you mean?

Yoga07 (approved):

PR #885, opened 2021-12-28
feat(node): refactor node to support tokio-console and resolve issues flagged by the console

lionel1704:


sn/src/bin/sn_node.rs
```diff
@@ -43,27 +43,31 @@ const BOOTSTRAP_RETRY_TIME: u64 = 3; // in minutes
...
+    //     .wrap_err("Failed to spawn node thread")?;
+
+    // match handle.join() {
+    //     Ok(result) => result,
+    //     Err(error) => {
+    //         // thread panic errors cannot be converted to `eyre::Report` as they are not `Sync`, so
+    //         // the best thing to do is propagate the panic.
+    //         std::panic::resume_unwind(error)
+    //     }
+    // }
```
bochaco, line 70: shall we remove all this commented out code?

sn/src/bin/sn_node.rs
```diff
@@ -43,27 +43,12 @@ const BOOTSTRAP_RETRY_TIME: u64 = 3; // in minutes
...
 
 /// Runs a Safe Network node.
-fn main() -> Result<()> {
+#[tokio::main]
+async fn main() -> Result<()> {
     color_eyre::install()?;
-
-    let handle = std::thread::Builder::new()
-        .name("sn_node".to_string())
-        .stack_size(16 * 1024 * 1024)
```
joshuef, line 51: @lionel1704 do we not need this stack size setting anymore?

lionel1704: When we added this, it was because of `cargo build ....` always failing. The stack size change was suggested by cargo itself. Without this setting the build is passing and the node seems to run okay as well.

PR #887, opened 2021-12-28
feat(node): use sn-consensus (BRB membership algorithm) to agree on members additions/removals

bochaco:
- Making section peers mutations atomic in local network knowledge
- Adapting node relocation logic to new approach with membership consensus
- Removing the JoinsAllowed proposal messaging logic
- Moving join/leave verifications to a common place

BREAKING CHANGE: removal of JoinsAllowed agreement message type

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -303,9 +303,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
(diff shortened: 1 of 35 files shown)


bochaco:
Superseded by PR #1037 and/or #1061 .

PR #892, opened 2021-12-30
Replace Chunk DB with a simple disk chunk store

grumbach:
To address disk sync issues and DB complexities, this PR introduces a simple system for chunk storage. 
Chunks are stored as files on the disk in a tree of directories `0` or `1` that represent the first bits of a chunk's address prefix.
We store data in the dir that matches the first `n` (say 20) bits of its name.

```rust
// chunk bin address (xorname):
1001110000100100101100000110000101000011110000000111001000100100110010001001011110111010110010010111001011100110111010010010101101000110110011110001100000000110001111110001101001000110100111101011111000101111011110100000100101100110001100000110000100000101

// will be stored at path:
"1/0/0/1/1/1/0/0/0/0/1/0/0/1/0/0/1/0/1/1/{{CHUNK_ZBASE32}}"
```

This PR removes the unused kv store (replaced by disk chunk store)

> It includes two currently `unused` helpers: `list_files_without_prefix`, `list_files_with_prefix` that we could use when an adult changes sections and has to sort his Chunks. 
> As the network splits we can take any data in dirs not in our prefix and delete them and the dirs (delete means store them back on the network and delete locally).


sn/src/routing/core/chunk_store/chunk_disk_store.rs
```diff
@@ -0,0 +1,301 @@
...
+    // If the provided bit count is larger than `self.bit_tree_depth`, uses `self.bit_tree_depth`
+    // to stay within the prefix tree path
+    fn prefix_tree_path(&self, xorname: XorName, bit_count: usize) -> PathBuf {
+        let bin = format!("{:b}", xorname);
+        let prefix_dir_list: Vec<String> = bin
+            .chars()
+            .take(std::cmp::min(bit_count, self.bit_tree_depth))
+            .map(|c| format!("{}", c))
+            .collect();
+        let prefix_dir_path: PathBuf = prefix_dir_list.iter().collect();
```
oetyng, line 58: I guess it wasn't possible to collect directly to `PathBuf` without intermediate collection into `prefix_dir_list`? (due to no conversion?)

sn/src/routing/core/chunk_store/chunk_disk_store.rs
```diff
@@ -0,0 +1,301 @@
...
+        std::fs::remove_file(filepath)?;
+        Ok(())
+    }
+
+    pub(crate) fn read_chunk(&self, addr: &ChunkAddress) -> Result<Chunk> {
+        let filepath = self.address_to_filepath(addr)?;
+
+        let mut f = std::fs::File::open(filepath)?;
+        let mut buffer = Vec::new();
+        let _bytes_read = f.read_to_end(&mut buffer)?;
```
oetyng, line 115: Have you considered an async read (as it's I/O)? 
Or were there reasons not to go with it? I can see that calling fn `get_chunk` was not async before, so maybe that's why?

sn/src/routing/core/chunk_store/chunk_disk_store.rs
```diff
@@ -0,0 +1,301 @@
...
+        let prefix_path = self.prefix_tree_path(prefix.name(), prefix.bit_count());
+        let outside_prefix = all_files
+            .into_iter()
+            .filter(|p| !Path::new(&p).starts_with(&prefix_path.as_path()))
+            .collect();
+        Ok(outside_prefix)
+    }
+
+    #[allow(unused)]
+    /// quickly find chunks related or not to a section, might be usefull when adults change sections
```
oetyng, line 154: Minimal typo: `usefull` => `useful`

sn/src/routing/core/chunk_store/chunk_disk_store.rs
```diff
@@ -0,0 +1,301 @@
...
+        let all_files = self.list_all_files()?;
+        let all_addrs = all_files
+            .iter()
+            .map(|filepath| self.filepath_to_address(filepath))
+            .collect();
+        all_addrs
+    }
+
+    #[allow(unused)]
+    /// quickly find chunks related or not to a section, might be usefull when adults change sections
```
oetyng, line 141: Minimal typo: `usefull` => `useful`

oetyng (commented):
Very neat and clean :metal:
Just a couple of Qs and minor typos.

sn/src/routing/core/chunk_store/chunk_disk_store.rs
grumbach: Looks like it is possible! Let me fix that!

sn/src/routing/core/chunk_store/chunk_disk_store.rs
grumbach: Yes. Having it async was the initial idea but since all the functions using it weren't it didn't make sense to turn everything async for a single async call.

oetyng: Hm, I wonder how far up it spreads. It would make sense generally to have a disk-read be I/O I think.

sn/src/dbs/to_db_key.rs
```diff
@@ -0,0 +1,27 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use super::{encoding::serialise, Result};
+use crate::types::{ChunkAddress, Keypair, PublicKey, RegisterAddress};
+use serde::Serialize;
+use xor_name::XorName;
+
+pub(crate) trait ToDbKey: Serialize {
```
dirvine, line 14: nice

sn/src/dbs/used_space.rs
```diff
@@ -91,6 +92,16 @@ impl UsedSpace {
             .map(|path| tokio::spawn(async move { get_size(path).await.map_err(Error::from) }));
         join_all(handles).await.iter().flatten().flatten().sum()
     }
+
+    pub(crate) async fn ratio(&self) -> f64 {
```
dirvine, line 96: This will be slow and should not ve called too much I recon

sn/src/routing/core/chunk_store/errors.rs
```diff
@@ -0,0 +1,52 @@
...
+    InvalidFilename,
+    /// I/O error.
+    #[error("I/O error: {0}")]
+    Io(#[from] io::Error),
+    /// NetworkData error.
+    #[error("Network data error:: {0}")]
+    NetworkData(#[from] crate::types::Error),
+    /// No filename found
+    #[error("Path contains no file name")]
+    NoFilename,
```
dirvine, line 36: not sure why this is different from `ChunkNotFound(XorName)`

dirvine (dismissed):
LGTM :+1:

sn/src/routing/core/chunk_store/errors.rs
grumbach: The `NoFilename` should not happen unless the function `filepath_to_address` was called with a path that contains no filename. 
This should never be happening in the current code, I just didn't want to swallow and hide potential errors (if that function is used somewhere else at some point).

sn/src/dbs/used_space.rs
oetyng: I did some perf tests on that fn some time ago and IIRC it was very fast. Surprisingly fast even.

Maybe @grumbach, if you can look up the time it takes again?

If it is noteworthy, then the fn `can_consume` which will be called every time just before this, also calls `total`, and could be updated to return `(bool, u64)` (where u64 is either total before or total + to be consumed), and then it can be returned from `write_chunk` to be used where it is currently in `chunk_store/mod.rs`, so we don't need to call it again there.
But that is slightly less clean, so I'd only do it if this call is not fast (how fast should it be to not bother with it?).

grumbach: Yes indeed, I moved this logic from the kvstore to there. It's called only when storing chunks to update the `last_recorded_level`. I used the same logic as before there, but calling for every store sounds a bit wasteful.

oetyng: Check out the speed again, it was like 0.000 ms when I ran locally IIRC. Not sure where we want to draw the line though.

sn/src/routing/core/chunk_store/chunk_disk_store.rs
oetyng: I noticed btw that there wasn't much spread, this is like a leaf, and just above it everything is async (by now, very few places would not be, mostly CPU intensive things like e.g. self encryption). 
So, it's all good to go async there.

Yoga07 (approved):

PR #898, opened 2022-01-01
refactor: rename blob to file

oetyng:


sn_api/src/app/files/file_system.rs
```diff
@@ -33,14 +33,14 @@ pub(crate) async fn upload_file_to_net(
 
     let mut mime_type_for_xorurl = mime_guess::from_path(&path).first_raw();
     let result = match safe
-        .store_public_bytes(data.to_owned(), mime_type_for_xorurl, dry_run)
```
joshuef, line 36: Feels like these APIs could stay the same.

what's `bytes` isn't always a `file`. (though the reverse is true).

PR #904, opened 2022-01-03
Storage level throttling

grumbach:
Optimise the chunk `store` by **SIGNIFICANTLY** reducing the amount of calls to the resource heavy used space `total` (that traverses and checks the size of every directory we track)....  to zero.

By simply keeping a counter of the bytes we write and delete!


sn/src/node/routing/core/register_storage.rs
```diff
@@ -124,10 +123,10 @@ impl RegisterStorage {
...
         auth: AuthorityProof<ServiceAuth>,
     ) -> Result<()> {
-        let required_space = std::mem::size_of::<RegisterCmd>() as u64;
-        if !self.used_space.can_consume(required_space).await {
-            return Err(Error::NotEnoughSpace);
-        }
+        // let required_space = std::mem::size_of::<RegisterCmd>() as u64;
+        // if !self.used_space.can_consume(required_space).await {
+        //     return Err(Error::NotEnoughSpace);
+        // }
```
lionel1704, line 129: Do we need this check? I don't see this requirement fulfilled anywhere else.

grumbach: Added the check back as an extra layer of security (otherwise a bad elder could just well store way over the adults max capacity). 
Should not be happening as Adults should stop receiving chunk store requests when over 90%, but let's be safe.

grumbach: The check is super cheap now (comparing integers)

sn/src/dbs/used_space.rs
```diff
@@ -37,7 +37,12 @@ impl UsedSpace {
         let _ = self.used_space.fetch_sub(size, Ordering::Relaxed);
     }
 
-    pub(crate) async fn ratio(&self) -> f64 {
+    pub(crate) fn can_add(&self, size: usize) -> bool {
+        let current_used_space = self.used_space.fetch_add(0, Ordering::Relaxed);
```
oetyng, line 41: Wouldn't `load()` fn be better? That should be a pure read op. This here is a write - even though only adding 0.
Not sure if the internals have some way of optimising away adding zero, but no reason to not go for the read fn?

joshuef (dismissed):

Yoga07 (dismissed):

joshuef (approved):

PR #914, opened 2022-01-05
chore: log EntryHash with XorName

maqi:

### DIFF
--- a/sn/src/types/register/reg_crdt.rs
+++ b/sn/src/types/register/reg_crdt.rs
@@ -11,7 +11,6 @@ use super::super::{
     RegisterAddress as Address, Signature, {utils, Error, PublicKey, Result},
 };
 use super::metadata::Entry;
-pub use crdts::merkle_reg::Hash as EntryHash;
 use crdts::{
     merkle_reg::{MerkleReg, Node},
     CmRDT,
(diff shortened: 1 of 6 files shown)


sn_api/src/app/safe_client.rs
```diff
@@ -231,15 +231,19 @@ impl SafeAppClient {
         address: RegisterAddress,
         hash: EntryHash,
     ) -> Result<Entry> {
-        debug!("Fetching Register hash {:?} at {:?}", hash, address);
+        debug!(
+            "Fetching Register hash {:?} at {:?}",
```
joshuef, line 235:
```suggestion
            "Fetching Register Xorname(hash) {:?} at address: {:?}",
```

joshuef (approved):

sn_api/src/errors.rs
```diff
@@ -68,7 +69,7 @@ pub enum Error {
     #[cfg(feature = "app")]
     /// HashNotFound
     #[error("No entry with hash {0:?}")]
-    HashNotFound(crate::register::EntryHash),
+    HashNotFound(XorName),
```
bochaco, line 72: The hash is not a xorname, why are we wrapping it up in a Xorname?

sn/src/types/register/reg_crdt.rs
```diff
@@ -11,17 +11,35 @@ use super::super::{
...
+use xor_name::XorName;
+
+/// Hash of the register entry. Logging as the same format of XorName.
+#[derive(Clone, Copy, Default, Serialize, Deserialize, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct EntryHash(pub crdts::merkle_reg::Hash);
+
+impl Debug for EntryHash {
+    fn fmt(&self, formatter: &mut Formatter) -> FmtResult {
+        // Using the wrting format of XorName for the logging consistency.
+        write!(formatter, "EntryHash({:?})", XorName(self.0))
```
bochaco, line 33: I think we should have explicitly the code for printing out the bytes here and in `Display` impl, even if it's the same as the `Xorname` impl of `Debug`/`Display`, they are different types, there is no relationship between them, and none of them is a logging utility, so this is misleading an confusing, we can have the following if we like this form:
```
        write!(
            formatter,
            "{:02x}{:02x}{:02x}({:08b})..",
            self[0], self[1], self[2], self
        )
```

joshuef (approved):

Issue #909, opened 2022-01-04
look into optional ACKs. After elders have sent on to adults.


joshuef:
We could use ACKs in a non-required fashion.

Worse case is we fall back onto current behaviour and optionally validate a PUT.

We can send an ACK back once a message has been processed at elders (ie, once the message has gone out to adults to store a chunk, OR once a register has been written).

The ack may then allow a client to stop querying/validating/waiting for AE flows to complete.

(Which in turn may simplify some tests)

maqi:
Couple of points would like to be clarified:
1, Regarding the whole Put flow that could be `Client -> Client's Account Holder (to verify having enough allowance) -> DataHolders' Account Holder (to credit data_holders and check space) -> DataHolders`
Does the Ack only happens on step 1 (Client's Account Holder to Client), or on all steps?
If the Put flow has been changed, which step is expected to be Acked?

2, Does the Ack itself need to contain the whole content of the original message or just a msg_id of it?

3, The Ack itself still need to go through the AE, or it can be an excemption?

joshuef:
I think point 1 is perhaps more 'all encompassing' than needs to be. 

I think there will be back and forth for paying for data. But the PUT of data + payment (what we have as just PUT) should hack an ACK from the receiving Elder, after they have processed the message.

2. Ack could just be for the `OperationId` (not the same as msg id. it's derived from the msg contents)
3. The Ack would not go through any AE as it would just be from Elder->Client.

PR #929, opened 2022-01-10
send CmdAck for client cmd during test

maqi:
Closes https://github.com/maidsafe/safe_network/issues/909

### DIFF
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -178,6 +178,8 @@ impl Session {
     ) -> Result<(), Error> {
         debug!("ServiceMsg with id {:?} received from {:?}", msg_id, src);
         let queries = session.pending_queries.clone();
+        #[cfg(any(test, feature = "test-utils"))]
+        let cmds = session.pending_cmds.clone();
         let error_sender = session.incoming_err_sender;
 
         let _handle = tokio::spawn(async move {
@@ -228,6 +230,27 @@ impl Session {
                         CmdError::Data(_error) => {
                             // do nothing just yet
                         }
+                    }
+                }
+                #[cfg(any(test, feature = "test-utils"))]
+                ServiceMsg::CmdAck { correlation_id } => {
+                    debug!(
+                        "CmdAck was received for Message{:?} w/ID: {:?} from {:?}",
+                        msg_id, correlation_id, src
+                    );
+                    if let Some(sender) = cmds.get(&correlation_id) {
+                        trace!(
+                            "Sending Ack from {:?} for cmd w/{:?} via channel.",
+                            src,
+                            correlation_id
+                        );
+                        let result = sender.send(src).await;
+                        if result.is_err() {
+                            trace!("Error sending cmd Ack on a channel for {:?} cmd_id {:?}: {:?}. (It has likely been removed)", msg_id, correlation_id, result)
+                        }
+                    } else {
+                        // Likely the channel is removed when received majority of Acks
+                        trace!("No channel found for cmd Ack of {:?}", correlation_id);
                     }
                 }
                 msg => {
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -196,7 +196,6 @@ impl Session {
 
     // Handle messages intended for client consumption (re: queries + commands)
     #[instrument(skip(session), level = "debug")]
-    #[allow(clippy::redundant_clone)]
     fn handle_client_msg(
         session: Session,
         msg_id: MessageId,
@@ -205,7 +204,7 @@ impl Session {
     ) -> Result<(), Error> {
         debug!("ServiceMsg with id {:?} received from {:?}", msg_id, src);
         let queries = session.pending_queries.clone();
-        let cmds = session.pending_cmds.clone();
+        let cmds = session.pending_cmds;
 
         let _handle = tokio::spawn(async move {
             match msg {
@@ -258,7 +257,7 @@ impl Session {
                     );
                     Self::send_cmd_response(cmds, correlation_id, src, None);
                 }
-                msg => {
+                _ => {
                     warn!("Ignoring unexpected message type received: {:?}", msg);
                 }
             };
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -178,7 +178,6 @@ impl Session {
     ) -> Result<(), Error> {
         debug!("ServiceMsg with id {:?} received from {:?}", msg_id, src);
         let queries = session.pending_queries.clone();
-        #[cfg(any(test, feature = "test-utils"))]
         let cmds = session.pending_cmds.clone();
         let error_sender = session.incoming_err_sender;
 
@@ -232,7 +231,6 @@ impl Session {
                         }
                     }
                 }
-                #[cfg(any(test, feature = "test-utils"))]
                 ServiceMsg::CmdAck { correlation_id } => {
                     debug!(
                         "CmdAck was received for Message{:?} w/ID: {:?} from {:?}",
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -9,7 +9,10 @@
 use super::Session;
 
 use crate::client::{
-    connections::messaging::{send_message, NUM_OF_ELDERS_SUBSET_FOR_QUERIES},
+    connections::{
+        messaging::{send_message, NUM_OF_ELDERS_SUBSET_FOR_QUERIES},
+        PendingCmdAcks,
+    },
     Error, Result,
 };
 use crate::messaging::{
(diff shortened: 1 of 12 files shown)


sn/src/client/connections/listeners.rs
```diff
@@ -217,6 +219,27 @@ impl Session {
                         }
                     }
                 }
+                #[cfg(any(test, feature = "test-utils"))]
+                ServiceMsg::CmdAck { correlation_id } => {
```
joshuef, line 223: I think we should use an `OperationId` here rather than `msg_id`. It should be unique per operation. (Currently it only exists for GETs, but there's no reason not ot have this for PUTs.

This _may_ be a bigger thing than this PR though.

sn/src/node/routing/core/msg_handling/service_msgs.rs
```diff
@@ -53,6 +53,29 @@ impl Core {
...
+    /// Forms a command to send the provided node Ack out
+    #[cfg(any(test, feature = "test-utils"))]
+    pub(crate) fn send_cmd_ack(&self, target: Peer, msg_id: MessageId) -> Result<Vec<Command>> {
+        let the_error_msg = ServiceMsg::CmdAck {
+            correlation_id: msg_id,
+        };
+
+        let dst = DstLocation::EndUser(EndUser(target.name()));
+
+        // FIXME: define which signature/authority this message should really carry,
```
joshuef, line 65: This should have the node's ed key I think, so we know where it came from.

sn/src/client/connections/listeners.rs
maqi: Yeah, agree with to have it and have a separate PR for the refactory work.

sn/src/client/connections/messaging.rs
```diff
@@ -130,10 +135,37 @@ impl Session {
         )
         .await;
 
-        // lets wait for any potential AE response while we're here.
-        // TODO: be smart about this. Check AE Retry cache for related msg id eg, continue early if we've seen some.
-        // (cannot continue earlier if everything goes okay first time though, which is a shame)
-        tokio::time::sleep(self.standard_wait).await;
```
joshuef, line 136: We want to fall back on this wait... Otherwise the client could be waiting forever for an `Ack` that never comes for one reason or another.

Basically, we want to keep such behaviour, but use ACKs to short-circuit it. 

So here, we'd perhaps change `standard_wait`'s default to be 1s. And if we get an ACK before that, we can continue.

maqi: right, will have a try

sn/src/node/routing/core/msg_handling/mod.rs
```diff
@@ -1019,15 +1019,16 @@ impl Core {
     }
 
     // TODO: Dedupe this w/ node
-    fn random_client_signature(client_msg: &ServiceMsg) -> Result<(MsgKind, Bytes)> {
-        let mut rng = OsRng;
-        let keypair = Keypair::new_ed25519(&mut rng);
+    // Define which signature/authority this message shall carry.
+    // Currently using node's Ed key. May need to use bls key share for concensus purpose.
+    async fn signature_to_client(&self, client_msg: &ServiceMsg) -> Result<(MsgKind, Bytes)> {
```
joshuef, line 1024: I'd say let's rename this as `ed_sign_client_message`, then if we do want BLS later, it's clearer.

i think we can drop these lines too?

```TODO: Dedupe this w/ node
    // Define which signature/authority this message shall carry.```

**Review by joshuef — APPROVED** [review]

File: sn/src/client/connections/messaging.rs [hunk 5]
```diff
@@ -451,12 +451,8 @@ impl Session {
             ..Default::default()
         };
 
-        let next_wait = backoff.next_backoff();
-
         // wait here to give a chance for AE responses to come in and be parsed
-        if let Some(wait) = next_wait {
-            tokio::time::sleep(wait).await;
-        }
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
joshuef, line 457: Why have we gone for a fied value over the backoff here?

We shouldn't need to do a fixed wait, or indeed any now as we wait until we have a properly populated SAP in our network knowledge 🤔

sn/src/node/routing/core/data_records/mod.rs
```diff
@@ -110,6 +110,7 @@ impl Core {
             query,
             auth: auth.into_inner(),
             origin: EndUser(correlation_id),
+            correlation_id: MessageId::from_xor_name(correlation_id),
```
joshuef, line 113: is the correlation id a xorname? I'm not sure this makes sense...

joshuef (changes requested):

sn/src/client/connections/messaging.rs
maqi: first, the previous backoff scheme here gives a sleep around 600ms. We don't have an event update notificator here for the network knowledge here, Hence we always just `sleep` here to give the chance of update. (and `wait till update` in the later on while block)

second, change to use a fixed wait just becuase with the previous scheme, the second sleep on the backoff could be much longer due to the Exponential scheme. So here the the first `fixed` sleep is just to reduce the prolonged second sleep. We could adjust the setup of the backoff scheme to achieve the same, just it needs coupld of rounds of verification on CI runs, so I put a fixed value here for a shortcut.

sn/src/node/routing/core/data_records/mod.rs
maqi: seems a misuse due to the rebase. 
will have a double check

sn/src/client/connections/messaging.rs
```diff
@@ -532,105 +534,116 @@ pub(super) async fn send_message(
...
-            .await;
-
-        if let (true, Ok(connection)) = (reused_connection, connection) {
-            trace!(
-                connection_id = connection.id(),
-                src = %connection.remote_address(),
-                "Client::ConnectionReused",
-            );
-        }
+    let _ = tokio::spawn(async move {
```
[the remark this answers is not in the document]
maqi: The return of the spawned stuff was `Ok(())` perviously, i.e. actually nothing meaningful to be handled.

I notice the blocking of this `send_message` being problematic becuase it only returns when completes all the sending operations (each operation contains part of establish/verify connection to peer). which hinders the listening thread to handle the response from peers. Especially in the case one of the peer having issue with the connection, which will block the whole for 1 min and make other valid responses becomes invalid.

sn/src/node/routing/core/data_records/mod.rs
maqi: https://github.com/maidsafe/safe_network/pull/929/files#diff-9fa5f0c9092474a17ab125854b74eacd5cc3732983736f5daf93eeca32917e15R95
it's the name of the address, which is a XorName.

It shall use a proper OperationId instead though, however as I mentioned in another comment, I'd prefer to have a separate PR to tackle all OperationId related refactory, as it now need to cover both the Put and Get flow, and certain places and tests need to be care for.

sn/src/client/connections/messaging.rs
joshuef: okay makes sense. Let's at least get this wait as a const there 👍

sn/src/node/routing/core/data_records/mod.rs
joshuef: Hmm, that link isn't clear for me at all.

Are you saying that originaly this `correlation_id` is a xorname? If so, ;ets just change the naming of that so it makes sense. (`correlation_id` implies related `msg_id` to me)

sn/src/client/connections/messaging.rs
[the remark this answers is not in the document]
joshuef: >  i.e. actually nothing meaningful to be handled
Aye, but `let _ = `means even if it were changed to be meaningfu we wont see it. So we shoud not use it and instead name the variable and just let it drop (manually if needs).

----- 

w/r/t blocking: okay, good to know!

sn/src/client/connections/mod.rs
```diff
@@ -42,7 +45,10 @@ pub(super) struct Session {
     // Channels for sending responses to upper layers
     pending_queries: PendingQueryResponses,
     // Channels for sending errors to upper layer
+    #[allow(dead_code)]
```
bochaco, line 48: If this is dead code, then perhaps we should just remove it? otherwise fix clippy?

sn/src/messaging/msg_id.rs
```diff
@@ -27,6 +27,11 @@ impl MessageId {
         Self(XorName::random().0)
     }
 
+    ///
```
bochaco, line 30: We are missing the comment here

sn/src/messaging/system/node_msgs.rs
```diff
@@ -81,6 +81,8 @@ pub enum NodeQuery {
         auth: ServiceAuth,
         /// The user that has initiated this query
         origin: EndUser,
+        ///
```
bochaco, line 84: We are missing the comment here

sn/src/client/connections/listeners.rs
```diff
@@ -155,8 +158,32 @@ impl Session {
...
+            }
+        } else {
+            // Likely the channel is removed when received majority of Acks
+            trace!("No channel found for cmd Ack of {:?}", correlation_id);
+        }
+    }
+
     // Handle messages intended for client consumption (re: queries + commands)
     #[instrument(skip(session), level = "debug")]
+    #[allow(clippy::redundant_clone)]
```
bochaco, line 186: Can't we fix this?

sn/src/node/routing/core/data_records/mod.rs
maqi: >Are you saying that originaly this correlation_id is a xorname?

Yes

> let's just change the naming of that so it makes sense.

ok

sn/src/client/connections/mod.rs
maqi: not sure.
I thought it might be used for the notification to the higher layer later on.

joshuef (approved):

maqi:
bors

PR #936, opened 2022-01-11
feat(node): Close client connections after responding

joshuef:


Yoga07 (approved):

joshuef:
I'm gonna close this just now.

Clients reuse connections a lot. Ideally we should be aggresively closing `idle` connections... which comes down to qp2p some... I suspect we need to ensure clients cannot trigger large idle timeouts etc.

PR #940, opened 2022-01-11
chore: removing SafeUrl from this repo and use sn_url crate instead

bochaco:


bochaco:
Closing it as more considerations are needed before moving it out from mono repo.

PR #949, opened 2022-01-14
refactor(sn_api): remove one layer of indirection

oetyng:
Removes one layer of indirection.
Also simplifies instantiation and handling of the safe instance.


oetyng:
Superseded by #955

PR #968, opened 2022-01-23
feat(sn/client): randomize elder contact on cmd.

joshuef:
Priously we always took the first 3 from the SAP's elders vec. This should spread
commands out over all elders more fairly.


dirvine (approved):
LGTM

PR #971, opened 2022-01-24
Fixing ci failures

maqi:

### DIFF
--- a/.github/workflows/bors.yml
+++ b/.github/workflows/bors.yml
@@ -402,8 +402,6 @@ jobs:
         if: failure()
         continue-on-error: true
 
-
-
   e2e-split:
     if: "!startsWith(github.event.pull_request.title, 'Automated version bump')"
     name: E2E tests w/ full network
@@ -560,5 +558,5 @@ jobs:
         if: matrix.os == 'self-hosted-ubuntu' && always()
         run: |
           rm -rf ~/.safe
-
+          killall -9 sn_node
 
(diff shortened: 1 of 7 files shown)


resources/scripts/all_nodes_joined.sh
```diff
@@ -17,7 +17,7 @@ echo "Checking logfiles to check all nodes have joined"
 log_dir="$HOME/.safe/node/local-test-network"
 
 # -u needed here to search log dirs
-nodes_joined=$(rg "ReceivedJoinApproval" "$log_dir" -g "*.log*"  -u -c | wc -l)
```
joshuef, line 20: @maqi this line is using a `LogMarker`, so would be good to keep that in place. If we want to target another one, that's fine and we can add one in the right place for this

joshuef (changes requested):

resources/scripts/all_nodes_joined.sh
maqi: 'ReceivedJoinApproval' is on Trace level, however the E2E full test starting nodes with log level of Info, which doesn't have it logged at all.
Hence I choosed another one that is logged as Info already.
I can change the place of 'ReceivedJoinApproval' to be on Info level, if you think that's better?

joshuef: Yeh, I think so. 

But we should also have the nodes logging trace level I think. I don't think that's been intentionally disabled 🤔

maqi: I thought it's for reduce the logs size uploaded to the github?
There are 33 nodes, log with Info could save over 98% space & upload/download time

joshuef (approved):

joshuef (approved):

PR #972, opened 2022-01-24
feat: retrieve immutable content via nrs

jacderida:
- 2b06d802c **refactor: nrs map fetch and rename multimap**

  The `MultimapKeyValues` type alias was renamed to `Multimap` to make it easier to work with. Some
  functions were also renamed to the same effect, e.g., `fetch_multimap_values` -> `fetch_multimap`.
  
  The function for retrieving the NRS map was also refactored. Some aspects were broken down into
  smaller functions to reduce some complex if/else cases. Things were also setup for the version in
  the last entry in the Multimap to be collected. This is going to be used with the returned
  NrsMapContainer, to facilitate returning links to immutable content, which doesn't have a version.
  Currently the code isn't setup to support that scenario.

- e87774f48 **feat: retrieve immutable content via nrs**

  Changes the resolver to enable retrieval of NRS links to immutable content. Previously, the
  NRS `target_url` was incorrectly being checked to see if it contained a version. We now validate
  this URL using the same process used by NRS, which won't assert that a file link has a version. To
  reduce duplication, the `validate_nrs_url` function in the NRS module was changed to public.

  Tests were added to both the API and CLI to cover the scenario.

  The NrsMap struct was extended to include a subname_version field. This is going to be used to
  request the map with a particular version of a subname. If no version was specified when the
  container was retrieved, then the field won't be set. This is why it's an Option. Since we're
  storing this field on the NrsMap, the `version` field on the `NrsMapContainer` was removed.

  A couple of CLI NRS tests were also re-enabled, one of which happened to be related to immutable
  content.

### DIFF
--- a/sn_api/src/app/nrs/nrs_map.rs
+++ b/sn_api/src/app/nrs/nrs_map.rs
@@ -6,6 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use crate::safeurl::VersionHash;
 use crate::{Error, Result, SafeUrl};
 use log::debug;
 use serde::{Deserialize, Serialize};
@@ -13,7 +14,7 @@ use std::collections::BTreeMap;
 
 pub(crate) type SubName = String;
 
-/// Mapping SubNames to SafeUrls
+/// Mapping Subnames to SafeUrls
 /// For a given Top Name : "example"
 ///
 /// | SubName Key   | Full Name        | SafeUrl Value|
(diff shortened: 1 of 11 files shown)


bochaco (approved):

sn_api/src/app/nrs/nrs_map.rs
```diff
@@ -22,9 +23,13 @@ pub(crate) type SubName = String;
 /// | "sub"         | "sub.example"    | "safe://eg2" |
 /// | "sub.sub"     | "sub.sub.example"| "safe://eg3" |
 ///
+/// The map also contains a version, which is the latest version in the Multimap where this map is
```
grumbach, line 26: What is this version refering to? Subnames are versionned but not the map itself right?

grumbach (approved):
LGTM! (assuming the version in the NRS map is going to be moved inside the map entry keys in a follow up PR)

PR #976, opened 2022-01-25
Active Data Republishing

Yoga07:
Tweaks the liveness tracker to detect nodes that begin to accumulate pending operations. These deviant nodes' data are pre-emptively replicated to avoid data loss and improve fault tolerance in the network.

### DIFF
--- a/sn/src/messaging/system/node_msgs.rs
+++ b/sn/src/messaging/system/node_msgs.rs
@@ -66,6 +66,8 @@ pub enum NodeEvent {
         /// Whether store failed due to full
         full: bool,
     },
+    /// Inform Adults of a possible deviant node
+    DeviantsDetected(BTreeSet<XorName>),
 }
 
 /// Query originating at a node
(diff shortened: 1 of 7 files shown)


sn/src/dbs/errors.rs
```diff
@@ -20,6 +20,9 @@ pub(crate) type Result<T, E = Error> = std::result::Result<T, E>;
 #[non_exhaustive]
 /// Node error variants.
 pub enum Error {
+    /// Temp error
+    #[error("Temporary error")]
+    Temp(String),
```
joshuef, line 25: No vague eerrors like this please. `String` is a catch all that _will_ get abused. 

Can you provide a specific error here

joshuef (changes requested):

joshuef (approved):

sn/src/node/core/liveness_tracking.rs
```diff
@@ -16,8 +16,8 @@ use std::collections::BTreeSet;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
-const NEIGHBOUR_COUNT: usize = 2;
-const MIN_PENDING_OPS: usize = 10;
+const NEIGHBOUR_COUNT: usize = 3;
```
joshuef, line 19: @Yoga07 it would be great to get a bit of documention for all these constants and how they affect the algo.

joshuef (changes requested):

joshuef (approved):

PR #985, opened 2022-01-31
chore: simplify data replication cmds and flow

oetyng:
We had `StoreData`, `ReplicateData` and `RepublishData`, where it was a bit confusing what did what.
It is now simplified to `ReplicateData` cmd and `CouldNotStoreData` event, where one is
used to send data to Adults, and the other is sent by Adults to Elders
when they can't handle the data.

### DIFF
--- a/sn/src/messaging/system/node_msgs.rs
+++ b/sn/src/messaging/system/node_msgs.rs
@@ -7,7 +7,9 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::messaging::{
-    data::{DataCmd, DataExchange, DataQuery, OperationId, QueryResponse, Result, StorageLevel},
+    data::{
+        DataCmd, DataQuery, MetadataExchange, OperationId, QueryResponse, Result, StorageLevel,
+    },
     EndUser, MessageId, ServiceAuth,
 };
 use crate::types::{
@@ -32,13 +34,6 @@ pub enum NodeCmd {
         /// Message source
         origin: EndUser,
     },
-    /// Data is stored by Adults
-    StoreData {
-        /// The data
-        data: ReplicatedData,
-        /// Message source
-        origin: EndUser,
-    },
     /// Notify Elders on nearing max capacity
     RecordStorageLevel {
         /// Node Id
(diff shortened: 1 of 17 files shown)


sn/src/messaging/system/node_msgs.rs
```diff
@@ -48,15 +43,15 @@ pub enum NodeCmd {
         /// The storage level reported by the node.
         level: StorageLevel,
     },
-    /// Replicate a given chunk at an Adult (sent from elders on receipt of RepublishData)
+    /// Replicate a given chunk at an Adult
```
joshuef, line 46: this could be chunk or reg, no?

sn/src/messaging/system/node_msgs.rs
```diff
@@ -48,15 +43,15 @@ pub enum NodeCmd {
...
     ReplicateData(ReplicatedData),
-    /// Tells the Elders to re-publish a chunk in the data section
-    RepublishData(ReplicatedData),
+    /// Sent by a full Adult, and tells the Elders to store a chunk at some other Adult in the section
+    BounceData(ReplicatedData),
```
joshuef, line 49: I'm unsure of this naming.

`Bounce` as in "bounce it off the elder to another adult?" 

Maybe something like `MarkForReplication` if this is going to cause `ReplicateData` to happen?

sn/src/messaging/system/node_msgs.rs
oetyng: vyep

sn/src/messaging/system/node_msgs.rs
oetyng: It Bounces from the Adult, because it can't handle it.

oetyng: The Adult could Replicated it immediately though..

oetyng: So, can be cleaned up by extending the semantics of a bounce msg to include that the Elders should mark the Adult as full.

oetyng: The event implemented instead now.

joshuef (approved):

PR #991, opened 2022-02-01
Cmd retries

joshuef:

### DIFF
--- a/sn/src/client/connections/messaging.rs
+++ b/sn/src/client/connections/messaging.rs
@@ -144,6 +144,11 @@ impl Session {
         )
         .await;
 
+        if res.is_err() {
+            // shortcircuit the ack awaiting
+            return res;
+        }
+
         let expected_acks = std::cmp::max(1, targets_count * 2 / 3);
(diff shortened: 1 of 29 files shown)


sn/src/client/connections/messaging.rs
```diff
@@ -571,6 +572,27 @@ impl Session {
     }
 }
 
+/// Logs a conneciton id as failed for a session
+#[instrument(skip_all, level = "trace")]
+pub(super) fn mark_connection_id_as_failed(
```
Yoga07, line 577: Is it necessary though? Because we record the last failed connection with the node. In case the new connection fails again we update the last failure and open a new connection. Asking because it increases the complexity a tad bit whereas the first approach seems fine to me.

joshuef: >  In case the new connection fails again we update the last failure and open a new connection.

This is where we do that. Before we had the same code in 3 or four places, now we just call this to do the same.

This then lets us open a new connection if it's been logged here.

Yoga07 (approved):

PR #993, opened 2022-02-03
Store Full Public Name in NRS Map and Remove API URL Sanitise

jacderida:
- 1115093a4 **feat: store full public name in nrs map**

  Entries will now use the full public name reference, rather than the just subname. Like so:
  * example -> link
  * a.example -> link
  * b.example -> link
  * a.b.example -> link

  This is to facilitate resolving NrsMapContainer content using its XorUrl, which effectively prints
  the NRS map. It was a feature we had that broke at some point. Right now, the XorUrl of the
  container will resolve the content the topname is linked to, or will result in an error if the
  there's no topname link.

- effc6fa5a **tests: rework nrs tests and provide more coverage**

  Added more coverage for code paths that weren't being tested:
  * `nrs_create` duplicate topname
  * `nrs_associate` with non-versioned files container link
  * `nrs_associate` with non-versioned NRS map container link
  * `nrs_associate` with register link
  * `nrs_associate` with invalid URL

  Some NRS tests were reworked as follows:
  * Make them single purpose. Some tests were testing a few things, which makes it a bit harder to
    identify what's failed.
  * Trim down. Some tests had assertions and setup that didn't seem quite relevant, like checking and
    assigning versions.
  * Test `nrs_associate` without calling `nrs_add`. This muddies the water a little bit because
    `nrs_add` calls associate.
  * Separate `nrs_associate` tests for topnames and subnames.

- 29fd62617 **refactor: remove url sanitisation from api**

  The `Safe::parse_url` function was removed from the API. This function 'sanitised' a URL by applying
  a `safe://` prefix to a URL string if the caller hadn't specified it.

  Initially, it was done to tidy up NRS code that was calling this function, but the same code was
  also calling a private function `parse_url` was making use of, so effectively the code was being
  called twice. More generally, we decided callers of the API should be responsible for passing a
  valid URL.

  The function was being called by various other parts of the API and also in the CLI, so these were
  changed to call `SafeUrl::from_url` directly.

  Some code was added to CLI commands to apply the `safe://` prefix if the user omitted it, so no
  functionality  was broken. A few test cases were added to cover it. A couple of NRS test cases for
  validating URLs were also removed as they no longer apply. This behaviour may actually have been
  incorrect in the first place.
  
  Also apply various clippy fixes.

- 58bf67879 **refactor: subname -> public name**

  Since we're now storing the full public name in the NRS map, the key should be referred to as the
  public name rather than the subname. Error messages were also updated.

- 9af70e778 **refactor: remove get_target_url function**

  This function contained an unnecessary check to see if the `safe://` prefix wasn't present in top
  name or public names, which we wouldn't expect it to be. I'm not sure what I was thinking with this.

  Also made some more readability changes based on PR feedback.

### DIFF
--- a/sn_api/src/app/nrs/mod.rs
+++ b/sn_api/src/app/nrs/mod.rs
@@ -34,13 +34,13 @@ pub fn validate_nrs_url(link: &SafeUrl) -> Result<()> {
             || content_type == ContentType::NrsMapContainer
         {
             return Err(Error::UnversionedContentError(format!(
-                "The linked content ({}) is versionable, therefore NRS requires the link to specify a hash: {}",
-                content_type, link
+                "{} content is versionable. NRS requires the supplied link to specify a version hash.",
+                content_type
             )));
         } else if data_type == DataType::Register {
             return Err(Error::UnversionedContentError(format!(
-                "The linked content ({}) is versionable, therefore NRS requires the link to specify a hash: {}",
-                data_type, link
+                "{} content is versionable. NRS requires the supplied link to specify a version hash.",
+                data_type
             )));
         }
     }
@@ -67,7 +67,6 @@ impl Safe {
     pub async fn nrs_create(&self, top_name: &str) -> Result<SafeUrl> {
         info!("Creating an NRS map for: {}", top_name);
 
-        // Check top_name, check if there is an NrsMapContainer there already
         let url_str = validate_nrs_top_name(top_name)?;
         let nrs_xorname = SafeUrl::from_nrsurl(&url_str)?.xorname();
         debug!("XorName for \"{:?}\" is \"{:?}\"", &url_str, &nrs_xorname);
@@ -75,7 +74,6 @@ impl Safe {
             return Err(Error::NrsNameAlreadyExists(top_name.to_owned()));
         }
 
-        // Create a new empty Multimap
         let _ = self
             .multimap_create(Some(nrs_xorname), NRS_MAP_TYPE_TAG, false)
             .await?;
@@ -137,7 +135,6 @@ impl Safe {
             public_name, link
         );
 
-        // make sure the topname is registered
         let sanitised_url = sanitised_url(public_name);
         let safe_url = Safe::parse_url(&sanitised_url)?;
         let top_name = safe_url.top_name();
@@ -148,7 +145,6 @@ impl Safe {
             Err(e) => Err(e),
         }?;
 
-        // associate with peace of mind
         let new_url = self.nrs_associate(public_name, link).await?;
         Ok((new_url, did_register_topname))
     }
@@ -171,10 +167,8 @@ impl Safe {
             public_name
         );
 
-        // check public_name
         let url_str = validate_nrs_public_name(public_name)?;
 
-        // get current latest for subname
         let safe_url = Safe::parse_url(public_name)?;
         let current_versions = self
             .fetch_multimap_values_by_key(&safe_url, public_name.as_bytes())
@@ -183,9 +177,7 @@ impl Safe {
             .map(|(hash, _)| hash)
             .collect();
 
-        // remove
         let entry_hash = self.multimap_remove(&url_str, current_versions).await?;
-
         Ok(get_versioned_nrs_url(url_str, entry_hash)?)
     }
 
@@ -361,11 +353,10 @@ mod tests {
     use super::*;
     use crate::{
         app::test_helpers::{new_safe_instance, random_nrs_name},
-        retry_loop, retry_loop_for_pattern, Error, SafeUrl,
+        Error, SafeUrl,
     };
-    use anyhow::{anyhow, bail, Result};
-    use rand::Rng;
-    use std::{matches, str::FromStr};
+    use anyhow::{anyhow, Result};
+    use std::matches;
 
     const TEST_DATA_FILE: &str = "./testdata/test.md";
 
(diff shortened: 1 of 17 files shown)


sn_api/src/app/nrs/nrs_map.rs
```diff
@@ -12,44 +12,44 @@ use log::debug;
...
+/// | "sub.example"     | "sub.example"    | "safe://sub.example"     |
+/// | "sub.sub.example" | "sub.sub.example"| "safe://sub.sub.example" |
+///
+/// The map also has a subname version field that optionally specifies a subname at a particular
+/// version, since it's possible to have multiple entries for a given subname. If no version was
+/// requested when the map is retrieved, it will be set to `None`.
 #[derive(Debug, PartialEq, Default, Serialize, Deserialize, Clone)]
 pub struct NrsMap {
-    pub map: BTreeMap<SubName, SafeUrl>,
+    pub map: BTreeMap<Subname, SafeUrl>,
```
grumbach, line 36: The `Subname` type now refers to a full public name, we should probably rename it.

sn_api/src/app/nrs/nrs_map.rs
```diff
@@ -12,44 +12,44 @@ use log::debug;
...
-        match self.map.get(sub_name) {
+        match self.map.get(public_name) {
             Some(link) => {
-                debug!("NRS: Subname resolution is: {} => {}", sub_name, link);
+                debug!("NRS: Subname resolution is: {} => {}", public_name, link);
                 Ok(link.to_owned())
             }
             None => {
-                debug!("NRS: No link found for subname(s): {}", sub_name);
+                debug!("NRS: No link found for subname(s): {}", public_name);
```
grumbach, line 49: And maybe adapt the error messages as well

sn_api/src/app/nrs/mod.rs
```diff
@@ -324,53 +304,52 @@ fn convert_multimap_to_nrs_map(multimap: &Multimap) -> Result<NrsMap> {
...
     }
-    Ok(sanitised_url)
+    Ok(url)
 }
 
-// Makes sure there’s a (and only one) "safe://" in front of input name
-fn sanitised_url(name: &str) -> String {
-    format!("safe://{}", name.strip_prefix("safe://").unwrap_or(name))
+fn get_url_from_name(name: &str) -> Result<SafeUrl> {
+    if !name.starts_with("safe://") {
```
grumbach, line 338: Is this check necessairy? A name is not supposed to contain safe:// in the first place rigth?

sn_api/src/app/resolver/mod.rs
```diff
@@ -132,8 +132,8 @@ impl Safe {
     ///
     /// # });
     /// ```
-    pub async fn inspect(&self, url: &str) -> Result<Vec<SafeData>> {
-        let safe_url = Safe::parse_url(url)?;
+    pub async fn inspect(&mut self, url: &str) -> Result<Vec<SafeData>> {
```
grumbach, line 135: Why is inspect requiring a `mut` se￼￼lf here?

grumbach (commented):
Looking good! A couple of comments, but generally a great improvement!

sn_api/src/app/resolver/mod.rs
jacderida: This was a mistake. It was something that was changed in this PR:
https://github.com/maidsafe/safe_network/pull/987/files

I got a conflict with that and I didn't merge it correctly. It should have been removed with my clippy fixes.

sn_api/src/app/nrs/mod.rs
jacderida: That's done because `SafeUrl::from_url` doesn't work unless the name is prefixed with `safe://`. The calling code uses the URL that's returned.

sn_api/src/app/nrs/nrs_map.rs
jacderida: Yeah, this is a good suggestion, thanks.

sn_cli/src/subcommands/cat.rs
```diff
@@ -30,17 +30,18 @@ pub struct CatCommands {
 }
 
 pub async fn cat_commander(cmd: CatCommands, output_fmt: OutputFmt, safe: &Safe) -> Result<()> {
-    let url = get_from_arg_or_stdin(cmd.location, None)?;
-    debug!("Running cat for: {:?}", &url);
+    let link = get_from_arg_or_stdin(cmd.location, None)?;
+    let url = get_target_url(&link)?;
+    debug!("Running cat for: {:?}", &url.to_string());
```
bochaco, line 35: You shouldn't need to call `to_string`, and just use the `{}` for printing it out instead of `{:?}` which is the debug impl.

sn_api/src/app/nrs/mod.rs
grumbach: I'm wondering if the `get_url_from_name(name: &str)` `name` here could be confusing, I'm kind of expecting an actual name here as argument (possibly even a `PublicName` type) so without the "safe://", but if you're saying people can call this with a `safe://` scheme+name, then maybe `get_url_from_name` is not the right name? (sorry about all those names)

sn_cli/src/subcommands/nrs.rs
```diff
@@ -80,7 +80,7 @@ async fn run_register_subcommand(
     output_fmt: OutputFmt,
 ) -> Result<()> {
     if let Some(ref link) = link {
-        validate_target_link(link)?;
+        get_target_url(link)?;
```
bochaco, line 83: Isn't clippy complaining about not using the returned value?

sn_api/src/app/nrs/mod.rs
jacderida: Removed the use of this function. Not quite sure what I was thinking here, to be honest.

sn_cli/src/subcommands/nrs.rs
jacderida: This was left in as a mistake. The function gets called later in the code and the return value is used.

grumbach (approved):
LGTM

happybeing:
It's there an explanation somewhere if how this differs from the earlier scheme?

It sounds very different, so I'm curious about what is different and the effects it has on applications.

jacderida:
> It's there an explanation somewhere if how this differs from the earlier scheme?
> 
> It sounds very different, so I'm curious about what is different and the effects it has on applications.

I'm not sure exactly what you mean by "scheme" in this context, but this just changes how the NRS entries are stored internally; there are no external changes.

happybeing:
I mainly understand the change adequately! Previously the public name was used to look up a data structure which contained information for each subname about the 'service' it implemented (e.g. www, or perhaps application defined services), and a pointer (xor address) to the data needed to deliver the service.

On the face of it the service aspect has been dropped, is that the case?

grumbach:
Hey @happybeing thanks for that question!

> On the face of it the service aspect has been dropped, is that the case?

No. But we changed the way we access and store data on the NRS map (API side). 

> the public name was used to look up a data structure which contained information for each subname, about the 'service' it implemented

This is still the case, the change in that PR is how we represent that data internally. Previously if one would query "docs.google":

1) we would look for the "google" NRS map container (to simplify, we hash "google" and get a xor address from that)
2) look for the subname "docs" in it
3) and return the SafeUrl associated to it (SafeUrls are typed and associated to a kind of service)

What this PR does is change the way we store data in the NRS map container (step 2) so the steps are the same except for step 2:

2) look for the whole public name (subname+topname) "docs.google" in it

The reason why we decided to store the entire public name in there is because people might query the NrsMap directly using its xor address on the network without knowing it is "google"'s NRS map. What they would get is:
```
"maps" -> Url1
"docs" -> Url2
...
```
They'd have no idea what those subnames refer to, you can't unhash the xor address back to "google" to know it's subnames for google. 
Now, in this PR, since we store the entire public name they will get:
```
"maps.google" -> Url1
"docs.google" -> Url2
...
```

Hope that helped make things clearer?

happybeing:
If I've got it, you've potentially reduced the lookups to one in cases (such as www) where the only metadata needed is the type, as that's now part of the URL obtained from the map. Regardless, sounds good. Thank you.

PR #997, opened 2022-02-07
feat(testnet): add `flame` arg to use cargo flamegraph

joshuef:
uses the launcher option to enable flamegraph generation per node

### DIFF
--- a/sn/Cargo.toml
+++ b/sn/Cargo.toml
@@ -10,9 +10,6 @@ readme = "README.md"
 repository = "https://github.com/maidsafe/safe_network"
 version = "0.55.4"
 
-#[profile.release]
-#incremental = true
-
 [[bench]]
 name = "upload_bytes"
 harness = false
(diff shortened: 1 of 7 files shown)


sn/src/types/utils.rs
```diff
@@ -96,6 +96,8 @@ pub(crate) async fn compare_and_write_prefix_map_to_disk(
         .await
         .map_err(|e| Error::FileHandling(e.to_string()))?;
 
+    debug!("END of prefixmapwrite");
```
Yoga07, line 99: Can be removed?

sn/Cargo.toml
```diff
@@ -80,7 +76,7 @@ serde_bytes = "0.11.5"
 serde_json = "1.0.53"
 signature = "1.1.10"
 sled = "0.34.6"
-sn_launch_tool = "0.9.4"
+sn_launch_tool = {path="../../sn_launch_tool"}
```
Yoga07, line 79: Point to crates.io?

Yoga07 (approved):

PR #1002, opened 2022-02-08
refactor(node): reorganise internal node messaging mod and remove redundant private APIs

bochaco:
- Reorganise internal node messaging mod and remove redundant private APIs
- Fixing missing AE update to be sent to sibling new Elders upon split.
- Removing some node's public API which shouldn't be needed for any user of it (e.g. a UI).
- Fix xorurl cmds tests and enable them to run in CI

BREAKING CHANGE: some APIs were removed from node's public interface.


joshuef (approved):

joshuef:
failed due to bad section startup there. reborsing

bors r+

PR #1004, opened 2022-02-09
fix: derive shell in a cross-platform way

b-zee:
Fixes building on Nix(OS).


joshuef (approved):

b-zee:
Come on bors, you can do it!

PR #1014, opened 2022-02-14
Connection tweaks

joshuef:

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -32,7 +32,7 @@ jobs:
           cache-on-failure: true
 
       - name: Build launch local network
-        run: cargo build --release --features=always-joinable
+        run: cargo build --release
 
       - name: Run testnet
         run: ./target/release/testnet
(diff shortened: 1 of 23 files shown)


sn/src/types/peer.rs
```diff
@@ -221,10 +221,30 @@ impl Peer {
             }
         }
 
-        // If we couldn't get the read lock synchronously, we conservatively take the write lock.
+        // If we couldn't get the read lock synchronously, that's probably becaue it was being written...
```
maqi, line 224: is `being written` being the only reason of failed to `try_read` ?
if not, then a forced `read` may cause a deadlock?

sn/src/types/peer.rs
```diff
@@ -221,10 +221,30 @@ impl Peer {
...
+                // because we just tested that a connection is set.
+                return Ok(RwLockReadGuard::try_map(guard, Option::as_ref)
+                    .expect("write-locked value can't have changed"));
+            }
+        }
+
+        // We _still_ dont have the connection. So it's time to connect and write.
+        // first we'll try and make a new connection. We can always throw this away if another already exist when
+        // we take the write lock
+        let new_connection = connect(self.addr).await?;
```
maqi, line 242: will this mean there could be multiple `new_connection` to be created to the same peer, as it is only locked by a `read` above ?

PR #1021, opened 2022-02-18
fix(liveness): add MIN_PENDING_OPS threshold for liveness checks

Yoga07:


sn/src/node/core/data/records/liveness_tracking.rs
```diff
@@ -221,29 +224,30 @@ impl Liveness {
...
-                tracing::info!(
-                    "Pending ops for {}: {} Neighbour max: {}",
-                    node,
-                    pending_operations_count,
-                    max_pending_by_neighbours
-                );
-                unresponsive_nodes.push((node, pending_operations_count));
-            }
+            // Checks begin only if op count crosses MIN_PENDING_OPS
+            if pending_operations_count > MIN_PENDING_OPS {
```
joshuef, line 229: does this need to be a nested `if` statement? Could it not just be `&&` (if its getting long we can assign these variables above which might also increase clarity too? `over_threshold`, `excessive_compared_to_neighbours` eg)

joshuef (approved):

PR #1023, opened 2022-02-19
feat(chunks): batch store

oetyng:
Sled batching write speed far outperforms both sled singles and the
hierarchical dir store.
Additionally, read access times are improved for recently stored chunks.


oetyng:
Superseded by #1064

PR #1026, opened 2022-02-20
Fix connection deadlock

oetyng:

### DIFF
--- a/sn/src/client/client_api/mod.rs
+++ b/sn/src/client/client_api/mod.rs
@@ -27,9 +27,7 @@ use crate::types::{
 use bytes::Bytes;
 use itertools::Itertools;
 use rand::rngs::OsRng;
-use std::collections::BTreeSet;
-use std::net::SocketAddr;
-use std::sync::Arc;
+use std::{collections::BTreeSet, net::SocketAddr, sync::Arc};
 use tokio::{
     sync::{mpsc::Receiver, RwLock},
     time::Duration,
@@ -160,8 +158,7 @@ impl Client {
             config.local_addr,
             config.cmd_ack_wait,
             prefix_map.clone(),
-        )
-        .await?;
+        )?;
 
         let client = Self {
             keypair,
@@ -207,7 +204,8 @@ impl Client {
             if let Some(sap) = prefix_map.closest_or_opposite(&XorName::random(), None) {
                 sap.elders_vec()
             } else {
-                // these peers will be nonsense peers, and dropped after we connect. Reaplced by whatever SectionAuthorityProvider peers we have received
+                // these peers will be nonsense peers, and dropped after we connect. Replaced by whatever SectionAuthorityProvider peers we have received
+                // therefore we use a random name for them initially
                 bootstrap_nodes
                     .iter()
                     .copied()
--- a/sn/src/client/connections/listeners.rs
+++ b/sn/src/client/connections/listeners.rs
@@ -11,7 +11,7 @@ use super::Session;
 use crate::at_least_one_correct_elder;
 use crate::client::{
     connections::{
-        messaging::{mark_connection_id_as_failed, send_msg, NUM_OF_ELDERS_SUBSET_FOR_QUERIES},
+        messaging::{send_msg, NUM_OF_ELDERS_SUBSET_FOR_QUERIES},
         PendingCmdAcks,
     },
     Error, Result,
@@ -26,7 +26,7 @@ use crate::types::{log_markers::LogMarker, utils::compare_and_write_prefix_map_t
 
 use bytes::Bytes;
 use itertools::Itertools;
-use qp2p::{Close, ConnectionError, ConnectionIncoming, SendError};
+use qp2p::{Close, ConnectionError, ConnectionIncoming as IncomingMsgs, SendError};
 use rand::rngs::OsRng;
 use rand::seq::SliceRandom;
 use secured_linked_list::SecuredLinkedList;
@@ -38,30 +38,32 @@ impl Session {
     #[instrument(skip_all, level = "debug")]
     pub(crate) fn spawn_msg_listener_thread(
         session: Session,
-        connection_id: usize,
-        connected_peer: Peer,
-        mut incoming_msgs: ConnectionIncoming,
+        peer: Peer,
+        conn: qp2p::Connection,
+        mut incoming_msgs: IncomingMsgs,
     ) {
-        let src = connected_peer.addr();
-        debug!("Listening for incoming msgs from {}", connected_peer);
+        let mut first = true;
+        let addr = peer.addr();
+        let node_id = peer.id();
+        let connection_id = conn.id();
 
-        trace!(
-            "{} to {} (id: {})",
-            LogMarker::ConnectionOpened,
-            src,
-            connection_id
-        );
+        debug!("Listening for incoming msgs from {:?}", peer);
 
         let _handle = tokio::spawn(async move {
             loop {
-                match Self::listen_for_incoming_msg(src, &mut incoming_msgs).await {
+                match Self::listen_for_incoming_msg(addr, &mut incoming_msgs).await {
                     Ok(Some(msg)) => {
-                        if let Err(err) = Self::handle_msg(msg, connected_peer.clone(), session.clone()).await {
+                        if first {
+                            first = false;
+                            session.peer_links.add_incoming(&node_id, conn.clone()).await;
+                        }
+
+                        if let Err(err) = Self::handle_msg(msg, peer.clone(), session.clone()).await {
                             error!("Error while handling incoming msg: {:?}. Listening for next msg...", err);
                         }
                     },
                     Ok(None) => {
-                        info!("Incoming msg listener has closed for connection {}.", connection_id);
+                        // once the msg loop breaks, we know this specific connection is closed
                         break;
                     }
                     Err( Error::QuicP2pSend(SendError::ConnectionLost(
@@ -71,15 +73,10 @@ impl Session {
                             "Connection was closed by the node: {:?}",
                             String::from_utf8(reason.to_vec())
                         );
-
-                        mark_connection_id_as_failed(session.clone(), connected_peer.name(), connection_id);
-
                     },
                     Err(Error::QuicP2p(qp2p_err)) => {
                           // TODO: Can we recover here?
                           info!("Error from Qp2p received, closing listener loop. {:?}", qp2p_err);
-
-
                           break;
                     },
                     Err(error) => {
@@ -89,19 +86,18 @@ impl Session {
             }
 
             // once the msg loop breaks, we know the connection is closed
-            trace!("{} to {} (id: {})", LogMarker::ConnectionClosed, src, connection_id);
-        }.instrument(info_span!("Listening for incoming msgs from {}", ?src))).in_current_span();
+            trace!("{} to {} (id: {})", LogMarker::ConnectionClosed, addr, connection_id);
+        }.instrument(info_span!("Listening for incoming msgs from {}", ?addr))).in_current_span();
     }
 
     #[instrument(skip_all, level = "debug")]
     pub(crate) async fn listen_for_incoming_msg(
         src: SocketAddr,
-        incoming_msgs: &mut ConnectionIncoming,
+        incoming_msgs: &mut IncomingMsgs,
     ) -> Result<Option<MsgType>, Error> {
         if let Some(msg) = incoming_msgs.next().await? {
             trace!("Incoming msg from {:?}", src);
             let msg_type = WireMsg::deserialize(msg)?;
-
             Ok(Some(msg_type))
         } else {
             Ok(None)
@@ -226,10 +222,9 @@ impl Session {
                         if let Some(entry) = queries.get(&op_id) {
                             let all_senders = entry.value();
                             for (_msg_id, sender) in all_senders {
-                                trace!("Sending response for query w/{:?} via channel.", op_id);
-                                let result = sender.try_send(response.clone());
-                                if result.is_err() {
-                                    trace!("Error sending query response on a channel for {:?} op_id {:?}: {:?}. (It has likely been removed)", msg_id, op_id, result)
+                                let res = sender.try_send(response.clone());
+                                if res.is_err() {
+                                    trace!("Error relaying query response internally on a channel for {:?} op_id {:?}: {:?}. (It has likely been removed)", msg_id, op_id, res)
                                 }
                             }
                         } else {
@@ -304,7 +299,7 @@ impl Session {
         if let Some((msg_id, elders, service_msg, dst_location, auth)) =
             Self::new_target_elders(bounced_msg.clone(), &target_sap).await?
         {
-            let ae_msg_src_name = src_peer.clone().name();
+            let ae_msg_src_name = src_peer.name();
             // here we send this to only one elder for each AE message we get in. We _should_ have one per elder we sent to.
             // deterministically send to most elder based upon sender
             let target_elder = elders
@@ -326,8 +321,7 @@ impl Session {
 
                 debug!("Resending original message on AE-Redirect with updated details. Expecting an AE-Retry next");
 
-                let endpoint = session.endpoint.clone();
-                send_msg(session, vec![elder], wire_msg, endpoint, msg_id).await?;
+                send_msg(session, vec![elder], wire_msg, msg_id).await?;
             } else {
                 error!("No elder determined for resending AE message");
             }
--- a/sn/src/client/connections/messaging.rs
+++ b/sn/src/client/connections/messaging.rs
@@ -8,24 +8,21 @@
 
 use super::{QueryResult, Session};
 
-use crate::client::{connections::CmdResponse, Error};
-use crate::elder_count;
+use crate::client::{connections::CmdResponse, Error, Result};
 use crate::messaging::{
     data::{CmdError, DataQuery, QueryResponse},
     DstLocation, MsgId, MsgKind, ServiceAuth, WireMsg,
 };
-use crate::types::{prefix_map::NetworkPrefixMap, Peer, PublicKey};
+use crate::types::{prefix_map::NetworkPrefixMap, Peer, PeerLinks, PublicKey, SendToOneError};
+use crate::{at_least_one_correct_elder, elder_count};
 
 use backoff::{backoff::Backoff, ExponentialBackoff};
 use bytes::Bytes;
 use dashmap::DashMap;
 use futures::future::join_all;
 use qp2p::{Close, Config as QuicP2pConfig, ConnectionError, Endpoint, SendError};
-use rand::rngs::OsRng;
-use rand::seq::SliceRandom;
-use std::path::PathBuf;
-use std::time::Duration;
-use std::{net::SocketAddr, sync::Arc};
+use rand::{rngs::OsRng, seq::SliceRandom};
+use std::{net::SocketAddr, path::PathBuf, sync::Arc, time::Duration};
 use tokio::{
     sync::mpsc::{channel, Sender},
     sync::RwLock,
@@ -46,7 +43,7 @@ const INITIAL_WAIT: u64 = 1;
 impl Session {
     /// Acquire a session by bootstrapping to a section, maintaining connections to several nodes.
     #[instrument(skip(err_sender), level = "debug")]
-    pub(crate) async fn new(
+    pub(crate) fn new(
         client_pk: PublicKey,
         genesis_key: bls::PublicKey,
         qp2p_config: QuicP2pConfig,
@@ -54,10 +51,9 @@ impl Session {
         local_addr: SocketAddr,
         cmd_ack_wait: Duration,
         prefix_map: NetworkPrefixMap,
-    ) -> Result<Session, Error> {
-        trace!("Trying to bootstrap to the network");
-
+    ) -> Result<Session> {
         let endpoint = Endpoint::new_client(local_addr, qp2p_config)?;
+        let peer_links = PeerLinks::new(endpoint.clone());
 
         let session = Session {
             pending_queries: Arc::new(DashMap::default()),
@@ -68,7 +64,7 @@ impl Session {
             genesis_key,
             initial_connection_check_msg_id: Arc::new(RwLock::new(None)),
             cmd_ack_wait,
-            elder_last_closed_connections: Arc::new(DashMap::default()),
+            peer_links,
         };
 
         Ok(session)
@@ -80,38 +76,13 @@ impl Session {
         dst_address: XorName,
         auth: ServiceAuth,
         payload: Bytes,
-        targets_count: usize,
-    ) -> Result<(), Error> {
+    ) -> Result<()> {
         let endpoint = self.endpoint.clone();
         // TODO: Consider other approach: Keep a session per section!
 
-        // Get DataSection elders details.
-        let (mut elders, section_pk) =
-            if let Some(sap) = self.network.closest_or_opposite(&dst_address, None) {
-                let sap_elders = sap.elders_vec();
-
-                trace!("SAP elders found {:?}", sap_elders);
-
-                (sap_elders, sap.section_key())
-            } else {
-                return Err(Error::NoNetworkKnowledge);
-            };
+        let (section_pk, elders) = self.get_cmd_elders(dst_address).await?;
 
         let msg_id = MsgId::new();
-
-        // any SAP that does not hold elders_count() is indicative of a broken network (after genesis)
-        if elders.len() < elder_count() {
-            error!("Insufficient knowledge to send to {:?}", dst_address);
-            return Err(Error::InsufficientElderKnowledge {
-                connections: elders.len(),
-                required: targets_count,
-                section_pk,
-            });
-        }
-
-        elders.shuffle(&mut OsRng);
-        // now we use only the required count
-        let elders: Vec<_> = elders.into_iter().take(targets_count).collect();
 
         debug!(
             "Sending cmd w/id {:?}, from {}, to {} Elders w/ dst: {:?}",
@@ -129,26 +100,20 @@ impl Session {
         let msg_kind = MsgKind::ServiceMsg(auth);
         let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, dst_location)?;
 
+        let elders_len = elders.len();
         // The insertion of channel will be executed AFTER the completion of the `send_message`.
-        let (sender, mut receiver) = channel::<CmdResponse>(targets_count);
+        let (sender, mut receiver) = channel::<CmdResponse>(elders_len);
         let _ = self.pending_cmds.insert(msg_id, sender);
         trace!("Inserted channel for cmd {:?}", msg_id);
 
-        let res = send_msg(
-            self.clone(),
-            elders.clone(),
-            wire_msg,
-            self.endpoint.clone(),
-            msg_id,
-        )
-        .await;
+        let res = send_msg(self.clone(), elders, wire_msg, msg_id).await;
 
         if res.as_ref().is_err() {
             // shortcircuit the ack awaiting
             return res;
         }
 
-        let expected_acks = std::cmp::max(1, targets_count * 2 / 3);
+        let expected_acks = std::cmp::max(1, elders_len * 2 / 3);
         // We are not wait for the receive of majority of cmd Acks.
         // This could be further strict to wait for ALL the Acks get received.
         // The period is expected to have AE completed, hence no extra wait is required.
@@ -217,7 +182,7 @@ impl Session {
         query: DataQuery,
         auth: ServiceAuth,
         payload: Bytes,
-    ) -> Result<QueryResult, Error> {
+    ) -> Result<QueryResult> {
         let endpoint = self.endpoint.clone();
 
         let chunk_addr = if let DataQuery::GetChunk(address) = query {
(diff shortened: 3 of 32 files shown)


joshuef:
I'm aware this isn't finished yet, but some thoughts:

My first impression of this was that it looks like we're moving back towards a `ConnectionPool`... which is something @connec spent a long time getting us away from. 

That's not a critique of this approach, just a note that we should be careful with what we're doing and why. 

---

The connection queue is neat 👍  try each as we send, any fail and its removed and we try another or make a new one. This makes a lot of sense.

Though I do worry about `u8::Max` conns.  This could be an attack vector, with each conn being quite memory intensive. If the first one is working, we don't seem to clear up any more... (should we? I'm not sure).

I'm also concerned about the central pool of conns. It's quite neat having each `NamedPeer` hold it's own connection. Dropping the peer drops conns. Nice and clean.

So I'm wondering if we can't have `NamedPeers` holding their own connections? It could be the same as `Connections::new` as we have here. But we'd be checking our current network knowledge on incoming conns to add to existing `NamedPeer`s in our knowledge or... doing _something_ with this unexpected connection... 

Indeed, maybe there `u8::Max` is as good as any other number for max connections. But over a threshold we can (down the line) trigger some `deviant` functionality...

---

It looks to me like `NamelessPeer` is only used during bootstrap. And there, we _should_ be able to avoid that, perhaps by removing the concept of `node_connection_info.config` and just passing around a prefixmap? (though that may be for another PR... ) 

But otherwise we have `NamedPeers` all the time. Which is a nice improvement over the `UnknownPeer` struct for sending to ourselves 👍 👍

oetyng:
> My first impression of this was that it looks like we're moving back towards a `ConnectionPool`... which is something @connec spent a long time getting us away from.

We didn't actually move away from it though, since we in effect had a pool in the SAP, where we had all nodes (and connections to them, since peer held it).d And then clients out there somewhere else.

> Though I do worry about `u8::Max` conns. This could be an attack vector, with each conn being quite memory intensive. If the first one is working, we don't seem to clear up any more... (should we? I'm not sure).

I wrote some code yesterday where unused conns expire. Will push soon.
With or without this, the attack is possible: flood the node with incoming connections.

> I'm also concerned about the central pool of conns. It's quite neat having each `NamedPeer` hold it's own connection. Dropping the peer drops conns. Nice and clean.

I'm on the other hand concerned about having it all spread out for any code to use at any time without any coordination/limits. And still, it is only a chimera that there is no pool, since we hold a sap full of peers which have connections..

With this code there is no problematic Q about which connection a node uses (the one I initiated or the one they initiated), it is abstracted away and they will converge on the same connection. Also there is no burst of connections created when sending a burst of messages.
While both nodes run this code (i.e. not some hacky impl on the other side), then there will be 1 connection only between the nodes, regardless of how many messages are being sent.

maqi:
> > I'm also concerned about the central pool of conns. It's quite neat having each `NamedPeer` hold it's own connection. Dropping the peer drops conns. Nice and clean.
> 
> I'm on the other hand concerned about having it all spread out for any code to use at any time without any coordination/limits. And still, it is only a chimera that there is no pool, since we hold a sap full of peers which have connections..

I think the new code is to have the new struct of `Connections` to hold all communications to other nodes.
And the new struct of `Connection` is holding the list of connections to a particular peer.
Hence, I think it is in the track of `having each NamedPeer hold it's own connection`. though maybe some renaming might make it more obvious

oetyng:
> Indeed, I see that. But if it were all coordinated through a NamedPeer, and that's in our network knowledge then I don't see that as an issue? Comms become "Give me a peer".
> What I'm suggesting is that we have connections in a defined place, which is tied to our network knowledge. But connections to each peer and coordinated by that `NamedPeer`. (But therein we have the connection handling as you've laid out).

Yeah, I see what you mean.

There are some subtleties in current NetworkKnowledge impl that needs to be handled for that, but I can get on to it and try it out.


> Just for clarity: They could both still initiate at the same time and end up with two conns in the pool no? I don't see how this PR prevents this situation? It will be cleared up and converge. (As long as they don't keep messaging at the exact same time which is implauible). But it there could still be more than one conn at a time I think?

Yes, they can _start out_ with two, but very quickly end up with just one. So there "beeing" just one, refers to the time after initial contact done with.


> Yeh for sure. But good not to let things get our of control 👍

The capacity is such a limit to not let things go out of control if other side is malicious - so it's not the problem.
Not detecting excess connections being made and blocking that node (as you point out), is the problem.

If both run this same code, then there will be just 1 connection between the nodes at any given time (after initial potential concurrent calls).

oetyng:
> As far as I can see there's no blocker to having a `Connections` as you've laid in the PR but per `NamedPeer`. Which to me seems like the best of both worlds?
> Indeed, I see that. But if it were all coordinated through a NamedPeer, and that's in our network knowledge then I don't see that as an issue? Comms become "Give me a peer".
> What I'm suggesting is that we have connections in a defined place, which is tied to our network knowledge. But connections to each peer and coordinated by that NamedPeer. (But therein we have the connection handling as you've laid out).

There are actually issues with this (which is basically why we are here refactoring from start).
The Peer is/was passed around so much that it is by no means that simple that dropping the peer from SAP loses it.
That's why it is so perilous to send around an I/O resource in messages going all over the place.

It seems like these two approaches cannot be combined readily as is now. It becomes very complicated.
The first approach had the ambition to cleanly maintain connections by what our section members are. It's not doing that though. The struct is used all over, and the connection is leaked all over.

The second approach solves the problems arising from the failed ambition.
To achieve the "connection dropped when peer dropped" with second approach, we simply call `comms.rs` to remove a connection when a peer is not current anymore. That is trivial.

joshuef:
>  The Peer is/was passed around so much that it is by no means that simple that dropping the peer from SAP loses it.

Yeh this may be the crux of another issue. (Although I think if dropped from the SAP, all conns will _eventually_ be dropped as in flight cmds play out and drop anything they're holding).

I wonder if we refactor how/when we target messages going out, we need to refactor to have SAP access as the last step, to get the Conn _at the send point only_ (but from or x-referencing our `NetworkKnowledge` or some such... 🤷 . Just throwing around ideas just now).


> The second approach solves the problems arising from the failed ambition.
To achieve the "connection dropped when peer dropped" with second approach, we simply call comms.rs to remove a connection when a peer is not current anymore. That is trivial.

Yeh this is where it all plays into the issue you've outlined above. (a `NamedPeer` is attempted to be connected to _after_ the SAP is updated and its dropped... we'd never lose that conn now; leaving the SAP wont happen again).

🧠 ⚡ 
Hmmmm. We need a reliable way to mark a connection as part of our `NetworkKnowledge` perhaps? Do we want to tie conns to our current SAP's PK? Updated on send? We could remove old conns after X keys then?

maqi:
> We need a reliable way to mark a connection as part of our `NetworkKnowledge` perhaps? Do we want to tie conns to our current SAP's PK?

You mean a connection to a peer is preferred to be retained even after that peer removed from NetworkKnowledge ? 
Then I feel this means we will have to hold such peer-less connections in a `Connection pool` like stuff?
If we say a connection is always dropped when that peer is removed from NetworkKnowledge, then we shall always able to do so via the `drop` api ?
So, I am not see tie connection with pk will be much useful?

or, you mean in this new refactory, the `expire` is defined by the position along the pk-chain?

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+// we do, they would open very few connections).
+// 255 is way more than we need and expect, thus gives ample room for
+// unforseen bursts, but at the same time puts a sane cap on the max
+// number continuously held for an - obviously malfunctioning - peer (i.e. edge case).
+const CAPACITY: u8 = u8::MAX;
+const UNUSED_TTL: Duration = Duration::from_secs(120);
+
+///
+#[derive(Clone, Debug)]
+pub(crate) struct Connection {
```
maqi, line 40: this acutally contains connections to a peer node.
The name of `Connection` is kind of confusion, especially with the normal concept of qp2p::Connection.

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+    id: PeerId,
+    endpoint: Endpoint,
+    create_mutex: Arc<Mutex<usize>>,
+    data: Arc<RwLock<BTreeMap<ConnId, ExpiringConn>>>,
```
maqi, line 44: the name of `data` is bit vague, maybe make it more explicit, say `working_connections` ?

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    pub(crate) async fn add(&self, conn: qp2p::Connection) {
+        self.insert(conn).await;
+    }
+
+    /// Disposes of the connection and all underlying resources.
+    /// Also any clones of this connection that are held, will be cleaned up.
+    /// This is due to the fact that we do never leak the qp2p::Connection outside of this struct,
+    /// since that struct is cloneable and uses Arc internally.
+    pub(crate) async fn disconnect(self) {
```
maqi, line 75: this disconnect from the peer totally.
do we also need a function like just drop/close a particular connection (based on connection_id in case of a failure) to the peer among the connections to that?

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+        match conn.send_with(msg, priority, retry_config).await {
+            Ok(()) => {
+                self.remove_expired().await;
+                Ok(())
+            }
+            Err(error) => {
+                // clean up failing connections at once, no nead to leak it outside of here
+                // next send (e.g. when retrying) will use/create a new connection
+                let id = &conn.id();
+                {
```
maqi, line 129: does this bracket pair required?
the lock around data/queue shall get immediately released after the remove call?

sn/src/types/connections/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+use tokio::sync::RwLock;
+use xor_name::XorName;
+
+type PeerId = (XorName, SocketAddr);
+
+/// This is tailored to the use-case of connecting on send.
+/// It keeps a connection instance per node, and is designed to make sure
+/// underlying I/O resources are not leaked, overused or left dangling.
+#[derive(Clone, Debug)]
+pub(crate) struct Connections {
```
maqi, line 24: This is actually a connection map holding all connections to peers?
consider make the naming more explicit?

sn/src/types/connections/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    data: Arc<RwLock<BTreeMap<PeerId, Connection>>>,
```
maqi, line 25: consider use `Arc<DashMap<PeerId, Connection>>` instead
and make the naming being more explicit?

sn/src/types/connections/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    /// Drops the connection and all underlying resources.
+    pub(crate) async fn disconnect(&self, id: PeerId) {
+        let mut list = self.data.write().await;
+        let conn = match list.remove(&id) {
+            // someone else inserted in the meanwhile, so use that
+            Some(conn) => conn,
+            // still not in list, go ahead and create + insert
```
maqi, line 92: comment incorrect

sn/src/types/connections/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+            None => {
+                let conn = Connection::new(*id, self.endpoint.clone());
+                let _ = list.insert(*id, conn.clone());
+                conn
+            }
+        }
+    }
+
+    /// Drops the connection and all underlying resources.
```
maqi, line 86: a mark, this actually drops ALL the connections to that Peer.
worth to double check whether it is expected by the caller, as this is slightly different to the previous implementation (only having one connection to a peer)

sn/src/client/connections/listeners.rs
```diff
@@ -71,15 +75,10 @@ impl Session {
                             "Connection was closed by the node: {:?}",
                             String::from_utf8(reason.to_vec())
                         );
-
-                        mark_connection_id_as_failed(session.clone(), connected_peer.name(), connection_id);
```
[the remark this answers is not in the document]
oetyng: That is encapsulated within the connection. See where send results in error in connection.rs.

sn/src/node/api/dispatcher.rs
```diff
@@ -341,14 +341,18 @@ impl Dispatcher {
...
             MsgKind::ServiceMsg(_) => {
-                if let Err(err) = self
-                    .node
-                    .comm
-                    .send_on_existing_connection_to_client(recipients, wire_msg.clone())
-                    .await
-                {
-                    error!("Failed sending message {:?} to recipients {:?} on existing connection with error {:?}",
-                            wire_msg, recipients, err);
+                // we should never be sending such a msg to more than one recipient
```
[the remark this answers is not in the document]
oetyng: YAGNI.

We implement that when it is something we need and use. Before actuslly needed, such code confuses (and opens up to bugs).

sn/src/node/core/bootstrap/join.rs
```diff
@@ -115,14 +114,13 @@ impl<'a> Join<'a> {
     // - `ResourceChallenge`: carry out resource proof calculation.
     // - `Approval`: returns the initial `Section` value to use by this node,
     //    completing the bootstrap.
-    async fn run(self, bootstrap_peer: UnnamedPeer) -> Result<(NodeInfo, NetworkKnowledge)> {
+    async fn run(self, bootstrap_peer: NamelessPeer) -> Result<(NodeInfo, NetworkKnowledge)> {
```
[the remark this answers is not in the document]
oetyng: Could be used as well.

sn/src/types/connections/connection.rs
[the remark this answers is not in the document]
oetyng: The purpose of not using dashmap there is to have more fine grained control over concurrency.

Dashmap doesn't do that, it just allows to add concurrently. Subtle difference.

In these particular cases we use the rwlock to get a "create only one" where dashmap would give "last writer wins".

The difference is subtle but essential for this case.

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+const UNUSED_TTL: Duration = Duration::from_secs(120);
+
+///
+#[derive(Clone, Debug)]
+pub(crate) struct Connection {
+    id: PeerId,
+    endpoint: Endpoint,
+    create_mutex: Arc<Mutex<usize>>,
+    data: Arc<RwLock<BTreeMap<ConnId, ExpiringConn>>>,
+    queue: Arc<RwLock<DoublePriorityQueue<ConnId, Priority>>>,
```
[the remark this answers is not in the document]
oetyng: The purpose of not using dashmap there is to have more fine grained control over concurrency.

Dashmap doesn't do that, it just allows to add concurrently. Subtle difference.

In these particular cases we use the rwlock to get a "create only one" where dashmap would give "last writer wins".

The difference is subtle but essential for this case.

sn/src/types/connections/connection.rs
oetyng: Yes, a more distinct name would be better.

sn/src/types/connections/connection.rs
oetyng: No, we don’t need to expose that. There is only one connection from the viewpoint of the caller to this code.

The cleanup of individual `qp2p::Connections` is encapsulated within the `Connection`. See where send results in error in `connection.rs`.

sn/src/types/connections/connection.rs
oetyng: Wasn't sure. Are you 100% certain?

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+
+            // update all prios, starting from zero prio again
+            // the iter is sorted from lowest to highest, and the first call after prio == u64::MAX will overflow and give 0.
+            for (id, _old_prio) in clone.into_sorted_iter() {
+                let _ =
+                    queue.change_priority(&id, self.access_counter.fetch_add(1, Ordering::SeqCst));
+            }
+
+            // return next prio to the original caller
+            self.access_counter.fetch_add(1, Ordering::SeqCst)
```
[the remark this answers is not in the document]
oetyng: We initiate it to zero.

At every access it is incremented.

Right before it overflows (at u64::MAX) we reset all prios.
Then we add to it, making it overflow, i.e. it's at zero again.

There is no subtraction.

sn/src/types/connections/connection.rs
```diff
@@ -0,0 +1,345 @@
...
+            }
+        }
+    }
+
+    /// Remove connections that exceed capacity, oldest first.
+    async fn drop_excess(&self) {
+        let len = { self.queue.read().await.len() };
+        if len >= CAPACITY as usize {
+            // remove the least recently used connections
+            let popped = { self.queue.write().await.pop_min() };
```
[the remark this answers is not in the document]
oetyng: Every access fetches the item with highest prio (`peek_max`). It touches the item, which increases its prio (fetch add access counter and change prio with that number).

Popping the item with lowest prio will pop the least recently used item (`pop_min`).

sn/src/types/connections/mod.rs
oetyng: The purpose of not using dashmap there is to have more fine grained control over concurrency.

Dashmap doesn't do that, it just allows to add concurrently. Subtle difference.

In these particular cases we use the rwlock to get a "create only one" where dashmap would give "last writer wins".

The difference is subtle but essential for this case.

sn/src/client/connections/messaging.rs
```diff
@@ -571,39 +491,72 @@ impl Session {
...
-    let new_failed_conns = if let Some((_name, mut conns)) = prev {
-        conns.push(connection_id);
-        conns
-    } else {
-        vec![connection_id]
-    };
-    let _nonexistant_prior_entry = session
-        .elder_last_closed_connections
-        .insert(peer_name, new_failed_conns);
+    async fn get_query_elders(&self, dst: XorName) -> Result<(bls::PublicKey, Vec<NamedPeer>)> {
```
maqi, line 495: get_query_elders and get_cmd_elders are almost the same except using different number to trim the candidates.
can have a further deduplication?

sn/src/client/connections/messaging.rs
```diff
@@ -129,26 +102,20 @@ impl Session {
         let msg_kind = MsgKind::ServiceMsg(auth);
         let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, dst_location)?;
 
+        let elder_count = elders.len();
```
maqi, line 105: elders_len to be consistent with the other naming

sn/src/client/connections/messaging.rs
```diff
@@ -228,23 +195,9 @@ impl Session {
...
-
-        // We select the NUM_OF_ELDERS_SUBSET_FOR_QUERIES closest Elders we are querying
-        let chosen_elders: Vec<_> = elders
-            .into_iter()
-            .take(NUM_OF_ELDERS_SUBSET_FOR_QUERIES)
-            .collect();
-
-        let elders_len = chosen_elders.len();
+        let elders_len = elders.len();
         if elders_len < NUM_OF_ELDERS_SUBSET_FOR_QUERIES && elders_len > 1 {
```
maqi, line 201: this block of code can actually within the `get_query_elders` function to further improve the readability ?

sn/src/client/connections/messaging.rs
```diff
@@ -513,33 +452,14 @@ impl Session {
                         // mark as tried all
                         tried_every_contact = true;
 
-                        // Now mark prior attempted Peers as failed for retries in future.
```
maqi, line 516: does this mean in such case we can say this connection_id to that peer can be closed/dropped ?
note: just that partically connection, not ALL connections to that peer

sn/src/client/connections/messaging.rs
```diff
@@ -614,90 +567,36 @@ pub(super) async fn send_msg(
...
+                Session::spawn_msg_listener_thread(
+                    session.clone(),
+                    peer.clone(),
+                    conn,
+                    incoming_msgs,
+                );
+            };
+
+            let result = match connection
+                .send_with(msg_bytes_clone, priority, None, listen)
```
maqi, line 588: a note: this `send_with` only use one connection to peer to send.
In case of sending failure, shall other connections to that peer get tried to be used to send as well ?

also, here is the naming of `Connections` and `Connections::Connection` causing confusion, as the `connection` here is actually an instance holding a list of `qp2p::connection`s to the same peer.

sn/src/client/connections/messaging.rs
```diff
@@ -707,28 +606,28 @@ pub(super) async fn send_msg(
...
     for r in results {
         match r {
-            Ok((peer_name, connection_id, send_result)) => match send_result {
+            Ok((peer_name, send_result)) => match send_result {
                 Err(Error::QuicP2pSend(SendError::ConnectionLost(ConnectionError::Closed(
                     Close::Application { reason, error_code },
                 )))) => {
                     warn!(
-                        "Connection was closed by the node: {:?}",
+                        "Connection was closed by node {}, reason: {:?}",
```
maqi, line 614: better includes the `connection_id` in the log.
Also, would like to confirm that here is the `ALL connections` to that peer got closed, or just `one of the connections` to that peer got closed.

sn/src/client/connections/messaging.rs
```diff
@@ -707,28 +606,28 @@ pub(super) async fn send_msg(
...
-                    mark_connection_id_as_failed(session.clone(), peer_name, connection_id);
                     last_error = Some(Error::QuicP2pSend(SendError::ConnectionLost(
                         ConnectionError::Closed(Close::Application { reason, error_code }),
                     )));
                 }
                 Err(Error::QuicP2pSend(SendError::ConnectionLost(error))) => {
-                    warn!("Connection was lost: {:?}", error);
-
-                    mark_connection_id_as_failed(session.clone(), peer_name, connection_id);
+                    warn!("Connection to {} was lost: {:?}", peer_name, error);
```
maqi, line 623: same comment as above.
better include connection_id in the log.
and make sure it's ALL connections to the peer closed or just one of the connections to the peer closed.

sn/src/client/connections/messaging.rs
```diff
@@ -707,28 +606,28 @@ pub(super) async fn send_msg(
...
                     last_error = Some(Error::QuicP2pSend(SendError::ConnectionLost(error)));
                 }
                 Err(error) => {
-                    warn!("Issue during {:?} send: {:?}", msg_id, error);
+                    warn!(
```
maqi, line 627: better include the connection_id in the log

sn/src/node/api/dispatcher.rs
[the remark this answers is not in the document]
maqi: yeah.
However the current approach(`.get(0)`) will carry out work doesn't match the API.
Which may cause the caller misuse the function.
maybe change the function API to only take one recipient. Or put an assertion to ensure only one recipient provided ?

sn/src/types/connections/connection.rs
maqi: However, in edge case that we have two qp2p::Connection in hold and failed to send msg using one of them.
then we supposed to switch to the other one.
and the current naming could cause confusion and misuse. 
as I mentioned at https://github.com/maidsafe/safe_network/pull/1026#discussion_r812059494

sn/src/types/connections/connection.rs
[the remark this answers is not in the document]
maqi: >Popping the item with lowest prio will pop the least recently used item (pop_min)

but could also be the latest one hasn't get a chance to be used?
Say conn_A inserted first then used, it's priority is then 1. then conn_B inserted, with priority 0
then a pop_min here will give conn_B ?

also, pop_max gives an `over weight` to old connection that get inserted first and be used for once.
maybe it's good, however normally the later inserted connection shall be the better connection in good chance?

sn/src/types/connections/connection.rs
maqi: at least clippy doesn't complain with the removal of the brackets

sn/src/client/client_api/mod.rs
```diff
@@ -207,11 +204,13 @@ impl Client {
...
             } else {
-                // these peers will be nonsense peers, and dropped after we connect. Reaplced by whatever SectionAuthorityProvider peers we have received
+                // these peers will be nonsense peers, and dropped after we connect. Replaced by whatever SectionAuthorityProvider peers we have received
+                // therefore we use a dummy name (our own) for them initially
+                let dummy_name = client.public_key().into();
                 bootstrap_nodes
                     .iter()
                     .copied()
-                    .map(|socket| Peer::new(XorName::random(), socket))
+                    .map(|socket| NamedPeer::new(dummy_name, socket))
```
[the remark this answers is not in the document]
oetyng: I think there was a use of our name (with comment about it) at some place and that it got moved here, but not sure. Either way, the random would work fine.

sn/src/node/api/dispatcher.rs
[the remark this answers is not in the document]
oetyng: That was the intention, but that refactor got very sprawly, so had to cap it somewhere, to be continued in some other PR.
Assertion would not be so good I think, I think we don't want any of those in the code?

So, the case is that currently we do not have a system where we send a msg to more than one client at a time.
Our code allowed it though, and would let such a bug slip by without notice.

What happens now is that we won't let it slip through, and we will log it if it does happen. It is quite consistent with how we solve such things in other places now. However, that is not what I'd consider optimal. I'd rather refactor it all the way up, but as mentioned, it got too sprawly and deviated too much from this PR then.

sn/src/types/connections/connection.rs
oetyng: The caller would try to send on their connection, it fails and they would retry, as many times as their retry policy states.

The caller has no notion of switching to another connection.

They have one connection. If sending a msg fails then _under the hood_, the connection disposes of that failed qp2p::Connection, and the caller who retries will (unknowingly) pick next qp2p::Connection or create a new one.

sn/src/types/connections/connection.rs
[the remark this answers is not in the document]
oetyng: > Say conn_A inserted first then used

Insertion is when calling send. So it is one and the same thing. There is no insertion and then _not_ use it.

> then conn_B inserted, with priority 0

If conn_B inserts (sends) after conn_A, then conn_B priority will be 1 higher than conn_A.
Next access will get conn_B.

> then a pop_min here will give conn_B ?

It would give conn_A.

sn/src/client/connections/messaging.rs
oetyng: > In case of sending failure, shall other qp2p::connections to that peer get tried to be used to send as well ?
It could. 
However, as most often there will be just 1 connection there, it means you need to create a new one. And the question becomes how many retries do you do. And at that point it becomes obvious I think, that it is upper layer that call connection, that decides the retry policy.

It wasn't there before, I considered introducing it, but think that is for a follow up PR.

>  as the connection here is actually an instance holding a list of qp2p::connections to the same peer.
The goal is to have our node layer not actually need to know about qp2p::Connections.
At that point, there is a Connection. It encapsulates the complexity, instead of spreading it around in our node code.
Now, that for that encapsulation to be complete, a bit more work will be needed. But I think the goal with this PR is first and foremost to give a bit more stability, and set the path forward, which following PRs will continue on.

sn/src/client/connections/messaging.rs
oetyng: This is code in between two states of refactor actually.

Qp2p errors shouldn't be handled at this level, and there should be no concern for the client about qp2p::Connections closed or not.

sn/src/client/connections/messaging.rs
oetyng: Yeah, could do that. But this is code in between two states of refactor actually.

Qp2p errors shouldn't be handled at this level, and there should be no concern for the client about qp2p::Connections closed or not.

sn/src/client/connections/messaging.rs
oetyng: Yeah, could do that. But this is code in between two states of refactor actually.

Qp2p errors shouldn't be handled at this level, and there should be no concern for the client about qp2p::Connections closed or not.

sn/src/client/connections/messaging.rs
oetyng: could probably and that would be nice, I'll defer that for now though I think

sn/src/node/core/data/records/mod.rs
```diff
@@ -264,8 +264,11 @@ impl Node {
         candidates
     }
 
-    // Takes a message and forms cmds to send to specified targets
+    // Takes a message for specified targets, and builds internal send cmds
+    // for sending to each of the targets.
     // Targets are XorName specified so must be within the section
+    // (Within which section?? Further down, we seem to expect different sections,
+    // since we call self.section_key_by_name(&target) ???)
```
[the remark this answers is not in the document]
oetyng: I think my comment there serves as a Q about the clarity of the existing comment together with the method signature and body.

Probably, the best thing would be to raise an issue, and remove the additional comment. I think these sorts of comments serves as a placeholder in lack of a vivid issue-tracking culture.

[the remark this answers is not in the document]
maqi: In that case, then yeah better raise an issue and remove this comment

sn/src/client/connections/listeners.rs
```diff
@@ -38,30 +40,32 @@ impl Session {
...
+                            session.connections.add_incoming(&node_id, conn.clone()).await;
+                        }
+
+                        if let Err(err) = Self::handle_msg(msg, peer.clone(), session.clone()).await {
                             error!("Error while handling incoming msg: {:?}. Listening for next msg...", err);
                         }
                     },
                     Ok(None) => {
-                        info!("Incoming msg listener has closed for connection {}.", connection_id);
+                        // once the msg loop breaks, we know the connection is closed
```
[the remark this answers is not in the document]
oetyng: Cleanup would be taken care of within the link.

> also, the log was useful for the debugging work, why removed?

I thought the log on the subsequent line was enough:
`trace!("{} to {} (id: {})", LogMarker::ConnectionClosed, addr, connection_id);`

sn/src/client/connections/messaging.rs
oetyng: > or just one of the connections to that peer got closed.

A single connection got closed. I'm finding ways to make this clearer atm.

sn/src/client/connections/messaging.rs
oetyng: yes

joshuef (approved):

PR #1030, opened 2022-02-21
Pull model

Yoga07:
Refactors the existing push model data replication to pull model. Therefore, on churn, nodes will be notified of data they should be holding and the nodes will be then responsible for pulling it from the other replicas.


sn/src/messaging/system/node_msgs.rs
```diff
@@ -45,6 +45,12 @@ pub enum NodeCmd {
     },
     /// Tells an Adult to store a replica of the data
     ReplicateData(ReplicatedData),
+    /// Tells an Adult to fetch and replicate data from the sender
+    SendReplicateDataAddress(ReplicatedDataAddress),
+    /// Fetch the given replicated data we are missing
+    FetchReplicateData(ReplicatedDataAddress),
+    /// Provide requested data for replication
+    ProvideReplicateData(ReplicatedData),
```
oetyng, line 53: These names are a bit off considering that they are placed as NodeCmds.

Will look at other alternatives, just wanted to mention it.

oetyng: As a quick example (not final suggestion):

`SendReplicateDataAddress => ReplicateDataAt` // **tells** a node to replicate data found at given address (we are not telling the node to send something)

`NodeCmd::FetchReplicateData => NodeQuery::Replication(GetData { .. })` // this is a simple data query, albeit between nodes now

`NodeCmd::ProvideReplicateData` .. This should be a query response to the above, as to follow the established pattern and not create new sub-patterns here.
However, it requires a larger refactor, since NodeQueryResponses so far has been `Responses from Adults to Elders, on client initiated queries for data`.

This is the first time we do Adult-to-Adult system msgs.

joshuef: I wonder (as something for _after_ this initial PR) if we _dont_ get a response if we want to trigger a client msg via elders for the data. That will get liveness checks kicked in.

A form of of taddling on adults not behaving.

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,108 @@
...
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::types::ReplicatedDataAddress;
+use std::collections::btree_map::Entry;
+use std::collections::{BTreeMap, BTreeSet};
+use xor_name::XorName;
+
+/// Keeps track of all the data that needs to be replicated.
```
joshuef, line 14: This data the needs to be replicated _at our node_ or _to some node_ ?

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl DataReplicator {
+    pub(crate) fn new() -> Self {
+        DataReplicator {
+            to_be_transmitted: BTreeMap::new(),
+        }
+    }
+
+    // Keep track of nodes and the replicas we need to provide them
```
joshuef, line 27: `///` three slashes for func docs.

This sin't super clear to me what we're doing here.

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,108 @@
...
+        } else {
+            warn!("Given node: {target} not on replicator");
+        }
+    }
+
+    // Returns:
+    // Some(true) if we still need to hold the data after handing out a replica
+    // Some(false) if we can delete the data since we have handed out to all replicas
+    // None if we do not handle the data at all
+    pub(crate) fn get_for_replication(
```
joshuef, line 74: This feels quite confusing.

I think we should probably have separate checks on if we want the data vs if it should be replicated elsewhere

sn/src/messaging/system/node_msgs.rs
happybeing: Taddling?!

Yoga07: > I wonder (as something for _after_ this initial PR) if we _dont_ get a response if we want to trigger a client msg via elders for the data. That will get liveness checks kicked in.
> 
> A form of of taddling on adults not behaving.

Yea, it should be possible to add punish the node by just increasing it's ops counter with a random entry which won't clear out.

sn/src/node/core/data/records/data_replicator.rs
Yoga07: At other nodes. So we keep track of whom all and what data to provide to them.

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,108 @@
...
+        DataReplicator {
+            to_be_transmitted: BTreeMap::new(),
+        }
+    }
+
+    // Keep track of nodes and the replicas we need to provide them
+    pub(crate) fn add_to_transmitter(
+        &mut self,
+        data_address: &ReplicatedDataAddress,
+        targets: &BTreeSet<XorName>,
```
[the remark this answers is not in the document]
Yoga07: Targets are calculated here specifically: https://github.com/maidsafe/safe_network/blob/38fb057da44a0e243186410df0c39361a21ec46e/sn/src/node/core/data/storage/mod.rs#L176

So that provides the targets to the `DataReplicator` whilst sending out the message to notify the adults.

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,108 @@
...
+                    addresses.push(*data_address);
+                }
+                Entry::Vacant(e) => {
+                    let _ = e.insert(vec![*data_address]);
+                }
+            }
+        }
+    }
+
+    // Remove data address from replicator in case the target already has the data
```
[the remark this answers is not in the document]
Yoga07: >If it's a pull model we shouldn't need to do this or?

We need to do this to clear out the cache of addresses that we send out.

>An adult would receive a list of data that it should have and then pull anything it doesn't have?

We can send a list though as of now, every data is pushed in its own message. 
The Idea behind that was we might be underway with ingesting data from the received list, but a parallel notification message from another node would trigger another redundant replication messaging round. So I seemed better to atomically fetch and write every data than to list-ify it. Though it can be done with the current impl too with a few changes.

sn/src/node/core/data/records/data_replicator.rs
Yoga07: I believe it get's a bit confusing because of the naming. Should be addressed in [23a420f](https://github.com/maidsafe/safe_network/pull/1030/commits/23a420fa373c9904fc58caf110225d7d0a2e7946).

sn/src/node/core/data/records/data_replicator.rs
```diff
@@ -0,0 +1,172 @@
...
+
+        // Adults with Prefix 0
+        let mut adults = (0..10)
+            .map(|_| XorName::random().with_bit(0, false))
+            .collect::<Vec<XorName>>();
+
+        // Adults with Prefix 1
+        let adults1 = (0..10)
+            .map(|_| XorName::random().with_bit(0, true))
+            .collect::<Vec<XorName>>();
```
joshuef, line 122: @Yoga07 these two use the same `with_bit`

sn/src/node/core/messaging/handling/mod.rs
```diff
@@ -693,46 +693,152 @@ impl Node {
...
+                        // This may return a DatabaseFull error... but we should have reported storage increase
+                        // well before this
+                        match self.data_storage.store(&data).await {
+                            Ok(level_report) => {
+                                info!("Storage level report: {:?}", level_report);
+                                cmds.extend(self.record_storage_level_if_any(level_report).await);
+                            }
+                            Err(error) => {
+                                match error {
+                                    //DbError::Io(_) | DbError::Sled(_) => false, // potential transient errors
```
joshuef, line 715: remove commented code

sn/src/node/core/messaging/handling/mod.rs
```diff
@@ -693,46 +693,152 @@ impl Node {
...
-                    // We are an adult here, so just store away!
-                    // This may return a DatabaseFull error... but we should have reported storage increase
-                    // well before this
-                    match self.data_storage.store(&data).await {
-                        Ok(level_report) => {
-                            info!("Storage level report: {:?}", level_report);
-                            return Ok(self.record_storage_level_if_any(level_report).await);
+                    let mut cmds = vec![];
+
+                    for data in data_collection {
```
joshuef, line 704: Is this the same flow as coming from elders on initial storage?

sn/src/node/core/messaging/handling/mod.rs
```diff
@@ -693,46 +693,152 @@ impl Node {
...
+                    Ok(vec![Cmd::SignOutgoingSystemMsg {
+                        msg: SystemMsg::NodeCmd(NodeCmd::ReplicateData(data_collection)),
+                        dst: crate::messaging::DstLocation::Node {
+                            name: sender.name(),
+                            section_pk: self.section_key_by_name(&sender.name()).await,
+                        },
+                    }])
                 };
             }
+            SystemMsg::NodeEvent(NodeEvent::ReplicateDataAlreadyPresent(data_addresses)) => {
```
joshuef, line 834: I'm unclear as to what this message is for... 

Either we're asking for some data or not? I'm not sure why we need to inform that we have _some_ data?

joshuef (changes requested):

sn/src/node/core/data/records/data_replicator.rs
Yoga07: Adults with prefix 1 have `with_bit` index `0` set to `true`.

joshuef: ah sorry, i was looking at the bit `0` part there. Okay

sn/src/node/core/messaging/handling/mod.rs
Yoga07: Yep, we can reuse the same flow for writing data to nodes than adding a new message for that atm. Though we can add a new message for this in case we want to separate out Elders given data and Adult given data in the future.

sn/src/node/core/messaging/handling/mod.rs
Yoga07: This was used to notify that we already hold the to-be-replicated data to the node that provides data so that it clears its data after completing replication. This can be removed since we've moved away from deleting data after replication.

sn/src/node/core/data/records/data_replicator.rs
joshuef: Why do we need to keep track at all?

We just say "this is stuff you should have" and they can ask "heres stuff that i dont have"?

joshuef (approved):

PR #1031, opened 2022-02-21
Naming refactor

dirvine:
chore(refactor) Update deps, ED node key rename 

Updated dependencies and removed a test that no longer is valid (signature format test). Serde now reportedly much faster but I have not benchmarked that. Also renamed node: public_key to node_ed_pk tp make it more explicit and not to be confused with the PublicKey type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,6 +322,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "backoff"
+version = "0.4.0"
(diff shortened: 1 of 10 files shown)


joshuef:
@dirvine your commits are missing a `:` to be recognised as [conventional](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with--to-draw-attention-to-breaking-change)

eg: `chore(refactor) Update deps, ED node key rename` -> `chore(refactor): Update deps, ED node key rename`

PR #1033, opened 2022-02-22
chore: add put_get chunk soak example

joshuef:
Adds an example to launch many clients uploading many files, and then
verify those uploads.

This is handy to test node load and client throughput.

Currently it's not used in CI anywhere.

### DIFF
--- a/sn/examples/put_get.rs
+++ b/sn/examples/put_get.rs
@@ -1,0 +1,174 @@
+// Copyright 2022 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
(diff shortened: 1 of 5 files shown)


sn/examples/put_get.rs
```diff
@@ -0,0 +1,174 @@
...
+        let put_handle: tokio::task::JoinHandle<Result<()>> = tokio::spawn(async move {
+            let (address, hash) = upload_data_using_client(client, i).await?;
+            all_data_put.write().await.push((address, hash));
+            Ok(())
+        });
+
+        put_tasks.push(put_handle);
+    }
+
+    futures::future::join_all(put_tasks).await;
```
Yoga07, line 92: Could we log the time/duration it takes to finish up the whole upload after this? And the same could be done for reads below.

Yoga07 (approved):

PR #1044, opened 2022-02-26
chore(logging): display binary prefix in logging code

Thierry61:
Display usual prefix binary representation instead of the name.

This is useful for dashboards analyzing log files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,9 +84,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.55"
+version = "1.0.56"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "159bb86af3a200e19a068f4224eae4c8bb2d0fa054c7e5d1cacd5cef95e684cd"
+checksum = "4361135be9122e0870de935d7c439aef945b9f9ddd4199a553b5270b49c82a27"
 
 [[package]]
 name = "arrayref"
@@ -167,9 +167,9 @@ dependencies = [
 
 [[package]]
 name = "async-global-executor"
-version = "2.0.2"
+version = "2.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9586ec52317f36de58453159d48351bc244bc24ced3effc1fce22f3d48664af6"
+checksum = "c026b7e44f1316b567ee750fea85103f87fcb80792b860e979f221259796ca0a"
 dependencies = [
  "async-channel",
  "async-executor",
@@ -268,9 +268,9 @@ dependencies = [
 
 [[package]]
 name = "async-task"
-version = "4.1.0"
+version = "4.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "677d306121baf53310a3fd342d88dc0824f6bbeace68347593658525565abee8"
+checksum = "30696a84d817107fc028e049980e09d5e140e8da8f1caeb17e8e950658a3cea9"
 
 [[package]]
 name = "async-trait"
@@ -1966,9 +1966,9 @@ dependencies = [
 
 [[package]]
 name = "ipnet"
-version = "2.3.1"
+version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "68f2d64f2edebec4ce84ad108148e67e1064789bee435edc5b60ad398714a3a9"
+checksum = "35e70ee094dc02fd9c13fdad4940090f22dbd6ac7c9e7094a46cf0232a50bc7c"
 
 [[package]]
 name = "isatty"
@@ -2396,9 +2396,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
-version = "1.9.0"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da32515d9f6e6e489d7bc9d84c71b060db7247dc035bbe44eac88cf87486d8d5"
+checksum = "87f3e037eac156d1775da914196f0f37741a274155e34a0b7e427c35d2a2ecb9"
 
 [[package]]
 name = "oorandom"
@@ -3190,9 +3190,9 @@ dependencies = [
 
 [[package]]
 name = "regex"
-version = "1.5.4"
+version = "1.5.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d07a8629359eb56f1e2fb1652bb04212c072a87ba68546a04065d525673ac461"
+checksum = "1a11647b6b25ff05a515cb92c365cec08801e83423a235b51e231e1808747286"
 dependencies = [
  "aho-corasick",
  "memchr",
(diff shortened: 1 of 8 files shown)


joshuef:
@Thierry61 Thanks for this. main change seems fine.

You'll need to update the following logs to be correct though: https://github.com/maidsafe/safe_network/pull/1044/files#diff-87e0a1c9b86a90561838367d66d6f23f01f6e2faf23488ad4dc270c2e58c8a13L66-L67

That's looking for a node's `name` as I understand it.

Also, we [follow conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for commit messages to auto generate changelogs for release. So if you could reformat your message to follow that it would be great 🙇 (eg `chore: display binary prefix...`)

Yoga07:
Hi @Thierry61 ,

This refactor changes printing from:
```
2022-02-28T10:32:54.755072Z DEBUG sn_launch_tool: Launching node #3...
Starting logging to directory: "C:\\Users\\yogesh\\.safe/node\\local-test-network\\sn-node-3"
Node PID: 23888, prefix: Prefix(), name: 0f4b97(00001111).., age: 96, connection info:
"127.0.0.1:49364"
"000000.." Node resource usage: Process { memory: 73457, virtual_memory: 4762624, cpu_usage: 4.166667, disk_usage: DiskUsage { total_written_bytes: 258504, written_bytes: 19690, total_read_bytes: 852716, read_bytes: 158 } }
```
to this:
```
2022-02-28T10:18:10.663954Z DEBUG sn_launch_tool: Launching node #14...
Starting logging to directory: "C:\\Users\\yogesh\\.safe/node\\local-test-network\\sn-node-14"
"(0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000)" Node resource usage: Process { memory: 79785, virtual_memory: 4741402, cpu_usage: 0.35047743, disk_usage: DiskUsage { total_written_bytes: 5550740, written_bytes: 5349776, total_read_bytes: 855853, read_bytes: 2620 } }
Node PID: 36684, prefix: Prefix(), name: 05f2b2(00000101).., age: 64, connection info:
"127.0.0.1:64759"
```
Is that verbose logging intended? Also FYI, printing the type `Prefix` prints the bits of the xorname prefix eg: `Prefix(01011..)` and strips the trailing zeros.

Thierry61:
> if you could reformat your message to follow that it would be great

I modified it.

If it is still not OK, you should be able to do it yourself, as "Allow edits by maintainers" button is checked.

Thierry61:
> printing the type `Prefix` prints the bits of the xorname prefix and strips the trailing zeros

No, printing the type Prefix doesn’t do this. Previously the name of the prefix (`ctx.prefix().await.name()`) was printed which does this, but is not the prefix (`ctx.prefix().await`)

For whatever reason when this function is called in your test the prefix has 256 bits, which is not a normal condition but is the reality. This reality shouldn’t be hidden by printing the prefix name instead of the usual prefix representation.

Or, you can hide the reality in this very specific condition, but this shouldn’t impact usual prefix representation in normal conditions (and I let you code this specific case!).

Thierry61:
> That's looking for a node's `name` as I understand it.

I don’t understand, I just replaced printing the name of the prefix by printing the prefix.

Please note that I only tested the code with an empty prefix, my network was too small to trigger splits.

joshuef:
> I don’t understand, I just replaced printing the name of the prefix by printing the prefix.

Ah sorry, I appear to have grabbed the wrong link there. 🤦 

https://github.com/maidsafe/safe_network/pull/1044/files#diff-87e0a1c9b86a90561838367d66d6f23f01f6e2faf23488ad4dc270c2e58c8a13L66-L67

Here the `prefix` var is used expecting it to be `name`. :| Which yes, We need the variable naming to be updated here as that's obviously wrong.

But basically, updating the `prefix` var to be `prefix` and not `name` will mess with the pending logs. So we just need a `name` var created and used in those logs below too.
----

> I modified it.

It's still missing the scope. (https://www.conventionalcommits.org/en/v1.0.0/#summary)

If you can get that in if there are more changes that'd be great. 
----

>This reality shouldn’t be hidden by printing the prefix name instead of the usual prefix representation.

Yeh that's fair enough 👍 

@Thierry61 Just for clarity, what are you looking to achieve here? What is missing from a log such as `prefix: Prefix(101)` eg?

Thierry61:
> Just for clarity, what are you looking to achieve here?

As mentioned in [this topic](https://safenetforum.org/t/update-24-february-2022/36669/14?u=tfa) I didn’t find a proper way to get current node prefix in the logs:

- “Known prefix after update:” is generated in a loop which means several of them can be listed
- “Section updated: prefix:” is reserved for elders and unavailable for adults
- prefix at the end of “Node resource usage:” isn’t in the right format (000000… instead of ())

This PR is about the third point. I am not sure this is the best way to get the prefix for the dashboard, but anyway the prefix should be displayed using its usual binary representation and not its name.

But if you have a better way to get the prefix of current node I am interested.

Thierry61:
> But basically, updating the `prefix` var to be `prefix` and not `name` will mess with the pending logs

I don't use print_resources_usage flag, so what can be done is to move definition of prefix variable and duplicate it in 2 versions:
- original one (`let prefix: &str = &format!("{}", ctx.prefix().await.name());`) in print_resources_usage branch
- new one in (`let prefix: &str = &format!("({:b})", ctx.prefix().await);`) in else branch

This way your logs (or more exactly your stdout) are not messed with.

But you are aware that prefix name doesn't represent much.

Yoga07:
Hey @Thierry61 ,

By type `Prefix` I meant the struct itself and not the variable `prefix`. So for clarification:
```
println!("{:?}", format!("{:?}", ctx.prefix().await));
prints: "Prefix(0)" 

println!("{:?}", format!("{:?}", ctx.prefix().await.name()));
prints: "000000(00000000).."
```

>This PR is about the third point. I am not sure this is the best way to get the prefix for the dashboard, but anyway the prefix should be displayed using its usual binary representation and not its name.
>But if you have a better way to get the prefix of current node I am interested.

Do the above types of formatting not suffice your requirements? It should be easy to get the only prefix out of `Prefix(xxx)` yes? (i.e the trailing zeros are removed already)

Thierry61:
> Do the above types of formatting not suffice your requirements?

First formatting (`Prefix(0)`) would be OK.

But second one (`000000(00000000)..`) obviously not because `00000000` is different from `0` (prefix depth is not right).

Thierry61:
I committed modification suggested by @Yoga07.

Another remark is the unusual trace statement that doesn’t adhere to the principle of least surprise (prefix is the first argument but appears last in the output).

https://github.com/maidsafe/safe_network/blob/507aa3cf0f23ab149c5865e1d9d0c11ddde1c39e/sn/src/node/logging/mod.rs#L71-L75

Why not a standard trace statement with format string as first argument ?

Yoga07:
> Another remark is the unusual trace statement that doesn’t adhere to the principle of least surprise (prefix is the first argument but appears last in the output).
> 
> https://github.com/maidsafe/safe_network/blob/507aa3cf0f23ab149c5865e1d9d0c11ddde1c39e/sn/src/node/logging/mod.rs#L71-L75
> 
> Why not a standard trace statement with format string as first argument ?

Absolutely. I've opened https://github.com/maidsafe/safe_network/pull/1062 including this change just to have the commits 
 a bit neater. 
Thanks, @Thierry61 :)

PR #1049, opened 2022-03-03
Ci tweaks

joshuef:


jacderida (approved):

joshuef:
#1053 takes this but adds a conn retry at node in case the `link` conn failed for any reason, we basically try creating a fresh conn at least once

PR #1058, opened 2022-03-09
Feat dysfunctional detection

joshuef:
This is just an initial PR separating out dysfunctional node detection from the rest of the codebase.
This hopefully gives us a clear place for any logic relating to this (and any action to be taken from this, such as `ProposeOffline` should be done _only_ after dysfunction has been detected). I'm hopeful we can improve things with more methods for detection and improved calculation of issues (with more testing) as we go.

The `DysfunctionDetector` will take in a few paramaters by node name, and when need, weigh any issues against the node's neighbours, allowing for more holistic view of network issues.

We can then call `get_suspicious_node_names` or `get_dysfunctional_node_names` depending on our use case and take action upon the resulting nodes.


dys_function/src/connections.rs
```diff
@@ -0,0 +1,191 @@
...
+    /// Calculate a score of this node, as compared to its closest neighbours...
+    pub(crate) fn calculate_connections_score(&self) -> BTreeMap<XorName, f32> {
+        let mut score_map = BTreeMap::default();
+        // loop over all node/neighbour comparisons
+        for (node, neighbours) in self.get_node_and_neighbours_vec() {
+            let nodes_count = if let Some(entry) = self.communication_issues.get(&node) {
+                *entry.value()
+            } else {
+                0
+            };
```
davidrusu, line 33:
```suggestion
            let nodes_count = self.communication_issues
                .get(&node)
                .map(|entry| *entry.value())
                .unwrap_or(0);
```

dys_function/src/connections.rs
```diff
@@ -0,0 +1,191 @@
...
+
+            let mut all_neighbourhood_counts = vec![];
+            for neighbour in neighbours {
+                if let Some(entry) = self.communication_issues.get(&neighbour) {
+                    let val = *entry.value();
+
+                    all_neighbourhood_counts.push(val as f32);
+                }
+            }
```
davidrusu, line 42:
```suggestion
            let all_neighbourhood_counts = Vec::from_iter(
                neighbours.iter().filter_map(|n| self.communication_issues.get(n))
            );
```

davidrusu: @joshuef shouldn't we be including `0`'s for nodes without communication issues? Otherwise with the current logic, 1 node having communication issues, would mean that the 1 node will be the only value in the mean list so their score will be equal to the mean.

joshuef: 👍 just curious: is this more efficient? or just a bit easier on the eyes/brain?

dys_function/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+/// Calculates the avg value in a data set
+/// https://rust-lang-nursery.github.io/rust-cookbook/science/mathematics/statistics.html
+pub(crate) fn get_mean_of(data: &[f32]) -> Option<f32> {
+    let sum = data.iter().sum::<f32>();
+    let count = data.len();
+    match count {
+        positive if positive > 0 => Some(sum / count as f32),
+        _ => None,
+    }
```
davidrusu, line 173: I think an `if count > 0 { .. }` would be clearer here

dys_function/src/network_knowledge.rs
```diff
@@ -0,0 +1,229 @@
...
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::{get_mean_of, DysfunctionDetection, NodeIdentifier};
+use std::collections::BTreeMap;
+use xor_name::XorName;
+
+impl DysfunctionDetection {
+    /// Track an issue with a node's network knowledge
+    pub fn track_knowledge_issue(&self, node_id: NodeIdentifier) {
```
davidrusu, line 15: This module looks to be identical to the connection tracking code.

Can we have one function that gets called with the respective tracker data:
```rust
self.track_issue(node_id, &self.knowledge_issues)
self.track_issue(node_id, &self.connection_issues)
```
```rust
let knowledge_score = self.calculate_score(node_id, &self.knowledge_issues)
let connection_score = self.calculate_score(node_id, &self.connection_issues)
```

dys_function/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+type OperationId = [u8; 32];
+
+#[derive(Clone, Debug)]
+/// Dysfunctional node tracking. Allows various potential issues to be tracked and weighted,
+/// with unresposive or suspect nodes being noted on request, against which action can then be taken.
+pub struct DysfunctionDetection {
+    neighbour_count: usize,
+    communication_issues: Arc<DashMap<NodeIdentifier, usize>>, // count of comm issues
+    knowledge_issues: Arc<DashMap<NodeIdentifier, usize>>,     // count of comm issues
+    unfulfilled_ops: Arc<DashMap<NodeIdentifier, Arc<RwLock<Vec<OperationId>>>>>, // OperationId = [u8; 32]
```
davidrusu, line 71: We could replace the count with the (local) timestamp of when the dysfunction occurred. (You can read unix timestamp as u64 like so: https://doc.rust-lang.org/std/time/struct.SystemTime.html#examples)

```suggestion
    communication_issues: Arc<DashMap<NodeIdentifier, Arc<RwLock<VecDeque<u64>>>>>, // count of comm issues
    knowledge_issues: Arc<DashMap<NodeIdentifier, Arc<RwLock<VecDeque<u64>>>>>,     // count of knowledge issues
```

That way we can dequeue old comm/knowledge issues, i.e. remove issues older than 10 minutes ago. (Also, fault detection logic starts to look more uniform across unfilled_ops and comm/knowledge_issues)

dys_function/src/operations.rs
```diff
@@ -0,0 +1,252 @@
...
+
+    /// Calculate a score of this node, as compared to its closest neighbours...
+    /// The average score of all neighbours is calculated, and the standard deviation therein.
+    /// We then calculate the node's z-score and multiply excess pending ops by this
+    pub(crate) async fn calculate_ops_score(&self) -> BTreeMap<XorName, f32> {
+        let mut score_map = BTreeMap::default();
+        // loop over all node/neighbour comparisons
+        for (node, neighbours) in self.get_node_and_neighbours_vec() {
+            // let node = *node;
+            let nodes_count = if let Some(entry) = self.unfulfilled_ops.get(&node) {
```
Yoga07, line 83: Should this be `ops_count`? Since it's the pending operations we are fetching.

dys_function/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    /// Removes any tracked nodes not present in the passed `current_members`
+    pub fn retain_members_only(&self, current_members: BTreeSet<XorName>) {
+        let all_keys: Vec<_> = self.current_nodes();
+
+        for key in &all_keys {
+            if !current_members.contains(key) {
+                let _prev = self.communication_issues.remove(key);
+                let _prev = self.unfulfilled_ops.remove(key);
+                let _prev = self.closest_nodes_to.remove(key);
```
Yoga07, line 142: Should we reduce the `self.neighbour_count` since we are removing a node from the current set?

dys_function/src/detection.rs
```diff
@@ -0,0 +1,114 @@
...
+
+        node_neighbours
+    }
+
+    async fn get_weighted_scores(&self) -> (BTreeMap<XorName, f32>, f32) {
+        trace!("Getting weighted scores");
+        let ops_scores = self.calculate_ops_score().await;
+
+        let conn_scores = self.calculate_connections_score();
+        let knowledge_scores = self.calculate_knowledge_score();
```
Yoga07, line 41: The 'average among neighbours' calc part of the logic within these functions seems to be duplicate. Would be nice if we could merge them.

sn/src/node/core/mod.rs
```diff
@@ -207,6 +212,24 @@ impl Node {
...
+    pub(crate) async fn get_dysfunctional_names(&self) -> BTreeSet<XorName> {
+        self.dysfunction_tracking
+            .get_dysfunctional_node_names()
+            .await
+    }
+
+    /// returns names that are relatively dysfunctional
+    pub(crate) async fn get_suspicious_node_names(&self) -> BTreeSet<XorName> {
+        self.dysfunction_tracking
+            .get_dysfunctional_node_names()
```
Yoga07, line 225: We seem to be getting dysfunctional nodes when checking for suspicion. Should be `.get_suspicious_node_names()`?

Also, the only difference between dys and sus are the ratios, but we seem to be calculating weighted scores serparately for them everytime. Would be better if we can both merge their methods and call on them with an enum.
Eg:
```
self.check_for_maliciousness(Severity::Dys);
self.check_for_maliciousness(Severity::Sus);
```

dys_function/src/operations.rs
```diff
@@ -0,0 +1,252 @@
...
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use crate::{get_mean_of, DysfunctionDetection, NodeIdentifier, OperationId};
+use std::collections::BTreeMap;
+use xor_name::XorName;
+
+impl DysfunctionDetection {
+    /// Inserts a pending_operation, and is deemed as such until we get the appropriate response from the node
+    /// Returns false if the operation already existed.
+    pub async fn add_a_pending_request_operation(
```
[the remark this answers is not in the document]
joshuef: Hmm, i think they are penalized by the op handing in the tracker, no? I think kperhaps it was removed even if we had bad data before, but now it should only be removed if it's good (i think)

dys_function/src/operations.rs
joshuef: it's "the node's" count. Aiming for generic across the counters, but as @davidrusu points out, we can probs do this better.

sn_dysfunction/src/network_knowledge.rs
```diff
@@ -0,0 +1,214 @@
...
+impl DysfunctionDetection {
+    /// Track an issue with a node's network knowledge
+    pub async fn track_knowledge_issue(&self, node_id: NodeIdentifier) -> Result<()> {
+        // initial entry setup if non existent
+        let mut entry = self.communication_issues.entry(node_id).or_default();
+
+        let mut queue = entry.value_mut().write().await;
+        queue.push_back(
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)?
```
Yoga07, line 21: Can we use `Instant` instead of calculating duration?

sn/src/node/core/connectivity.rs
```diff
@@ -29,16 +29,11 @@ impl Node {
             return Ok(vec![]);
         }
 
-        let mut cmds = self.propose_offline(name).await?;
-        cmds.push(Cmd::StartConnectivityTest(name));
+        self.log_comm_issue(name).await.map_err(Error::from)?;
```
davidrusu, line 32:
```suggestion
        self.log_comm_issue(name).await?;
```

davidrusu (approved):
👍 looks good

sn_dysfunction/src/network_knowledge.rs
Yoga07: To be followed up in an upcoming PR

sn/src/node/core/messaging/handling/service_msgs.rs
```diff
@@ -103,55 +152,16 @@ impl Node {
             return Ok(cmds);
         };
 
-        // Clear expired queries from the cache.
-        self.pending_data_queries.remove_expired().await;
+        // if there's anything sus going on at this point, lets trigger some activity
+        let sus_cmds = self.notify_about_newly_suspect_nodes().await?;
```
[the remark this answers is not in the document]
Yoga07: To be followed up in an upcoming PR

Yoga07 (approved):

Issue #1057, opened 2022-03-09
Allow valid AE updates to be processed at nodes, even if sent to an unknown (new but valid, child PK)


joshuef:
As per: 

https://maidsafe.discourse.team/t/whether-prefix-map-shall-hold-parent-when-only-know-of-one-child/522/9

If we get a message for an unkown PK, but it's an AE-Update with a valid chain _beyond_ what we know. We should accept it and update.

With this, we should then be able to remove Parent keys from the prefixmap even if we only know _one_ child key.

(We then also need to test that the prefixmap returns `siblingB` even if `siblingA` is closest (but unkown) (as that's the fastest path to the correct prefix)

PR #1060, opened 2022-03-14
feat: prune parent whenever a child inserted

maqi:
Close https://github.com/maidsafe/safe_network/issues/1057

### DIFF
--- a/sn/src/types/prefix_map/mod.rs
+++ b/sn/src/types/prefix_map/mod.rs
@@ -55,8 +55,7 @@ impl NetworkPrefixMap {
     }
 
     /// Inserts new entry into the map. Replaces previous entry at the same prefix.
-    /// Removes those ancestors of the inserted prefix that are now fully covered by their
-    /// descendants.
+    /// Removes those ancestors of the inserted prefix.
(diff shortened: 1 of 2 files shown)


sn/src/types/prefix_map/mod.rs
```diff
@@ -445,18 +440,17 @@ mod tests {
...
-        assert!(map.insert(sap0.clone()));
+        assert!(map.insert(sap0));
 
-        // Insert the first sibling. Parent remain in the map.
+        // Insert the first sibling. Parent get pruned in the map.
         let sap00 = gen_section_auth(p00)?;
         assert!(map.insert(sap00.clone()));
 
         assert_eq!(map.get(&p00), Some(sap00.value.clone()));
         assert_eq!(map.get(&p01), None);
```
Yoga07, line 450: Would it be better to assert that when we use the `section_by_name(p01)` method we get `p00`?

maqi: this unit test is mainly confirming the entries using the `get` method.
the functionality of `section_by_name` is confirmed by other tests, mainly `get_match` test.
And this scenario is really covered within the `get_match` test.

PR #1061, opened 2022-03-14
wip: membership integration

davidrusu:


sn/src/messaging/system/join.rs
```diff
@@ -21,8 +21,6 @@ pub struct JoinRequest {
     pub section_key: BlsPublicKey,
     /// Proof of the resource proofing.
     pub resource_proof_response: Option<ResourceProofResponse>,
-    /// Aggregated approval from the Elders
-    pub aggregated: Option<SectionAuth<NodeState>>,
```
davidrusu, line 25: `JoinRequest` are no longer aggregated by the joining node, instead they are used to initiate a Membership consensus round.

sn/src/messaging/system/join.rs
```diff
@@ -71,21 +69,6 @@ pub enum JoinResponse {
...
-        sig_share: SigShare,
-        /// Current `SectionAuthorityProvider` of the section.
-        section_auth: SectionAuthorityProvider,
-        /// Section signature over the `SectionAuthorityProvider`.
-        section_signed: KeyedSig,
-        /// Section chain of the current section
-        section_chain: SecuredLinkedList,
-        /// Signed NodeState of section members
-        members: SectionPeers,
-    },
```
davidrusu, line 88: ApprovalShare's are no longer aggregated client side anymore, instead signatures are aggregated during Membership consensus.

sn/src/node/api/cmds.rs
```diff
@@ -82,12 +82,6 @@ pub(crate) enum Cmd {
     /// Schedule a timeout after the given duration. When the timeout expires, a `HandleTimeout`
     /// cmd is raised. The token is used to identify the timeout.
     ScheduleTimeout { duration: Duration, token: u64 },
-    /// Test peer's connectivity
-    SendAcceptedOnlineShare {
-        peer: Peer,
-        // Previous name if relocated.
-        previous_name: Option<XorName>,
-    },
```
davidrusu, line 90: The `SendAcceptedOnlineShare` command was used to send the `JoinRequest::ApprovalShare` to the joining node, this is no-longer needed since we are not aggregating approval's client side anymore.

sn/src/node/core/bootstrap/join.rs
```diff
@@ -208,114 +199,6 @@ impl<'a> Join<'a> {
 
                     return Ok((self.node, network_knowledge));
                 }
-                JoinResponse::ApprovalShare {
```
davidrusu, line 211: We don't use the `JoinResponse::ApprovalShare` message anymore, the `JoinResponse::Approval` will be sent once this nodes is approved to join, the handling logic above will complete the bootstrapping phase for us.

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -25,50 +25,54 @@ const FIRST_SECTION_MIN_ELDER_AGE: u8 = 90;
...
-            .acquire()
-            .await
-            .map_err(|_| Error::PermitAcquisitionFailed)?;
+            let node_state = NodeState {
+                name: peer.name(),
+                addr: peer.addr(),
+                state: MembershipState::Joined,
+                previous_name: None,
+            };
+            return self.propose_membership_join(node_state).await;
```
davidrusu, line 74: The joining node has produced a valid resource proof, we'll initiate a membership consensus run to accept this node into the section.

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -240,24 +244,14 @@ impl Node {
         join_request: JoinAsRelocatedRequest,
         known_keys: Vec<BlsPublicKey>,
     ) -> Result<Vec<Cmd>> {
-        let _permit = self
-            .current_joins_semaphore
-            .acquire()
-            .await
-            .map_err(|_| Error::PermitAcquisitionFailed)?;
```
davidrusu, line 247: `Node::current_joins_semaphore` has been supplanted by Membership, `sn_consensus` will resolve multiple concurrent join requests into a coherent decision made across all elders.

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -267,61 +261,49 @@ impl Node {
...
-                    peer
-                );
+            if !join_request.relocate_proof.self_verify() {
+                debug!("Ignoring JoinAsRelocatedRequest from {peer} - invalid sig.");
+                return Ok(vec![]);
+            }
+            if !known_keys.contains(&join_request.relocate_proof.sig.public_key) {
+                debug!("Ignoring JoinAsRelocatedRequest from {peer} - untrusted src.");
                 return Ok(vec![]);
             }
```
davidrusu, line 287: This logic is equivalent, I just found it easier to follow when I broke up the two validations. (also, logs will be clearer as to why this join was rejected)

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -267,61 +261,49 @@ impl Node {
...
-                "Ignoring JoinAsRelocatedRequest from {} - invalid node name signature.",
-                peer
-            );
+            debug!("Ignoring JoinAsRelocatedRequest from {peer} - invalid node name signature.");
             return Ok(vec![]);
         }
 
-        let prefix = self.network_knowledge.prefix().await;
-        if !prefix.matches(&relocate_details.dst) {
+        if !our_prefix.matches(&relocate_details.dst) {
```
davidrusu, line 303: We had read `our_prefix` earlier, we then re-read it here on L319.

sn/src/node/membership/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+
+        if gen == 0 {
+            return Ok(members);
+        }
+
+        for (history_gen, consensus) in self.history.iter() {
+            let decision = if let Some(decision) = consensus.decision.as_ref() {
+                decision
+            } else {
+                panic!(
```
oetyng, line 92: Not sure if this is a temp remnant from development, but convention we have: Always create a logic path instead of using panics.

sn/src/node/membership/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+
+use super::{recommended_section_size, split};
+use crate::messaging::system::{MembershipState, NodeState};
+
+type Generation = u64;
+
+#[derive(Debug, Clone)]
+pub(crate) struct Membership {
+    consensus: Consensus<NodeState>,
+    bootstrap_members: BTreeSet<NodeState>,
```
oetyng, line 19: So, one thing that was already the case before this PR, just wanted to point it out in case you see a swift way you'd like to address it (now or later):

`bootstrap_members` is a set of what is supposed to be nodes in the section at that time. However, `NodeState` allows for representing something quite different, like a set of all `Left`, which doesn't make sense.

To me, that means something is off in the design. It shouldn't be possible to represent invalid states like that.

Same thing comes up when we ask for members at a certain generation. What we expect to get back is a set of nodes who are in the section at that generation. But the `NodeState` allows for something nonsensical like a set of `Left` or `Relocated`. The logic makes it not happen - right now - but the types used are "unsafe" (as in: type safety).

sn/src/node/membership/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+    gen: Generation,
+    history: BTreeMap<Generation, Consensus<NodeState>>,
```
oetyng, line 21: Ok, I see. So we are basically event-sourcing here. We have an initial state, and then a log of changes (history), and we replay (hydrate) up to any point in time we chose.

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -267,72 +248,37 @@ impl Node {
...
         }
 
-        if self.network_knowledge.is_section_member(&peer.name()).await {
-            debug!(
-                "Ignoring JoinAsRelocatedRequest from {} - already member of our section.",
-                peer
-            );
-            return Ok(vec![]);
-        }
-
```
davidrusu, line 284: This validation was moved to `Membership::validate_node_state()`

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -267,72 +248,37 @@ impl Node {
...
-        let prefix = self.network_knowledge.prefix().await;
-        if !prefix.matches(&relocate_details.dst) {
-            debug!(
-                "Ignoring JoinAsRelocatedRequest from {} - destination {} doesn't match \
-                         our prefix {:?}.",
-                peer, relocate_details.dst, prefix
-            );
-            return Ok(vec![]);
-        }
```
davidrusu, line 327: This validation has been moved to `Membership::validate_node_state()`

sn/src/node/core/messaging/handling/join.rs
```diff
@@ -267,72 +248,37 @@ impl Node {
...
-
-        // Requires the node name matches the age.
-        let age = relocate_details.age;
-        if age != peer.age() {
-            debug!(
-                "Ignoring JoinAsRelocatedRequest from {} - relocation age ({}) doesn't match peer's age ({}).",
-                peer, age, peer.age(),
-            );
```
davidrusu, line 335: Also moved to `Membership::validate_node_state`

sn/src/node/core/messaging/handling/membership.rs
```diff
@@ -0,0 +1,63 @@
...
+impl Node {
+    pub(crate) async fn handle_membership_vote(
+        &self,
+        peer: Peer,
+        signed_vote: SignedVote<NodeState>,
+    ) -> Result<Vec<Cmd>> {
+        debug!("Received membership vote {:?} from {}", signed_vote, peer);
+
+        let cmds = if let Some(membership) = self.membership.write().await.as_mut() {
+            assert!(self.is_elder().await);
```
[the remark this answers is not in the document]
davidrusu: Removed the assert

sn/src/node/core/messaging/handling/membership.rs
```diff
@@ -0,0 +1,63 @@
...
+                    ]
+                }
+                Ok(VoteResponse::WaitingForMoreVotes) => vec![],
+                Err(e) => {
+                    error!("Error while processing vote {:?}", e);
+                    vec![]
+                }
+            }
+        } else {
+            assert!(self.is_not_elder().await);
```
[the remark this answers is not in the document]
davidrusu: Removed the assert

sn/src/messaging/system/mod.rs
```diff
@@ -91,6 +92,8 @@ pub enum SystemMsg {
     BackPressure(LoadReport),
     /// Send from a section to the node to be immediately relocated.
     Relocate(SectionAuth<NodeState>),
+    /// Membership Vote
+    Membership(SignedVote<NodeState>),
```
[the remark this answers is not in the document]
davidrusu: Yeah that could be better, renamed to MembershipVote

sn/src/node/membership/mod.rs
davidrusu: Yeah, fair point, I think that would require a change to the `NodeState` struct itself, perhaps removing the `MembershipState` enum from the node state entirely and rely on different sets i.e. `current_members`, `offline`, `relocated` to determine which state a node is in.

We can explore this in a future pr

sn/src/node/membership/mod.rs
davidrusu: removed the panic

sn_interface/src/messaging/system/agreement.rs
```diff
@@ -7,50 +7,109 @@
...
 pub struct DkgSessionId {
-    /// A hash of the peers and prefix of the specific session.
-    #[debug(with = "Self::fmt_hash")]
-    pub hash: Digest256,
+    /// Prefix of the session we are elder candidates for
+    pub prefix: Prefix,
+    /// Other Elders in this dkg session
+    pub elders: BTreeMap<XorName, SocketAddr>,
     /// The generation, as in the length of the section chain main branch.
     pub generation: u64,
```
davidrusu, line 34: rename to section_chain_length

sn/src/node/core/messaging/handling/agreement.rs
```diff
@@ -148,74 +146,13 @@ impl Node {
         node_state: NodeState,
         sig: KeyedSig,
     ) -> Result<Vec<Cmd>> {
```
davidrusu, line 148: handle_offline_agreement() is called after all elders "agree" to propose a node offline.

Previously we would act on this agreement immediately, now instead we propose a membership change

sn/src/node/core/messaging/handling/left.rs
```diff
@@ -0,0 +1,92 @@
...
+    types::log_markers::LogMarker,
+};
+
+use crate::node::{
+    api::cmds::Cmd,
+    core::{relocation::ChurnId, Node, Result},
+};
+
+impl Node {
+    pub(crate) async fn handle_node_left(
```
davidrusu, line 15: A lot of this code comes from `handle_offline_agreement()`. This now gets called after we reach agreement that a node has left the section.

sn/src/node/core/messaging/handling/mod.rs
```diff
@@ -631,26 +558,22 @@ impl Node {
                 )
                 .await
             }
-            SystemMsg::DkgStart {
-                session_id,
-                prefix,
-                elders,
-            } => {
+            SystemMsg::DkgStart(session_id) => {
```
davidrusu, line 561: We've moved `prefix` and `elders` into the `session_id` struct.

sn/src/node/membership/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+                .get(&gen)
+                .map(|(_, c)| c)
+                .ok_or(Error::BadGeneration {
+                    requested_gen: gen,
+                    gen: self.gen,
+                })
+        }
+    }
+
+    fn consensus_at_gen_mut(&mut self, gen: Generation) -> Result<&mut Consensus<NodeState>> {
```
grumbach, line 133: Should an issue for this? Being able to mutate history seems off.

sn/src/node/membership/mod.rs
```diff
@@ -0,0 +1,455 @@
...
+                    // joining node would be part of the `zeros` child section
+                    false => zeros.len() < split_section_size_cap,
+                }
+            }
+            None => false,
+        }
+    }
+
+    #[allow(dead_code)]
+    fn validate_node_age(&self, node_state: &NodeState) -> bool {
```
grumbach, line 351: Why don't we use this check? Do we want to enhance it before?

sn/src/node/core/messaging/handling/dkg.rs
```diff
@@ -43,41 +35,38 @@ impl Node {
         let section_auth = self.network_knowledge().authority_provider().await;
 
         let mut peers = vec![];
-        for (name, addr) in elders.into_iter() {
+        for peer in session_id.elder_peers() {
```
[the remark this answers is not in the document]
davidrusu: fixed in: https://github.com/maidsafe/safe_network/pull/1120/commits/f179a94467b4123f020087ecae6cca97a5486a22

sn/src/node/core/messaging/handling/left.rs
```diff
@@ -0,0 +1,92 @@
...
+        self.liveness_retain_only(
+            self.network_knowledge
+                .adults()
+                .await
+                .iter()
+                .map(|peer| peer.name())
+                .collect(),
+        )
+        .await?;
+        *self.joins_allowed.write().await = true;
```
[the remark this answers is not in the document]
davidrusu: This code is copied over from `handle_offline_agreement()`, previously we would act on the `Offline` agreement immediately, now we just use it to trigger a membership change which on decision calls into this method.

sn/src/node/membership/mod.rs
davidrusu: done: https://github.com/maidsafe/safe_network/issues/1121

sn/src/node/membership/mod.rs
davidrusu: This was a new check that was not previously being done, it doesn't work, (it fails to validate very often). I left it in until we could better understand why node age was not behaving the way I expected.

PR #1065, opened 2022-03-17
Msg flow control improvements

oetyng:
- Refactors `back_pressure` module to use msgs per s, so that all peers know the number of msgs per s the other peers want to receive.
- Adds individual send queues for peers, with send rates based on their desired rate (`back_pressure` module calcs).
- Adds true prioritization of msgs, so that important infrastructure msgs are always sent before less important client service msgs.

`store_and_read_5mb_from_many_clients` with 50 client readers (currently 25 in the tests) get following results locally:

```
Time: 30 s
CPU: ~40 %, (very briefly above 50%)
Mem: ~60 mb / Elder, (very briefly up to 125 mb)
```
 
compared to the results on `main`:

```
Time: 704 s
CPU: 100 %, all the time
Mem: ~2 gig / Elder, all the time
```


sn/src/node/core/comm/back_pressure/load_monitoring.rs
```diff
@@ -0,0 +1,192 @@
...
+impl MsgCount {
+    pub(crate) fn new() -> Self {
+        Self {
+            running: Arc::new(AtomicUsize::new(0)),
+            snapshot: Arc::new(AtomicUsize::new(0)),
+        }
+    }
+
+    pub(crate) fn read(&self) -> usize {
+        self.snapshot.load(ORDER)
```
Yoga07, line 181: Can `snapshot` be outdated during this `read`?
Would it be better to update `self.snapshot` on here rather than having the user to call `snapshot()` manually?

sn/src/node/core/comm/back_pressure/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+            .unwrap_or(DEFAULT_MSGS_PER_S_AND_PEER)
+    }
+
+    /// Remove regulation for specific nodes when we don't need them anymore (e.g. they left).
+    pub(crate) async fn remove(&self, addr: SocketAddr) {
+        let _prev = self.reports.write().await.remove(&addr);
+    }
+
+    /// Sets requested msgs_per_s for a node, when it errored back saying it was strained.
+    pub(crate) async fn set(&self, addr: SocketAddr, msgs_per_s: usize) {
```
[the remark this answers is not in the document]
joshuef: > Just pondering, does it add a new attack vector to the network? How can we verify that the node genuinely asked for this/an impersonater asked for this limiting?

The messages should be signed and verified as normal no?

sn/src/node/core/comm/back_pressure/load_monitoring.rs
```diff
@@ -0,0 +1,192 @@
...
+                } else if period == SAMPLING_INTERVAL_FIFTEEN {
+                    load.fifteen / msg_count as f64
+                } else {
+                    MAX_CPU_LOAD / INITIAL_MSGS_PER_S as f64
+                };
+
+                let max_msgs_to_handle = MAX_CPU_LOAD / load_per_msg;
+                if let Some(counter) = self.msgs_per_s.get(&period) {
+                    counter.store(
+                        usize::max(INITIAL_MSGS_PER_S, max_msgs_to_handle as usize),
```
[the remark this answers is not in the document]
oetyng: What you suggest would not be correct.

Msgs per second is still msgs per s, regardless of if it was mesaured during 1 millisecond or 1000 years.

We are setting a default value here, which is later replaced with actual measurements.

Using the _same_ default value for all intervals, is like having had a constant load over all intervals.

What we want in the end is a default value for what the node reports to others. As we measure an avg of the three intervals, the same default for the three periods, results in the same default for the node reported value.

There's really no reason to use different values there.

sn/src/node/core/comm/back_pressure/mod.rs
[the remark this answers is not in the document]
oetyng: As joshuef says, signed and verified like any other msg.

sn/src/node/core/comm/back_pressure/load_monitoring.rs
oetyng: No, we can't snapshot on reading, because snapshotting is to be made only at the specific intervals.

Reading is made at other times as well, specifically when calling `fn msgs_per_s`.

If we snapshot at read, then the next read doesn't get the value of "number of msgs during the specified interval", but instead "number of msgs since last read", which is an arbitrary period and mostly very distinct from any of the specified periods.

So, in other words: no, that would change what it does :)

***

The value being "stale" is not an issue. The whole thing here is a very fuzzy value. We're measuring an avg of loads over three time intervals: 1, 5 and 15 minutes.
Also, it's the value we have, since its a measurement of the last completed interval - the current interval being measured is not yet finished.

So, we have background tasks which update values at specific intervals, and then consumer of the code reading (the avg of) those values at random times. We do not want the values to be updated when the caller reads them, only when the defined intervals pass.

(On a more general note, I am very deliberately avoiding mixing writes and reads. 
I.e. a read should not mutate state.)

sn/src/node/core/comm/link.rs
```diff
@@ -0,0 +1,366 @@
+// Copyright 2022 MaidSafe.net limited.
```
joshuef, line 1: How does this compare to the `types/Link`? can they be unified in some fashion?

sn/src/node/core/comm/back_pressure/load_monitoring.rs
```diff
@@ -0,0 +1,192 @@
...
+const ONE_MINUTE_AS_SECONDS: u64 = 60;
+
+const SAMPLING_INTERVAL_ONE: Duration = Duration::from_secs(ONE_MINUTE_AS_SECONDS);
+const SAMPLING_INTERVAL_FIVE: Duration = Duration::from_secs(5 * ONE_MINUTE_AS_SECONDS);
+const SAMPLING_INTERVAL_FIFTEEN: Duration = Duration::from_secs(15 * ONE_MINUTE_AS_SECONDS);
+
+const INITIAL_MSGS_PER_MINUTE: usize = ONE_MINUTE_AS_SECONDS as usize * INITIAL_MSGS_PER_S;
+
+const MAX_CPU_LOAD: f64 = 0.8;
+const ORDER: Ordering = Ordering::SeqCst;
```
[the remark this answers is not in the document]
oetyng: I think it makes it clear and easy to see that there is just one of the variants to be used. And since it's unimportant knowledge at the places where the const is used, it also cleans that up a bit, removes unnecessary information.

sn/src/node/core/comm/back_pressure/load_monitoring.rs
oetyng: No, we can't snapshot on reading, because snapshotting is to be made only at the specific intervals.

Reading is made at other times as well, specifically when calling fn msgs_per_s.

If we snapshot at read, then the next read doesn't get the value of "number of msgs during the specified interval", but instead "number of msgs since last read", which is an arbitrary period and mostly very distinct from any of the specified periods.

So, in other words: no, that would change what it does :)

The value being "stale" is not an issue. The whole thing here is a very fuzzy value. We're measuring an avg of loads over three time intervals: 1, 5 and 15 minutes.
Also, it's the value we have, since its a measurement of the last completed interval - the current interval being measured is not yet finished.

So, we have background tasks which update values at specific intervals, and then consumer of the code reading (the avg of) those values at random times. We do not want the values to be updated when the caller reads them, only when the defined intervals pass.

(On a more general note, I am very deliberately avoiding mixing writes and reads.
I.e. a read should not mutate state.)

sn/src/node/core/comm/back_pressure/mod.rs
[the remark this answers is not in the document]
oetyng: As joshuef says, signed and verified like any other msg.

sn/src/node/core/comm/back_pressure/load_monitoring.rs
```diff
@@ -0,0 +1,192 @@
...
+            let _instant = interval.tick().await;
+            if period == SAMPLING_INTERVAL_ONE {
+                {
+                    self.system.write().await.refresh_cpu();
+                }
+                *self.load_sample.write().await =
+                    normalize(self.system.read().await.load_average());
+            } else {
+                // allow the one min interval to come first and wait for it to finish
+                tokio::time::sleep(Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
Yoga07: Right. That clears it up 👍

sn/src/node/core/comm/back_pressure/load_monitoring.rs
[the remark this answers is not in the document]
Yoga07: Yep, that makes sense. Was a noob Q :)

sn/src/node/core/comm/back_pressure/mod.rs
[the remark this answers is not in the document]
Yoga07: >As joshuef says, signed and verified like any other msg.

Right, that removes the impersonation problem. 

Can bad nodes then deliberately set this limit for themselves? Slowing the performance of the section?

sn/src/node/core/comm/back_pressure/load_monitoring.rs
Yoga07: >We do not want the values to be updated when the caller reads them, only when the defined intervals pass.

Should probably add this as a comment there :) Clears it up 👍

sn/src/node/core/comm/back_pressure/mod.rs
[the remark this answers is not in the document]
joshuef: > Can bad nodes then deliberately set this limit for themselves? Slowing the performance of the section?
Hmmm. This is something we'll need to factor into dysfunction I think. If a node constantly claims to be overwhelmed it's of limited use to use and should eventually be punished for its slowness (i think).

Not something for this PR to to keep in mind down the line, I'd say. (I've taken a note 👍 )

sn/src/node/core/comm/link.rs
joshuef: Is it just the `listener` addition that's changed here? Otherwise the files look exactly the same.

If that's the casee I think we can just add that as an `Option` to the `types/Link` no? That'd be cleaner and remove 350 lines of code there I think. And keep it all unified from the get go.

joshuef: That said, if you're wanting to keep changes to this minimal, I'm fine with this being done in a follow up PR if you can jump on that once this is in.

sn/src/node/core/comm/back_pressure/load_monitoring.rs
oetyng: It is never really a good thing to make reads mutate state. 
So I don't think that specific sentence is what should go in as a comment in `fn read` (because we would never want to update on read, not just in this case).

sn/src/node/core/comm/link.rs
oetyng: It's  a whole lot more than just the listener param that changed. So the files are definitely not exactly the same.

The whole way that we start listening for msgs on connections.
The send fns take other params.

The thing is that I see all of the client/node shared comms code, needing various refactors and reorganisation to make sense.

This was not meant to stay as separate Link types.

So I just deferred that completely, since the changes affects the client code, and to use the same file I would need to refactor the client code as well.

Better make a full effort there in a separate PR.

joshuef: Aha, okay cool. Makes sense 👍

joshuef (approved):

joshuef (approved):

PR #1099, opened 2022-04-07
docs: adding a new crate to release process

jacderida:
I think this covers everything in the process. It's based on a review of df66875627aa41d06b7613085f05a97187c7175d, which was the addition of the `sn_dysfunction` crate.


dirvine (approved):
Perfecto

PR #1097, opened 2022-04-06
feat(api): adding first set of basic wallet APIs and CLI commands

bochaco:
This adds a first basic set of wallet APIs and CLI commands:
- Create a Wallet on Safe, storing it using private Register (wallet entries) + private Chunks (serialised Dbc's)
- Deposit DBCs into an existing Wallet (only bearer DBC's currently supported)
- Check the total balance of an existing Wallet, i.e. the balance of all DBC's deposited in the Wallet

The CLI `wallet deposit` command is currently compatible with the DBC's generated and printed out by the [mint-repl](https://github.com/maidsafe/sn_dbc#mint-repl-example) and [DBC Playground](https://github.com/maidsafe/sn_dbc_examples) output, once a DBC is generated by those example apps, it can be deposited in a Wallet and check its balance, e.g. the following commands deposit the genesis DBC into a Wallet on Safe and check its total balance:
```
$ safe wallet create
Wallet created at: "safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y"

$ safe wallet balance safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y
Wallet at "safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y" has a total balance of 0.000000000 safecoins

$ safe wallet deposit --name my-first-dbc safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y 550ec207952fcb671353628d68a3e6860ba86...
Spendable DBC deposited with name 'my-first-dbc' in Wallet located at "safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y"

$ safe wallet balance safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y
Wallet at "safe://hyryynyengccd6t6zrq4zgq89bo8sec976kguhhp38a6gajxa31ipz96pxjjy84y" has a total balance of 18446744073.709551615 safecoins
```

As next steps to work on in separate PRs:
- the sn-api unit tests can be improved if test DBCs can be generated with different amounts (currently a hard-coded one is used)
- Add some sn_cli tests
- Review how exactly we want to serialise and store the Wallets, currently using Private Register+Chunks
[...]

PR #1105, opened 2022-04-13
feat(api): first and basic implementation of Wallet reissue API and related CLI cmds

bochaco:
This is a follow up PR (2nd PR) to PR #1097 , introducing first and basic implementation of Wallet reissue API and CLI cmd.
- Reissue an output DBC from a Wallet (using [sn_dbc::TransactionBuilder](https://github.com/maidsafe/sn_dbc/blob/d275810f33376acc34cf7fd3f8045f585b264791/src/builder.rs#L36)) for a provided amount
- Reissue the change DBC which is automatically stored in the source Wallet the reissue was made from
- Spent DBCs are automatically soft-removed from the source Wallet (Multimap)
- Reissued DBCs are all bearer at this instance

The CLI `wallet reissue` and `wallet deposit` commands are currently compatible with the DBC's generated and printed out by the [mint-repl](https://github.com/maidsafe/sn_dbc#mint-repl-example) and [DBC Playground](https://github.com/maidsafe/sn_dbc_examples) output, once a DBC is generated by those example apps, it can be deposited in a Wallet and check its balance, e.g. the following commands deposit the genesis DBC into a Wallet on Safe, check its total balance, and reissue a DBC from the Wallet for 3.77 safecoins:
```
$ safe wallet create
Wallet created at: "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y"

$ safe wallet balance safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Wallet at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y" has a total balance of 0.000000000 safecoins

$ safe wallet deposit --name my-first-dbc safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y c4a1953dfa419234b8ab9a...
Spendable DBC deposited with name 'my-first-dbc' in Wallet located at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y"

$ safe wallet balance safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Wallet at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y" has a total balance of 18446744073.709551615 safecoins

$ safe wallet reissue 3.77 --from safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Success. Reissued DBC with 3.77 safecoins:
-------- DBC DATA --------
0000000000000000d3dfd5b56aeea1b906c82.......bb11e30f382dda25700000000
--------------------------

$ safe wallet balance safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Wallet at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y" has a total balance of 18446744069.939551615 safecoins
```

As next steps to work on in separate PRs:
- Verification of generated Tx and spentproofs upon reissuing DBCs (depends on spentbook implementation and messaging to be available)
- Log input DBCs as spent on the network's spentbook upon reissuing DBCs (depends on spentbook implementation and messaging to be available)
- Verification of spentproofs upon depositing DBCs in Wallets (depends on spentbook implementation and messaging to be available)
- Use of input decoys in the transactions upon reissuing DBCs
- Support for non-bearer DBCs
- Review how exactly we want to serialise and store the Wallets, currently using Private Register+Chunks
- The sn-api unit tests can be improved if test DBCs can be generated with different amounts. Generate test DBCs from an amount and SK instead of deserialising a hard-coded serialised DBC. Hard-coded serialised DBCs are currently generated with sn_dbc mint-repl example.

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -6,11 +6,15 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+pub use sn_dbc::Dbc;
+
+use super::helpers::parse_tokens_amount;
+use super::register::EntryHash;
 use crate::safeurl::{ContentType, SafeUrl, XorUrl};
 use crate::{Error, Result, Safe};
 use bytes::Bytes;
 use log::{debug, warn};
-pub use sn_dbc::Dbc;
+use sn_dbc::{rng, Owner, OwnerOnce, TransactionBuilder};
 use sn_interface::types::Token;
 use std::collections::{BTreeMap, BTreeSet};
 
@@ -19,7 +23,7 @@ const WALLET_TYPE_TAG: u64 = 1_000;
 
 /// Set of spendable DBC's mapped to their friendly name
 /// as defined/chosen by the user when depositing DBC's into a Wallet.
-pub type WalletSpendableDbcs = BTreeMap<String, Dbc>;
+pub type WalletSpendableDbcs = BTreeMap<String, (Dbc, EntryHash)>;
 
 impl Safe {
     /// Create an empty Wallet on Safe and return its XOR-URL
@@ -48,28 +52,18 @@ impl Safe {
             return Err(Error::InvalidInput("Only bearer DBC's are supported at this point by the Wallet. Please deposit a bearer DBC's.".to_string()));
         }
 
+        // TODO: check the input DBCs were spent and all other sort of verifications,
+        // perhaps all optional, and we may want a separate API to also do these verifications
+        // for the user to perform them without depositing the DBC into a Wallet.
+
+        let safeurl = self.parse_and_resolve_url(wallet_url).await?;
+
         let spendable_name = match spendable_name {
             Some(name) => name.to_string(),
             None => hex::encode(dbc.hash()),
         };
 
-        let dbc_bytes = Bytes::from(rmp_serde::to_vec_named(&dbc).map_err(|err| {
-            Error::Serialisation(format!(
-                "Failed to serialise DBC to insert it into the Wallet: {:?}",
-                err
-            ))
-        })?);
-
-        let safeurl = self.parse_and_resolve_url(wallet_url).await?.to_string();
-        let dbc_xorurl = self.store_private_bytes(dbc_bytes, None).await?;
-
-        let entry = (spendable_name.clone().into_bytes(), dbc_xorurl.into_bytes());
-        let _entry_hash = self
-            .multimap_insert(
-                &safeurl,
-                entry,
-                BTreeSet::default(), // TODO: provide root to replace if DBC already exists ??
-            )
+        self.insert_dbc_into_wallet(&safeurl, dbc, spendable_name.clone())
             .await?;
 
         debug!(
@@ -82,8 +76,8 @@ impl Safe {
 
     /// Fetch a Wallet from a Url performing all type of URL resolution required.
     /// Return the set of spendable DBCs found in the Wallet.
-    pub async fn wallet_get(&self, url: &str) -> Result<WalletSpendableDbcs> {
-        let safeurl = self.parse_and_resolve_url(url).await?;
+    pub async fn wallet_get(&self, wallet_url: &str) -> Result<WalletSpendableDbcs> {
+        let safeurl = self.parse_and_resolve_url(wallet_url).await?;
         debug!("Wallet URL was parsed and resolved to: {}", safeurl);
         self.fetch_wallet(&safeurl).await
     }
@@ -113,7 +107,7 @@ impl Safe {
         };
 
         let mut balances = WalletSpendableDbcs::default();
-        for (_entry_hash, (key, value)) in entries.iter() {
+        for (entry_hash, (key, value)) in entries.iter() {
             let xorurl_str = std::str::from_utf8(value)?;
             let dbc_xorurl = SafeUrl::from_xorurl(xorurl_str)?;
             let dbc_bytes = self.fetch_data(&dbc_xorurl, None).await?;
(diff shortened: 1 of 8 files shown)


sn_api/src/app/wallet.rs
```diff
@@ -169,23 +163,183 @@ impl Safe {
...
+                Token::from_nano(total_input_amount).to_string(),
+            ));
+        }
+
+        // We can now reissue the output DBCs
+        let (output_dbc, change_dbc) = self
+            .reissue_dbcs(input_dbcs_to_spend, output_amount, change_amount)
+            .await?;
+
+        if let Some(change_dbc) = change_dbc {
```
jacderida, line 237: I see in `test_wallet_reissue` we check for `change-dbc`. Is the `test_wallet_deposit_reissued_dbc` case covering when when `change_dbc` would be `None`?

sn_api/src/app/wallet.rs
```diff
@@ -362,6 +516,91 @@ mod tests {
...
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_wallet_not_enough_balance() -> Result<()> {
+        let safe = new_safe_instance().await?;
+        let wallet_xorurl = safe.wallet_create().await?;
+
+        match safe.wallet_reissue(&wallet_xorurl, "2.55").await {
```
jacderida, line 561: Minor point, but unless we consider the zero balance to be some special case (in which case it would be arguable that it should be in its own test case) I think checking one `wallet_reissue` call is fine.

sn_cli/tests/cli_wallet.rs
```diff
@@ -76,3 +76,93 @@ fn calling_safe_wallet_balance() -> Result<()> {
...
+        ["wallet", "reissue", "7.15", "--from", &wallet_xorurl],
+        Some(0),
+    )?
+    .assert()
+    .stdout(predicate::str::contains(
+        "Success. Reissued DBC with 7.15 safecoins",
+    ))
+    .success();
+
+    safe_cmd(["wallet", "balance", &wallet_xorurl], Some(0))?
```
jacderida, line 107: Since we have API tests covering the change of balance, I think it would be reasonable for the CLI test to assume that this is done correctly, and we wouldn't need the additional command and assertion.

sn_cli/tests/cli_wallet.rs
```diff
@@ -76,3 +76,93 @@ fn calling_safe_wallet_balance() -> Result<()> {
...
+            "deposit",
+            "--name",
+            "reissued-dbc",
+            &wallet_xorurl,
+            &reissued_dbc,
+        ],
+        Some(0),
+    )?;
+
+    safe_cmd(["wallet", "balance", &wallet_xorurl], Some(0))?
```
jacderida, line 159: Similarly here, I would say a balance command check is not necessary for the CLI test.

sn_api/src/app/wallet.rs
```diff
@@ -169,23 +163,183 @@ impl Safe {
...
+
+        let entry = (spendable_name.into_bytes(), dbc_xorurl.into_bytes());
+        let _entry_hash = self
+            .multimap_insert(&safeurl.to_string(), entry, BTreeSet::default())
+            .await?;
+
+        Ok(())
+    }
+
+    // Private helper to reissue DBCs using the sn_dbc API,
```
jacderida, line 277: Minor style issue: should we document functions using doc comments? Even if it's private, I think it's preferable to use those.

sn_api/src/app/wallet.rs
```diff
@@ -169,23 +163,183 @@ impl Safe {
...
+    pub async fn wallet_reissue(&self, wallet_url: &str, amount: &str) -> Result<Dbc> {
+        debug!(
+            "Reissuing Bearer-DBCs from Wallet at {}, for an amount of {} tokens",
+            wallet_url, amount
+        );
+
+        // Parse and validate the output amount is valid
+        let output_amount = parse_tokens_amount(amount)?;
+        if output_amount.as_nano() == 0 {
+            return Err(Error::InvalidAmount(
```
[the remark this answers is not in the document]
bochaco: correct, i'll add a test

sn_api/src/app/wallet.rs
bochaco: yeah, i think we need a testcase, at least for how I see it, the reason I'm restricting it is to save the user with unnecessary calls to the network and even potential bugs for the scenario which has no output or outcome, if the amount is 0 then there shouldn't be any reissue and the Wallet shall be left untouched. If we agree on this, I can add a test case.

sn_api/src/app/wallet.rs
bochaco: i think we are missing that scenario of no-change, i'll add a testcase.

sn_api/src/app/wallet.rs
jacderida: Ah I see. It seems like this is a special scenario that would be worth its own case.

sn_cli/tests/cli_wallet.rs
bochaco: I think we can keep this verification since there is a deserialisation and seriaisation happening at CLI level which is not covered by sn_api layer. When we reissue a DBC from sn_api, we obtain an instance of the Dbc object, the CLI then serialises it to present it to the stdout/user, then when the user copy and paste it in the `wallet deposit` command the CLI deserialises it and it passes a Dbc instance down to the sn_api::wallet_deposit API. Does it make sense? do you agree?

jacderida (approved):

PR #1106, opened 2022-04-14
sn_interface release changes

joshuef:

### DIFF
--- a/resources/scripts/bump_version.sh
+++ b/resources/scripts/bump_version.sh
@@ -105,10 +105,10 @@ function generate_new_commit_message() {
   commit_message="chore(release): "
 
   if [[ $sn_interface_has_changes == true ]]; then
-    commit_message="${commit_message}sn_interface-${sn_interface}/"
+    commit_message="${commit_message}sn_interface-${sn_interface_version}/"
   fi
   if [[ $sn_dysfunction_has_changes == true ]]; then
-    commit_message="${commit_message}sn_dysfunction-${sn_dysfunction}/"
+    commit_message="${commit_message}sn_dysfunction-${sn_dysfunction_version}/"
   fi
   if [[ $safe_network_has_changes == true ]]; then
     commit_message="${commit_message}safe_network-${sn_version}/"
--- a/resources/scripts/bump_version.sh
+++ b/resources/scripts/bump_version.sh
@@ -6,6 +6,7 @@ sn_version=""
 sn_api_version=""
 sn_cli_version=""
 sn_dysfunction_has_changes="false"
+sn_interface_has_changes="false"
 safe_network_has_changes="false"
 sn_api_has_changes="false"
 sn_cli_has_changes="false"
@@ -19,7 +20,7 @@ function perform_smart_release_dry_run() {
     --no-changelog-preview \
     --allow-fully-generated-changelogs \
     --no-changelog-github-release \
-    "sn_dysfunction" "safe_network" "sn_api" "sn_cli" 2>&1)
+    "sn_interface" "sn_dysfunction" "safe_network" "sn_api" "sn_cli" 2>&1)
   echo "Dry run output for smart-release:"
   echo $dry_run_output
 }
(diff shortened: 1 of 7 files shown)


resources/scripts/bump_version.sh
```diff
@@ -87,13 +95,18 @@ function generate_version_bump_commit() {
...
+    grep "^version" < sn_interface/Cargo.toml | head -n 1 | awk '{ print $3 }' | sed 's/\"//g')
   sn_dysfunction_version=$( \
     grep "^version" < sn_dysfunction/Cargo.toml | head -n 1 | awk '{ print $3 }' | sed 's/\"//g')
   sn_version=$(grep "^version" < sn/Cargo.toml | head -n 1 | awk '{ print $3 }' | sed 's/\"//g')
   sn_api_version=$(grep "^version" < sn_api/Cargo.toml | head -n 1 | awk '{ print $3 }' | sed 's/\"//g')
   sn_cli_version=$(grep "^version" < sn_cli/Cargo.toml | head -n 1 | awk '{ print $3 }' | sed 's/\"//g')
   commit_message="chore(release): "
 
+  if [[ $sn_interface_has_changes == true ]]; then
+    commit_message="${commit_message}sn_interface-${sn_interface}/"
```
jacderida, line 108: This should be `${sn_interface_version}`. Just noticed there's a mistake in the dysfunction one too! It should be `${sn_dysfunction_version}` and not `${sn_dysfunction}`.

resources/scripts/insert_changelog_entry.py
```diff
@@ -26,6 +26,9 @@ def insert_changelog_entry(entry, pattern):
         file.write(release_description)
 
 def main(sn_dysfunction_version, sn_version, sn_api_version, sn_cli_version):
```
jacderida, line 28: Are you missing the addition of an `sn_interface_version` argument to the `main` function there?

resources/scripts/output_versioning_info.sh
```diff
@@ -2,12 +2,14 @@
 
 function build_release_name() {
   gh_release_name="Safe Node Dysfunction v$sn_dysfunction_version/"
+  gh_release_name="Safe Network Interface v$sn_interface_version/"
```
jacderida, line 5: I think this needs to be `gh_release_name="${gh_release_name}Safe Network Interface v$sn_interface_version/"`.

Perhaps also the release name is getting a bit unwieldy now? Possible that we should maybe move the versions into the body of the GH release. However, don't need to do that on this branch.

resources/scripts/output_versioning_info.sh
```diff
@@ -2,12 +2,14 @@
...
   gh_release_name="${gh_release_name}Safe Network v$sn_version/"
   gh_release_name="${gh_release_name}Safe API v$sn_api_version/"
   gh_release_name="${gh_release_name}Safe CLI v$sn_cli_version"
 }
 
 function build_release_tag_name() {
+  gh_release_tag_name="$sn_interface_version-"
   gh_release_tag_name="$sn_dysfunction_version-"
```
jacderida, line 13: This would now need to change to `gh_release_tag_name=${gh_release_tag_name}$sn_dysfunction_version"`.

jacderida (commented):
Was a `cargo publish --dry-run` working OK for the crate?

resources/scripts/output_versioning_info.sh
joshuef: for both this and dysf?

resources/scripts/output_versioning_info.sh
jacderida: That's the weird Bash way of appending to a string. If you don't change it, the first string will be overwritten by the second one, but we actually want to append it.

resources/scripts/output_versioning_info.sh
jacderida: Just for the Dysfunction one. We're also trying to append to the string here.

resources/scripts/output_versioning_info.sh
joshuef: ah of course! 🤦

PR #1109, opened 2022-04-14
feat: compare against all nodes in section

jacderida:
When calculating scores, we compare against the average of all the nodes in the section, rather than
the 'neighbours' of the node. As a consequence, the `DysfunctionDetection` struct becomes simpler,
as we don't need to keep track of what nodes are 'closer' to others as the set of nodes changes.

In the course of this change, the Dysfunction API was updated.

First, the `ScoreType` enum was renamed `IssueType`, which now details the three issue types, rather
than have timed versus operational. Secondly, one `track_issue` function merges three
`track_x_issue` functions for adding issues to the tracker.

Both these decisions were influenced by property testing: there were properties that should have
been true for each issue type, and it was easier to write tests for those properties if they all
used a single API.

It's worth noting, the `track_issue` function provides an `Option` argument for supplying an
`op_id`. This value only applies when a `PendingRequestOperation` type is used. At first, the
pending request entry was declared as `PendingRequestOperation([u8; 32])`, which makes sense
initially. For adding the issue, you just pass the op ID along with the issue type. However, the
problem comes when you later want to select the issues of this type. For example:
```
let _ = op_scores.insert(
    *node,
    self.calculate_node_score(
        node,
        adults.clone(),
        IssueType::PendingRequestOperation,
    )
    .await,
);
```

If you have the `op_id` parameter on the enum entry, the code becomes very clunky:
```
let _ = op_scores.insert(
    *node,
    self.calculate_node_score(
        node,
        adults.clone(),
        IssueType::PendingRequestOperation([1; 32])
    )
    .await,
);
```

As can be seen, you need to supply an ugly placeholder value that has no effect. For this reason, I
decided to just supply the `op_id` as an optional argument on `track_issue`. Neither solution is
completely ideal, but I prefer this one.

Remaining changes were a bit more superficial. A `ScoreResults` helper struct was introduced to
return all three types of scores from the refactored `calculate_scores` function, rather than
returning a tuple of three hash tables, which would have been a bit cumbersome.


sn_dysfunction/src/detection.rs
```diff
@@ -18,13 +18,26 @@ static OP_WEIGHTING: f32 = 1.5;
...
-    Op,
+/// Represents the different type of issues that can be recorded by the Dysfunction Detection
+/// system.
+pub enum IssueType {
+    /// Represents a communication issue to be tracked by Dysfunction Detection.
+    Communication,
+    /// Represents a knowledge issue to be tracked by Dysfunction Detection.
+    Knowledge,
+    /// Represents a pending request operation issue to be tracked by Dysfunction Detection.
+    PendingRequestOperation,
```
joshuef, line 33: Can we not make this `PendingRequestOperation(OperationId)` ? That would clean up the exposed API somewhat.

sn_dysfunction/src/error.rs
```diff
@@ -19,6 +19,9 @@ pub enum Error {
     /// System time error
     #[error("Could not calculate system time")]
     CouldNotCalculateSystemTime,
+    /// Errors that occur while adding issues to the dysfunction tracker.
+    #[error("Error tracking issue: {0}")]
+    IssueTracking(String),
```
joshuef, line 24: @jacderida I think we're best avoiding a generic string here as it opens up `Error` abuse. By which I mean, it's super easy to just throw _anything_ in here.

I think it's better to just keep it specific and use a crate type if needed.

eg 
```rust
    /// Error that can occur while adding issues to the dysfunction tracker.
    #[error("An operation ID only applies to pending unfulfilled requests."]
    UnusedOpIdSupplied
``` 
instead of `IssueTracking("An operation ID only applies to pending unfulfilled requests.")`

joshuef (changes requested):

sn_dysfunction/src/detection.rs
jacderida: Hey Josh, yeah, I tried that. There's an explanation on the description as to why I didn't go with it and had the `op_id` argument as the lesser of two evils. If you know how to work around it, I'd be happy to change.

sn_dysfunction/src/error.rs
jacderida: Thanks. Will get this changed.

sn_dysfunction/src/detection.rs
joshuef: So you did! Sorry I missed that. Looked at this first thing and just dived right into the files changed 🤦 .

I wonder if it's a two enum situation? `Issue::PendingOperation(OpId)` and `IssueType::PendingOperation`... not sure that's much better. It does seem a shame to have to None a lot just for this one location though

joshuef (approved):

Yoga07:
> I wonder if it's a two enum situation? Issue::PendingOperation(OpId) and IssueType::PendingOperation... not sure that's much better. It does seem a shame to have to None a lot just for this one location though

We could probably also do `IssueType::PendingOperation(Option<OpId>)` it. IMO, having an OpId tied to the IssueType helps in reading the code and makes it easier to understand what operation it stands for.

sn_dysfunction/src/network_knowledge.rs
```diff
@@ -6,26 +6,10 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{error::Result, DysfunctionDetection, NodeIdentifier};
-use std::time::Instant;
-
-impl DysfunctionDetection {
-    /// Track an issue with a node's network knowledge
-    pub async fn track_knowledge_issue(&self, node_id: NodeIdentifier) -> Result<()> {
```
Yoga07, line 14: Same as the previous comment. We only have tests here. 
Maybe we could have all these tests under a single file?

sn_dysfunction/src/network_knowledge.rs
```diff
@@ -128,43 +112,35 @@ mod tests {
...
             .get_nodes_beyond_severity(DysfunctionSeverity::Suspicious)
             .await?;
-        // Assert that the new adult is detected as suspect.
         assert_eq!(sus.len(), 1, "only one node is sus");
         assert!(sus.contains(&new_adult), "our adult is not sus");
 
         let dysfunctional_nodes = dysfunctional_detection
             .get_nodes_beyond_severity(DysfunctionSeverity::Dysfunctional)
             .await?;
+        println!("len = {}", dysfunctional_nodes.len());
```
Yoga07, line 142: Can be removed?

Yoga07 (approved):

PR #1127, opened 2022-04-21
Back pressure debug

joshuef:
Previously we had backpressure running in response to every single message that came in. 

This was somehow disrupting section formation and DKG. Moving it to an optional feature helped stabilise the base test cases in `main` w/ the membership changes.

Here, I look to reenable backpressure as a default feature. 

I've moved it away from being triggered on _every_ incoming message, to be a periodic check on a loop over an interval.

It also now _only_ fires back pressure messages to the same section. 

(As things stand, intersection comms is currently limited to AE messages flows which should not overwhelm our nodes if backpressure within the section - which is where most messages arise - is working properly).

(We _could_ fire backpressure to every connected node peer eg, but we don't know if they are clients so may be wasting time there...)

### DIFF
--- a/sn_interface/src/types/log_markers.rs
+++ b/sn_interface/src/types/log_markers.rs
@@ -26,7 +26,7 @@ pub enum LogMarker {
     ProposalAgreed,
     // Handover
     HandoverConsensusTrigger,
-    HandoverMsgToBeHandled,
+    HandoverMsgBeingHandled,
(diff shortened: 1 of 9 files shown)


Yoga07 (approved):

joshuef:
Looks like we've a couple CI fails. One due to insufficient elders there. Dannnnggg. Will look deeper monday!

PR #1128, opened 2022-04-21
tests(resource_proof): test valid nonce signature

RolandSherwin:
Makes sure that the `nonce_signature` is signed by the correct peer.

### DIFF
--- a/sn_node/src/node/api/tests/mod.rs
+++ b/sn_node/src/node/api/tests/mod.rs
@@ -185,7 +185,12 @@ async fn membership_churn_starts_on_join_request_with_resource_proof() -> Result
     let data = rp.create_proof_data(&nonce);
     let mut prover = rp.create_prover(data.clone());
     let solution = prover.solve();
-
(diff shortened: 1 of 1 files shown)


sn_node/src/node/api/tests/mod.rs
```diff
@@ -224,6 +226,16 @@ async fn membership_churn_starts_on_join_request_with_resource_proof() -> Result
         .unwrap()
         .is_churn_in_progress());
 
+    // makes sure that the nonce signature is always valid
+    let random_peer = Peer::new(xor_name::rand::random(), gen_addr());
+    assert_eq!(
+        false,
```
Yoga07, line 232: This can be simplified as `assert!(!dispatcher.node.validate_resource_proof_response(&random_peer.name(), resource_proof_response).await);`

RolandSherwin: Ah yeah you're right. I have pushed the simplified code.

Yoga07 (approved):

PR #1134, opened 2022-04-23
Remove redundant warning about restarting authd

willief:
For a long time now a spurious warning has been emitted when changing networks.
```
Successfully switched to 'comnet' network in your system!
If you need write access to the 'comnet' network, you'll need to restart authd (safe auth restart), unlock a Safe and re-authorise the CLI again
```

This warning has been irrelevant ever since the auth crate functionality was subsumed into other crates.

Remove line 70 of  /sn_cli/src/subcommands/networks.rs


bochaco:
@willief , I'll be rebasing ad resolving the conflicts so this can be merged, I'll also revert the change of node default capacity to 2GB as I presume that wasn't an intended change?

PR #1140, opened 2022-04-26
tests(storage): model-based test for DataStorage module

RolandSherwin:
Model based testing using the following logic:
- Create random combinations of Store, Get, Query, Remove operations using `quickcheck` lib
- Do the operations on `DataStorage` module as well as a `HashMap` and make sure that their results don't diverge.

Few shortcomings:
1. Cannot do non-blocking async calls because `quickcheck` does not support it (as far as I know). Found [this](https://github.com/nytopop/quickcheck_async) library, but was unsuccessful using it. Thus takes ~25 seconds to do 100 quickcheck tests on my system.
2. Only implemented the tests for `ChunkStorage` field. 
3. The chunk sizes are in the range of 1 to 3 mb only.


RolandSherwin:
Looking into why its failing.

RolandSherwin:
The issue boils down to performing Get/Query on a chunk that was just stored. It happens very rarely and it can be verified by running the `data_storage_basics` on a loop.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +315,126 @@ mod tests {
...
+        // +1 to get None in get_index
+        let rand = rng.gen_range(0, size + 1);
+        match model.get_index(rand) {
+            Some((xor_name, _)) => *xor_name,
+            None => xor_name::rand::random(),
+        }
+    }
+
+    fn model_based_test_imp(ops: Vec<Op>) -> Result<(), &'static str> {
+        let mut model: IndexMap<XorName, ReplicatedData> = IndexMap::new();
```
Yoga07, line 361: Think we could just use a hashmap/btreemap here. We seem to be adding a new dependency just for getting a random key. We can do something like `btreemap.iter().nth(random_idx)` to get a random key.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +315,126 @@ mod tests {
...
+        let used_space = UsedSpace::new(usize::MAX);
+        let runtime = Runtime::new().unwrap();
+        let storage = DataStorage::new(path, used_space).unwrap();
+
+        for op in ops.into_iter() {
+            match op {
+                Op::Store(data) => {
+                    let data = *data;
+                    let _ = runtime.block_on(storage.store(&data)).unwrap();
+                    let data_copy = data.clone();
```
Yoga07, line 373: Is this clone necessary?

sn_node/src/node/core/data/storage/mod.rs
RolandSherwin: I wanted a fast way to get to a particular index in the hashmap. But maybe adding a new dependency is not worth the speed gain.

sn_node/src/node/core/data/storage/mod.rs
RolandSherwin: I added it during debugging. Have removed it now.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +315,127 @@ mod tests {
...
+    }
+    fn arbitrary_ops(count: impl Into<SizeRange>) -> impl Strategy<Value = Vec<Op>> {
+        proptest::collection::vec(arbitrary_single_op(), count)
+    }
+
+    fn get_random_key(model: &BTreeMap<XorName, ReplicatedData>) -> XorName {
+        let mut rng = rand::thread_rng();
+        let size = max(model.len(), 1);
+        // +1 to get None in get_index
+        let rand = rng.gen_range(0, size + 1);
```
davidrusu, line 355: This randomness should be generated on the Op side. i.e. `Op::Remove` should be `Op::Remove(idx: usize)`. Then you can use the `idx` in the `.nth(idx)` call below.

The problem with generating the key inside the test is that you will not be able to re-produce test-cases when a failure occurs

davidrusu: Same goes for `Op::Query`

davidrusu: You want all randomness to be coming from the inputs. Once the inputs are fixed, the test run should be deterministic

RolandSherwin:
Working on making the Store operation insert already present data.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +314,139 @@ mod tests {
...
+        }
+    }
+
+    fn model_based_test_imp(ops: Vec<Op>) -> Result<(), &'static str> {
+        let mut model: BTreeMap<XorName, ReplicatedData> = BTreeMap::new();
+        let temp_dir = tempdir().unwrap();
+        let path = temp_dir.path();
+        let used_space = UsedSpace::new(usize::MAX);
+        let runtime = Runtime::new().unwrap();
+        let storage = DataStorage::new(path, used_space).unwrap();
```
Yoga07, line 361: We ideally want to avoid using unwraps, excepts, and panics in all our tests.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +314,139 @@ mod tests {
...
+                            if stored_data.is_ok() {
+                                return Err("ReplicatedData found while it should not exist");
+                            }
+                        }
+                    }
+                }
+                Op::Remove(idx) => {
+                    let key = get_xor_name(&model, idx % (model.len() + 1));
+                    let addr = ReplicatedDataAddress::Chunk(ChunkAddress(key));
+                    let _ = runtime.block_on(storage.remove(&addr));
```
Yoga07, line 444: We should check if the results being returned from the Future inside are fine.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +314,139 @@ mod tests {
...
+                                }
+                            }
+                        }
+                        // random bytes
+                        _ => {
+                            let chunk = Chunk::new(random_bytes(chunk_size * 1024 * 1024));
+                            ReplicatedData::Chunk(chunk)
+                        }
+                    };
+                    let _ = runtime.block_on(storage.store(&data)).unwrap();
```
Yoga07, line 384: Same as above, we should check if the results being returned are fine and also drop the unwrap.

sn_node/src/node/core/data/storage/mod.rs
RolandSherwin: Some of the errors returned here include file `NotFound` error and similar io related ones. So if I were to check them here, I need to first fetch the data from the hashmap. Or another approach is to whitelist some errors. Or is it better to just not bother with them at all? since I'm doing all the checking in the Get/Query Operation.

Yoga07: Our test scenario expects a happy outcome, so anything that makes us sad should be returned as an error 🤓 
We can match the result to achieve the same.
```
match result {
    Ok(()) => {},
    Err(e) => return Err(e),
}
```

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +314,143 @@ mod tests {
...
+
+                    match model_res {
+                        Some(m_res) => {
+                            if let NodeQueryResponse::GetChunk(Ok(s_chunk)) = stored_res {
+                                if let ReplicatedData::Chunk(m_chunk) = m_res {
+                                    assert_eq!(*m_chunk, s_chunk);
+                                }
+                            }
+                        }
+                        None => if let NodeQueryResponse::GetChunk(Ok(_)) = stored_res {},
```
Yoga07, line 409: If the data does not exist in the model, it shouldn't be present in the disk too. Seems we are allowing that here?

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -299,4 +314,143 @@ mod tests {
...
+                                return Err(Error::DataExists);
+                            }
+                        }
+                    }
+                }
+                Op::Remove(idx) => {
+                    let key = get_xor_name(&model, idx % (model.len() + 1));
+                    let addr = ReplicatedDataAddress::Chunk(ChunkAddress(key));
+
+                    let model_res = runtime.block_on(storage.remove(&addr));
```
Yoga07, line 437: Rename to `storage_res`? Get's a bit confusing as we think of the hashmap as the model.

sn_node/src/node/core/data/storage/mod.rs
RolandSherwin: Oops. Forgot to include in the the Error statement. Fixed now

sn_node/src/node/core/data/storage/mod.rs
RolandSherwin: Yeah. Fixed it :)

Yoga07 (approved):

PR #1146, opened 2022-04-27
Dkg generations

joshuef:

### DIFF
--- a/sn_node/src/node/core/split_barrier.rs
+++ b/sn_node/src/node/core/split_barrier.rs
@@ -6,22 +6,24 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::mem;
-
 use crate::node::dkg::KeyedSig;
+use sn_consensus::Generation;
 use sn_interface::messaging::system::SectionAuth;
 use sn_interface::network_knowledge::SectionAuthorityProvider;
+use std::collections::BTreeMap;
 use xor_name::Prefix;
-
 type Entry = (SectionAuth<SectionAuthorityProvider>, KeyedSig);
+use std::sync::Arc;
+
+use tokio::sync::RwLock;
 
 // Helper structure to make sure we process a split by updating info about both our section and the
 // sibling section at the same time.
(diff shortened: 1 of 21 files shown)


sn_node/src/node/core/split_barrier.rs
```diff
@@ -31,29 +33,35 @@ impl SplitBarrier {
...
     #[instrument(skip(self), level = "trace", name = "split barrier processing")]
-    pub(crate) fn process(
+    pub(crate) async fn process(
         &mut self,
         our_prefix: &Prefix,
         section_auth: SectionAuth<SectionAuthorityProvider>,
         keyed_sig: KeyedSig,
+        generation: Generation,
     ) -> Vec<Entry> {
+        debug!(">>>>Processing split barrier for gen: {generation}");
```
Yoga07: Can be cleaned up?

sn_node/src/node/core/split_barrier.rs
```diff
@@ -31,29 +33,35 @@ impl SplitBarrier {
...
-            mem::take(&mut self.0)
+        let (mut give, mut keep): (Vec<Entry>, Vec<Entry>) =
+            gen_entries
                 .into_iter()
                 .partition(|(cached_section_auth, _)| {
                     cached_section_auth.prefix() == section_auth.prefix().sibling()
                 });
-        self.0 = keep;
 
+        // let _prev = gen_map_guard.insert(generation, keep.clone());
```
Yoga07, line 60: Can be cleaned up?

sn_node/src/node/core/split_barrier.rs
```diff
@@ -31,29 +33,35 @@ impl SplitBarrier {
...
         if give.is_empty() {
             // No sibling found. Cache this update until we see the sibling update.
-            self.0.push((section_auth, keyed_sig));
+            keep.push((section_auth, keyed_sig));
```
Yoga07, line 63: Could there be a scenario where give is some and keep is empty? Just out of curiosity.

joshuef: I dont think so. If we have one already, we get give (if its new). I'm not 100% here though, mostly just following the prior logic

PR #1159, opened 2022-05-04
wip: safe network gui

davidrusu:
To run the gui:

```
cd sn_node
cargo run --release --bin sn_node_gui
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,22 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
+version = "0.2.15"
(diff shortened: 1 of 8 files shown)


micwoj92:
This branch has conflicts that must be resolved

davidrusu:
@micwoj92 we're working on some code changes to the node internals that should make the `sn_playground` idea more workable, namely we're looking to moving out more of the networking and data storage modules from `Node`.

If that work pans out, we'll be able to mock out those networking/datastorage calls in the playground. That'd result in a fully virtual safe network running in memory.

PR #1168, opened 2022-05-05
Check incoming authed SAP in signed votes

grumbach:
- Make sure incoming vote's auth'd SAPs are correctly signed by elders before we forward them to consensus.

### DIFF
--- a/sn_node/src/node/core/messaging/handling/handover.rs
+++ b/sn_node/src/node/core/messaging/handling/handover.rs
@@ -88,18 +88,14 @@ impl Node {
     pub(crate) async fn handle_handover_msg(
         &self,
         signed_vote: SignedVote<SapCandidate>,
-    ) -> Vec<Cmd> {
+    ) -> Result<Vec<Cmd>> {
         debug!(
             ">>> {}: {:?}",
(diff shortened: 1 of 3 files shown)


grumbach:
Seems like the test `split_network_assert_health_check ` is failing on the latest main branch as well (without the changes included in this PR)

grumbach:
> Seems like the test `split_network_assert_health_check ` is failing on the latest main branch as well (without the changes included in this PR)

Now should be fixed

sn_node/src/node/core/messaging/handling/handover.rs
```diff
@@ -54,6 +95,12 @@ impl Node {
             signed_vote
         );
 
+        if let Err(err) = self.check_signed_vote_saps(&signed_vote) {
+            // NB TODO tracking who sent this invalid vote?
+            error!("Ignoring invalid handover vote: {}", err);
```
joshuef, line 100: Let's return an error here. I think we can then look to hook into dysfunction in a layer above.

joshuef (changes requested):
Return an error in order to react above to dysfunction there.

joshuef (approved):

PR #1173, opened 2022-05-09
Chore remove propose offline cmd voting is enough

joshuef:


joshuef:
Totally missed those replies @davidrusu . Thanks. yeh I was meaning to get your take on this. Nice one. Clooosing.

PR #1174, opened 2022-05-09
fix(dysfunction): relax dysfunction for knowledge and conn issues

joshuef:
Increases 10x the amount of conn or knowledge issues. We've been voting
off nodes far too quickly, even on droplet testnets


jacderida (approved):

jacderida:
It would be interesting to see what the impact is on my prop test using these values. The weights are dramatically different.

PR #1178, opened 2022-05-11
Section probing for all nodes to trigger rejoin if needed

grumbach:
- Section probing for all nodes every now and then
- Removed unused/ignored `dst` field in `AntiEntropyProbe`
- Trigger `ChurnJoinMiss` error and event when nodes realise they are not members of a section they should be in during AE
- Node rejoins/restart when they receive `ChurnJoinMiss`

### DIFF
--- a/sn_node/examples/routing_minimal.rs
+++ b/sn_node/examples/routing_minimal.rs
@@ -240,6 +240,9 @@ async fn handle_event(index: usize, node: &mut NodeApi, event: Event) -> bool {
         Event::MemberLeft { name, age } => {
             info!("Node #{} member left - name: {}, age: {}", index, name, age);
         }
+        Event::ChurnJoinMissError => {
+            info!("Node #{} detected churn join miss and will restart", index);
+        }
         Event::SectionSplit {
             elders,
             self_status_change,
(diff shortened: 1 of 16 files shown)


sn_node/examples/routing_minimal.rs
```diff
@@ -240,6 +240,9 @@ async fn handle_event(index: usize, node: &mut NodeApi, event: Event) -> bool {
         Event::MemberLeft { name, age } => {
             info!("Node #{} member left - name: {}, age: {}", index, name, age);
         }
+        Event::ChurnJoinMissError => {
+            info!("Node #{} detected churn join miss and will restart", index);
```
davidrusu, line 244:
```suggestion
            info!("Node #{index} detected churn join miss and will restart");
```

davidrusu (approved):
Looks good to me :ship:

sn_node/examples/routing_minimal.rs
grumbach: I tried to be consistent with the above, but I've also seen other parts of the code with that technique. I think we should come up with a norm for Maidsafe for this kind of things.

Issue #1118, opened 2022-04-17
`cargo build --tests --bins` fails


jonassmedegaard:
Compiling workspace with `cargo build --tests --bins` fails with many places referencing code from `safe_network` feature `test-utils`:
```
error[E0432]: unresolved import `crate::node::create_test_max_capacity_and_root_storage`
  --> sn/src/node/api/tests/mod.rs:20:5
   |
20 |     create_test_max_capacity_and_root_storage,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `create_test_max_capacity_and_root_storage` in `node`

error[E0432]: unresolved import `sn_interface::network_knowledge::test_utils`
  --> sn/src/node/api/tests/mod.rs:35:38
   |
35 | use sn_interface::network_knowledge::test_utils::{prove, section_signed};
   |                                      ^^^^^^^^^^ could not find `test_utils` in `network_knowledge`

error[E0432]: unresolved import `sn_interface::network_knowledge::test_utils`
   --> sn/src/node/core/delivery_group.rs:173:42
    |
173 |     use sn_interface::network_knowledge::test_utils::section_signed;
    |                                          ^^^^^^^^^^ could not find `test_utils` in `network_knowledge`

error[E0432]: unresolved import `crate::node::create_test_max_capacity_and_root_storage`
   --> sn/src/node/core/messaging/handling/anti_entropy.rs:494:34
    |
494 |         api::tests::create_comm, create_test_max_capacity_and_root_storage, MIN_ADULT_AGE,
    |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `create_test_max_capacity_and_root_storage` in `node`

error[E0432]: unresolved import `sn_interface::network_knowledge::test_utils`
   --> sn/src/node/core/messaging/handling/anti_entropy.rs:496:42
    |
496 |     use sn_interface::network_knowledge::test_utils::section_signed;
    |                                          ^^^^^^^^^^ could not find `test_utils` in `network_knowledge`

error[E0432]: unresolved import `sn_interface::network_knowledge::test_utils`
   --> sn/src/node/core/relocation.rs:168:42
    |
168 |     use sn_interface::network_knowledge::test_utils::section_signed;
    |                                          ^^^^^^^^^^ could not find `test_utils` in `network_knowledge`

error[E0433]: failed to resolve: could not find `test_utils` in `network_knowledge`
[...]

PR #1187, opened 2022-05-16
Refactor test-utils feature flag

RolandSherwin:
Fix for #1118

Objective:<s>
1. Remove `test-utils` feature flag such that</s>  `cargo test` can be run standalone.<s>
2. Classify between fns used for tests and for bench/example.</s>

Approach:<s>
Introduce two feature flags, `test` and `utils`, where they are used for tests and bench/examples respectively.</s> The different possible scenarios in the code and the approach taken for each are:
1. **Same Crate:** If a fn is used only for testing within the **same crate**, just use `cfg(test)`
2. **Diff Crate:** I.e., a fn defined in one crate is called in another crate. Can be solved by feature gating the function with `cfg(feature="test-utils")` and then importing the fn as `dev-dependency` with `test` feature enabled. Eg: Inside sn_node's cargo.toml, `sn_interface = { path = "../sn_interface", version = "^0.2.3", features= ["test-utils"] }`
4. **Same Crate, diff folder (test/example/bench):** Feature gate it if possible. **Else no `cfg`**


sn_client/src/lib.rs
```diff
@@ -156,14 +156,11 @@ where
...
-#[cfg(any(test, feature = "test-utils"))]
 use std::sync::Once;
 
-#[cfg(any(test, feature = "test-utils"))]
 static INIT: Once = Once::new();
 
 // TODO: dedupe this w/ sn_node?
-#[cfg(any(test, feature = "test-utils"))]
 /// Initialise logger for tests, this is run only once, even if called multiple times.
 pub fn init_test_logger() {
```
davidrusu, line 165: Is this used in non-test code? maybe move it into the `mod tests { ..}` block?

RolandSherwin: It is used in examples, so cannot have it inside `cfg(test)`. Also there seems to be 4 definitions of `init_test_logger` in the code, maybe shall I have a single one inside `sn_interface` under `feature="test"` and import it in other places?

joshuef: > maybe shall I have a single one inside sn_interface under feature="test" and import it in other places?

That sounds good. But then it's probably just `"init_logger"` perhaps?

joshuef:
@RolandSherwin why do we need `cfg(feature="test")` vs something like `cfg(feature="utils")`? 

What are those functions doing? (I just want to be sure they are actually testing code... ie faking some situation, vs just utility code)

joshuef:
> Ah! So for example, functions like init_test_logger() should be under utils while fns like create_test_client_with() should be under test?

As per a HO just now, let's leave off renaming/classifying exports and keep this PR tighter to just `run crate tests with no feature flags` as the aim.

Then later on we can check that everything we're exporting makes sense and classify them a bit better 👍

davidrusu (approved):

PR #1189, opened 2022-05-17
chore(client): send some msgs in bg threads

joshuef:
This should unblock client threads on initial contact and on queries

### DIFF
--- a/sn_client/src/connections/messaging.rs
+++ b/sn_client/src/connections/messaging.rs
@@ -237,7 +237,7 @@ impl Session {
         let msg_kind = AuthKind::Service(auth);
         let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, dst_location)?;
 
-        send_msg(self.clone(), elders, wire_msg, msg_id).await?;
+        send_msg_in_bg(self.clone(), elders, wire_msg, msg_id)?;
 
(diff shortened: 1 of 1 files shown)


sn_client/src/connections/messaging.rs
```diff
@@ -388,7 +388,7 @@ impl Session {
             .take(NODES_TO_CONTACT_PER_STARTUP_BATCH)
             .collect();
 
-        send_msg(self.clone(), initial_contacts, wire_msg.clone(), msg_id).await?;
+        send_msg_in_bg(self.clone(), initial_contacts, wire_msg.clone(), msg_id)?;
```
Yoga07, line 391: Should this be run in background? Because we have a few logics and checks after this to make sure we have made contact to the network. Better we have this run sequentially here alone?

joshuef: Actually this is where I was seeing us hanging. 

We're waiting on connections which are timing out. But we wait/loop below until we get AE messages in (which had actually already _come_ in... but we never got down to the check due to the timeout)

Yoga07: Righttttt... makes sense 👍

Yoga07 (approved):

PR #1191, opened 2022-05-18
Fix lru cache deadlock

joshuef:

### DIFF
--- a/sn_node/benches/data_storage.rs
+++ b/sn_node/benches/data_storage.rs
@@ -1,0 +1,162 @@
+// Copyright 2020 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use criterion::Criterion;
+
+use std::time::Duration;
+
+use eyre::{Result, WrapErr};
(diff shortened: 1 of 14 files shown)


sn_node/benches/data_storage.rs
```diff
@@ -0,0 +1,164 @@
...
+        src_name: sn_interface::types::PublicKey::Bls(public_key).into(),
+        sig,
+    }
+}
+
+fn custom_criterion() -> Criterion {
+    Criterion::default().sample_size(SAMPLE_SIZE)
+}
+
+// #[tokio::main]
```
Yoga07, line 104: Can be removed?

sn_node/benches/data_storage.rs
```diff
@@ -0,0 +1,164 @@
...
+fn main() -> Result<()> {
+    let mut criterion = custom_criterion();
+    criterion = criterion.measurement_time(Duration::from_millis(20_000));
```
Yoga07, line 107: Can the magic number be assigneg to a const?

Yoga07 (approved):

sn_node/benches/data_storage.rs
```diff
@@ -0,0 +1,164 @@
...
+use std::collections::BTreeMap;
+
+// sample size is _NOT_ the number of times the command is run...
+// https://bheisler.github.io/criterion.rs/book/analysis.html#measurement
+const SAMPLE_SIZE: usize = 10;
+
+fn bench_data_storage(c: &mut Criterion) -> Result<()> {
+    let mut group = c.benchmark_group("read-sampling");
+
+    let runtime = Runtime::new().unwrap();
```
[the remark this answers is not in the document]
joshuef: It should be fine in these benches. I was trying to remove them, but then we end up with async code and Criterion, inside of a runtime, where you have to _pass_ a runtime, was getting complex/painful for the compiler

Issue #1186, opened 2022-05-16
Handover: allow re-consensus when elders agreed on an empty set


grumbach:
Handover Consensus can only happen once, so the Handover struct can only ever be created once and is append only until elders reach consensus on a single decision. This decision is a set of values (SapCandidates) and elders have a deterministic way to identify the one that "won" this consensus round. As consensus is now, it might occur that elders agree on an empty set, in this case handover is stuck on that agreement and can't ever happen. As such, we need to accept Handover struct resets whenever we notice that the agreement was on the empty set.

PR #1202, opened 2022-05-24
feat: allow re-consensus when elders agree on an empty set #1186

grumbach:
- generations for handover
- AE for handover 
- empty consensus handling


sn_node/src/node/handover/handover_consensus.rs
```diff
@@ -46,35 +54,101 @@ impl Handover {
...
+            .filter(|(gen, _)| **gen >= from_gen)
+            .map(|(gen, consensus)| {
+                if let Some(decision) = consensus.decision.clone() {
+                    Ok(consensus.build_super_majority_vote(
+                        decision.votes.clone(),
+                        decision.faults,
+                        *gen,
+                    )?)
+                } else {
+                    Err(Error::CorruptedHandoverHistory(format!(
```
davidrusu, line 74: To avoid this error, you can make `failed_consensus_rounds` hold the final decision as well:
`failed_consensus_rounds: BTreeMap<Generation, (Decision, Consensus)>`

sn_node/src/node/handover/handover_consensus.rs
```diff
@@ -46,35 +55,99 @@ impl Handover {
...
+            from_gen,
+            self.gen,
+            self.id()
+        );
+        Ok(proof_votes)
+    }
+
+    pub(crate) fn handle_empty_set_decision(&mut self) {
+        if let Some(decision) = &self.consensus.decision {
+            if decision.proposals.is_empty() {
```
grumbach, line 90: Might need to change to handling this error with new Consensus version:
When `Decision::validate` gives `Err(Error::InvalidDecision)`
Instead of `if decision.proposals.is_empty() {`

https://github.com/maidsafe/sn_consensus/blob/cb8508dcc53023e09c980c3ee01ca28165d426f5/src/decision.rs#L52

PR #1209, opened 2022-05-25
First Steps Toward Owned DBCs

jacderida:
- 474316022 **feat!: extend client with dbc owner field**

  BREAKING CHANGE: new parameter added to `sn_client::api::Client::new`.

  Add a DBC `Owner` field to the client, which will be used in `sn_api` for assigning the Owner of
  the change output DBC.

  Like the client's keypair, it's optional when creating the client; if not supplied, a random key
  will be generated.

  As per this commit, the `connect` function in `sn_api` is supplying `None` for this new field, but
  it will be updated shortly.

- 81e3a21c5 **feat: command to generate secret key for dbc owner**

  A new `keys create-dbc-owner` command is added to generate a secret key for use with reissuing DBCs.
  The key gets serialised to a file in `~/.safe/cli`. It will be read from here when making
  connections to the network.

  This commit also features various other changes. Originally I had two separate commits, but I
  accidentally amended a previous commit and couldn't separate the content.

  The `keys create` command is using functions from a module related to authentication to determine
  where to write the keys it generates. These functions directly specified a credential file's
  location at the user's home directory. To help unit testing, I refactored this to provide the
  location from the `Config` struct. Previously I refactored `networks` commands in a similar fashion.
  The result is, during the unit test, the `assert_fs` library will create a temp directory for the
  config and so the user's real home directory won't be manipulated during the test suite. `assert_fs`
  will also remove the temp directory at the end of each test.

  With this change in place, I provided a couple of unit tests for the `keys create` command. This
  command doesn't hit the network, so integration tests aren't necessary. There are only really two
  code paths to go through: one for writing the keypair to file and one for not doing so.

  I didn't test the `keys show` command because it mostly just prints information to the console,
  and at this stage I don't know how to capture that for a unit test (or if it's even possible).

  Finally, I tidied up some commented out code that doesn't seem to apply any more, and added some
  additional documentation for the commands to try and make their intent a bit clearer.

- 5d129e9d0 **feat: use persistent dbc owner in sn_api**

  The network connection functions in the CLI are updated to pass the persistent DBC Owner. This is
  based on the existence of a key that was created with the new `create-dbc-owner` command. If the key
  exists, the connection will use the persistent DBC owner; if `None` is passed, a random key will be
  generated.

  During a reissue, the change DBC is then assigned the owner from the client. We don't yet assign the
  output DBC this owner, because we're still only supporting bearer DBCs. This may change for owned
  DBCs, which should come soon.

  I split the wallet reissue test out into a couple: one with a single input DBC and one with
  multiple. The original test had multiple inputs, but it seemed it was worth checking that the
  reissue would function if there were also just a single input.

- c4ffc3335 **chore: target for replicating ci test suite**

  This is just so I can run the test suite on a local network before I submit my work for PR. I
  realise it will get out of date due to our CI having frequent changes, but it's just a utility and I
  can update it at any point.
  
  - 0bbf4d882 **chore: remove use of test-utils from test runs**

  After doing a rebase from main, the test-utils feature was removed. I updated the testing targets
  and also replaced bad references to logger initialisation functions.

- 4945f9982 **feat: add bls type to keypair enum**

  Extends the `sn_interface::types::keys::Keypair` enum to support the BLS keypair type. We need this
  because we want to change the CLI to use BLS rather than Ed25519 keys, so we can support signing
  output DBCs with the same keypair we use to sign commands sent from the CLI.

  I modified the tests that check each keypair type can be serialized/deserialized. Previously there
  was one test case looping over each type of keypair, but I think it's better style to define each
  test case explicitly: you are calling out what scenarios you want to support and it makes the cases
  easier to understand at a glance, even if there is a small bit of repetition between them.

- 0e83e5831 **feat!: cli will now use bls keys**

  BREAKING CHANGE: the CLI will now use BLS keys rather than ed25519. This is so we can use the same
  keypair to sign change DBCs in a reissue.

  The `keys` commands were tidied here to move code related to serialization and hex representation
  into `sn_api` and the `Keypair` structs, with some code related to that also removed from `sn_api`.
  The CLI now doesn't need to call `serde` or `bincode` functions directly.

  The `show` command was modified to remove support for passing an `XorUrl` of a key, since this was
  relying on the public key that was encoded in the URL to be ed25519, and that public key happens to
  be a 32-byte structure; a BLS key couldn't be represented with 32 bytes.

  The previously added `create-dbc-owner` command is now removed because we don't need a separate
  key any more.


joshuef:
> Like the client's keypair, it's optional when creating the client; if not supplied, a random key
will be generated.

@jacderida Why do we need two keypairs here? Why would we not use the client keypair?

jacderida:
Hi @joshuef 

That was my initial thought, but @davidrusu advised the creation of another key. I think because the two are incompatible? Would you mind commenting on this please David?

The CLI currently creates an Ed25519 keypair, which I thought we could convert to BLS, but apparently not?

davidrusu:
Dbc's use BLS as owners, so we need each client who will be owning DBC's to have a BLS keypair.

If the ed25519 keypair had no other uses, we can drop it entirely. I don't know what uses it has right now.

Makefile
```diff
@@ -164,3 +164,20 @@ run-local-baby-fleming:
...
+ci-unit-tests:
+	cargo test --no-run --release --package sn_interface --package sn_dysfunction --package sn_node
+	cargo nextest run --profile ci --release --package sn_interface messaging prefix_map types
+	cargo nextest run --profile ci --release --package sn_dysfunction
+	cargo nextest run --profile ci --release --package sn_node dbs node routing
+	cargo nextest run --profile ci --release --package sn_cli --bin safe
+
+ci-e2e-tests:
+	cargo nextest run --profile ci --release --package sn_client --test-threads 2
+	cargo nextest run --profile ci --release --package sn_client --test-threads 2 ae_checks
```
davidrusu, line 177: Won't the `ae_checks` also be run in the previous line?

sn_api/src/app/keys.rs
```diff
@@ -37,4 +38,106 @@ impl Safe {
             Ok(encode(&derived_xorname))
         }
     }
+
+    /// Generate a new random BLS keypair along with a URL for the public key.
+    pub fn new_keypair_with_pk_url(&self) -> Result<(Keypair, SafeUrl)> {
+        let keypair = Keypair::new_bls();
+        let xorname = XorName::from(keypair.public_key());
+        let url = SafeUrl::encode_safekey(xorname, self.xorurl_base)?;
+        let url = SafeUrl::from_url(&url)?;
```
davidrusu, line 47: Very odd that we can't just create a SafeUrl for a xorname directly, something to look into in future work

davidrusu (approved):
Reads well, good testing throughout as well :ship:

PR #1216, opened 2022-06-03
RFC: Remove Private Scope

davidrusu:
Why? The main distinction between private and public data is:

1. private data can be deleted, public data can not
2. private data is encrypted separately from self_encryption. public data is not encrypted again (besides self_encryption)

Regarding 1, we don't have a way to implement deletion reliably, if the data is ever re-uploaded, or one of the adults fails to delete it for some reason (missed a packet), it would remain replicated and held by the network. So right now we don't have a story for how to delete reliably, but we are paying the complexity cost of keeping track of the public/private distinction: (see the large diff)

Regarding 2. We can expose the option to the client to encrypt some data before updating, we don't need to expose this private/public distinction to the network in order to support this use-case.

This was more of an exploration to see how much this public/private data distinction is costing us in terms of code complexity.


joshuef (approved):

davidrusu:
bors retry

davidrusu:
bors cancel

PR #1215, opened 2022-06-02
Handover SAP elder candidates checks

grumbach:
- Check candidate elders in SAPs from received votes, by correlating them with our membership knowledge
- Add membership generation in SAPs
- Remove membership generation from DKG `Session`, move it to the `DkgSessionId`
- Move all SapCandidates checks to the same place in [here](https://github.com/maidsafe/safe_network/pull/1215/files/c95030e63ae554b2169fd877054b1ab30558b008#diff-1a1f60894e2571c5b9878e6dd6a3400756d27de54479d87a44d7a82c475f381f)

PR #1217, opened 2022-06-03
feat: handover sap elder checks with membership knowledge

grumbach:
Squashed version of PR #1215  
https://github.com/maidsafe/safe_network/pull/1215


sn_node/src/node/core/messaging/handling/handover.rs
```diff
@@ -82,24 +87,146 @@ impl Node {
...
+            error!("Missing membership instance when checking handover SAP candidates");
+            Err(Error::MissingMembershipInstance)
+        }
+    }
+
+    async fn get_sap_for_prefix(&self, prefix: Prefix) -> Result<SectionAuthorityProvider> {
+        match self.network_knowledge.prefix_map().get(&prefix) {
+            Some(sap) => Ok(sap),
+            None => Err(Error::FailedToGetPreviousSAP),
+        }
```
davidrusu, line 103:
```suggestion
        self.network_knowledge.prefix_map().get(&prefix).ok_or(Error::FailedToGetPreviousSAP)
```

davidrusu (approved):

davidrusu (approved):

joshuef:
looks like bors failed with a CI timeout, but no strong reason for it that I could see (bit of a long build perhaps). May be the timeout is just a weeee bit too short. Reborsed and we can increase the timeout separately

PR #1218, opened 2022-06-03
Reissue DBC to a Particular Owner

jacderida:
- cd85844f9 **feat!: reissue dbc to a particular owner**

  BREAKING CHANGE: the `wallet_reissue` API is extended to add an `owner_public_key` argument. This is
  an `Option<bls::PublicKey>`. To reissue to a bearer DBC, use `None`, but to reissue to a particular
  owner, specify the key.

  I've also taken an opportunity to update the API documentation a little. In the doc comments and
  logging output, I changed "Wallet" references to "wallet": in this context, "wallet" is not a proper
  noun, so it shouldn't be capitalised (unless it happens to be at the start of a sentence). If you
  wanted to refer to the type `Wallet`, that would be capitalised, but it should also be referred to
  using enclosing backticks.

- a1c6aebe6 **feat!: wallet deposit read input dbc from stdin**

  BREAKING CHANGE: the positional `dbc` argument is changed to an optional argument.

  The DBC hex representation is 8000+ characters long, so this gives the user the option to provide
  the large input DBC from a file, like so:
  ```
  safe wallet deposit <safe-url> < ~/.safe/input_dbc
  ```

  Bash seems to allow you to provide this huge argument on the command line, but it doesn't work in
  Fish, which is the shell I use.

- e548388c6 **chore: upgrade sn_dbc to 3.2.0**

  This new release has utilities for serializing/deserializing `Dbc` to/from hex.

- 0d97494f7 **feat: add public key argument for owned dbcs**

  The `wallet reissue` command now has an additional optional argument, `--public-key`, which allows
  the user to reissue a DBC to be owned by the holder of that public key. The key should be BLS
  hex-encoded.

  The `wallet deposit` command will now require extension to provide the secret key when depositing an
  owned DBC. This will be done as a separate piece of work.

  Some additional changes were made in support or to tidy CLI-related code:
  * The conversion of DBCs to/from hex were removed from the CLI since this is now done on the `Dbc`
    type.
  * A CLI test that existed to test the above conversion code was removed since it's no longer
    necessary.
  * The naming scheme for the CLI wallet tests were elaborated and the redundant "calling_safe"
    prefixes were removed.


sn_api/src/app/keys.rs
```diff
@@ -102,9 +102,7 @@ mod tests {
         assert_eq!(url, url2);
 
         match keypair {
-            Keypair::Bls(_) => {
-                return Ok(());
-            }
+            Keypair::Bls(_) => Ok(()),
             _ => {
                 return Err(eyre!("A BLS keypair should be generated by default."));
             }
```
davidrusu, line 108: Apply the same change to the err arm as well?
```suggestion
            _ => Err(eyre!("A BLS keypair should be generated by default."))
```

davidrusu (approved):
Looks good :ship:

sn_api/src/app/keys.rs
jacderida: Ah, thanks. Clippy only flagged the `Ok` one. Will change this.

PR #1222, opened 2022-06-07
feat: dbc arg can refer to file or dbc data

jacderida:
- 586036cdd **chore: target to provide dbc test data**

  Little target that quickly writes out a few DBC data files that can then be used when testing wallet
  commands. I got tired of looking the strings up in the wallet test file and manually writing them
  out to file.

- 0e62a0c07 **feat: dbc arg can refer to file or dbc data**

  The `--dbc` argument for the `wallet deposit` command is extended so it can refer to either a file
  path or DBC data. It will be tested first for an existing file path, otherwise it will fall back to
  being interpreted as DBC data.

  Some people have been having trouble with pasting DBC data directly via the command line, depending
  on what OS or shell is in use. You can still also pass the data via stdin, but that's only available
  on Linux or macOS.


sn_cli/tests/cli_wallet.rs
```diff
@@ -50,6 +51,109 @@ fn wallet_deposit_should_deposit_a_dbc() -> Result<()> {
...
+
+    safe_cmd(
+        [
+            "wallet",
+            "deposit",
+            "--name",
+            "my-first-dbc",
+            "--dbc",
+            dbc_data_file
+                .path()
```
bochaco, line 71: I think you can simply call `display()` and don't need to unwrap a `Result`

bochaco (approved):

bochaco (approved):

PR #1228, opened 2022-06-09
fix: Node init and restart on error

joshuef:
Previously an error from the node was not triggering the restart of the
node instance. (eg, ChurnJoinMiss).

Here, the node bin start process has been tweaked. Kepeing all
node processes within the sn_node spawned thread. But, each 'node'
with its own runtime, to aid in shutting down any background processes
between restarts.

### DIFF
--- a/sn_node/src/bin/sn_node.rs
+++ b/sn_node/src/bin/sn_node.rs
@@ -28,9 +28,14 @@
 )]
 
 use color_eyre::{Section, SectionExt};
-use eyre::{eyre, ErrReport, Result, WrapErr};
+#[cfg(not(feature = "tokio-console"))]
+use eyre::Error;
+
+use eyre::{eyre, Context, ErrReport, Result};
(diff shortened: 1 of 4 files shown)


sn_node/src/bin/sn_node.rs
```diff
@@ -55,44 +62,180 @@ fn main() -> Result<()> {
...
+            const MAX_RESTARTS : usize  = 50;
+            let mut count = 0;
+
+            let mut res : Result<(), ErrReport> = Ok(());
+            loop {
+                count += 1;
+                if count > 50 {
+                    error!("Max ({MAX_RESTARTS}) node restarts reached");
+                    break
+                }
```
davidrusu, line 96: In ChurnJoinMiss, a node is removed from the network from no fault of their own.

If I understand this code correctly, the counting here will stop the node if he's part of >50 ChurnJoinMiss events. Which will eventually happen for long running nodes.

Can we remove this MAX_RESTARTS check and just perhaps just log the error triggering the restart?

PR #1234, opened 2022-06-13
Enable DKG issue tracking

joshuef:
This is atop #1232 so need that in first

### DIFF
--- a/sn_dysfunction/src/detection.rs
+++ b/sn_dysfunction/src/detection.rs
@@ -11,15 +11,15 @@ use std::collections::{BTreeMap, BTreeSet};
 use xor_name::XorName;
 
 use std::time::Duration;
-static RECENT_ISSUE_DURATION: Duration = Duration::from_secs(60 * 15);
+static RECENT_ISSUE_DURATION: Duration = Duration::from_secs(60 * 10); // 10 minutes
 
-static CONN_WEIGHTING: f32 = 1.0;
+static CONN_WEIGHTING: f32 = 2.0;
 static OP_WEIGHTING: f32 = 1.0;
-static KNOWLEDGE_WEIGHTING: f32 = 3.0;
-static DKG_WEIGHTING: f32 = 5.0;
+static KNOWLEDGE_WEIGHTING: f32 = 2.0;
+static DKG_WEIGHTING: f32 = 4.0;
 
 /// Z-score value above which a node is dysfunctional
-static DYSFUNCTIONAL_DEVIATION: f32 = 1.1;
+static DYSFUNCTIONAL_DEVIATION: f32 = 1.2;
 
 #[derive(Clone, Debug)]
 /// Represents the different type of issues that can be recorded by the Dysfunction Detection
@@ -294,6 +294,7 @@ mod tests {
     fn issue_type_strategy() -> impl Strategy<Value = IssueType> {
         prop_oneof![
             Just(IssueType::Communication),
+            Just(IssueType::Dkg),
             Just(IssueType::Knowledge),
             (any::<[u8; 32]>())
                 .prop_map(|x| IssueType::PendingRequestOperation(Some(OperationId(x))))
@@ -385,7 +386,7 @@ mod tests {
         /// Test that gives a range of nodes and a few bad nodes,
         /// we then check that we can reliably detect those nodes
         fn pt_correct_amount_of_dysf_nodes_should_be_detected(
-            good_nodes in 7..50, bad_nodes in 1..7, issue_count in 1000..5000, issue_type in issue_type_strategy())
+            good_nodes in 7..50, bad_nodes in 1..7, issue_count in 2000..10000, issue_type in issue_type_strategy())
             {
 
             // finish early as we're over byzantine levels
@@ -397,8 +398,8 @@ mod tests {
             let _outer_span = tracing::info_span!("pt_correct_amount_of_dysf_nodes_should_be_detected").entered();
 
             // tolerances...
-            // So here a dysf node fails 30% of the time
-            const DYSF_SUCCESS_RATIO : f32 = 0.3;
+            // So here a dysf node fails 35% of the time
+            const DYSF_SUCCESS_RATIO : f32 = 0.35;
             const NORMAL_SUCCESS_RATIO: f32 = 0.95;
             use rand::seq::SliceRandom;
 
--- a/sn_dysfunction/src/detection.rs
+++ b/sn_dysfunction/src/detection.rs
@@ -13,7 +13,7 @@ use xor_name::XorName;
 use std::time::Duration;
 static RECENT_ISSUE_DURATION: Duration = Duration::from_secs(60 * 15);
 
-static CONN_WEIGHTING: f32 = 2.0;
+static CONN_WEIGHTING: f32 = 1.0;
 static OP_WEIGHTING: f32 = 1.0;
 static KNOWLEDGE_WEIGHTING: f32 = 3.0;
 static DKG_WEIGHTING: f32 = 5.0;
--- a/sn_dysfunction/src/detection.rs
+++ b/sn_dysfunction/src/detection.rs
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::{error::Result, get_mean_of, DysfunctionDetection, OperationId};
+use crate::{error::Result, get_mean_of, std_deviation, DysfunctionDetection, OperationId};
 use std::collections::{BTreeMap, BTreeSet};
 use xor_name::XorName;
 
@@ -16,15 +16,17 @@ static RECENT_ISSUE_DURATION: Duration = Duration::from_secs(60 * 15);
 static CONN_WEIGHTING: f32 = 2.0;
 static OP_WEIGHTING: f32 = 1.0;
 static KNOWLEDGE_WEIGHTING: f32 = 3.0;
+static DKG_WEIGHTING: f32 = 5.0;
 
-// Ratio to mean scores should be over to be considered dys/sus
-static DYSFUNCTION_MEAN_RATIO: f32 = 9.5;
-static SUSPECT_MEAN_RATIO: f32 = 1.5;
+/// Z-score value above which a node is dysfunctional
+static DYSFUNCTIONAL_DEVIATION: f32 = 1.1;
 
 #[derive(Clone, Debug)]
 /// Represents the different type of issues that can be recorded by the Dysfunction Detection
 /// system.
 pub enum IssueType {
+    /// Represents a Dkg issue to be tracked by Dysfunction Detection.
+    Dkg,
     /// Represents a communication issue to be tracked by Dysfunction Detection.
     Communication,
     /// Represents a knowledge issue to be tracked by Dysfunction Detection.
@@ -36,6 +38,7 @@ pub enum IssueType {
 #[derive(Debug)]
 pub struct ScoreResults {
     pub communication_scores: BTreeMap<XorName, f32>,
+    pub dkg_scores: BTreeMap<XorName, f32>,
     pub knowledge_scores: BTreeMap<XorName, f32>,
     pub op_scores: BTreeMap<XorName, f32>,
 }
@@ -64,6 +67,7 @@ impl DysfunctionDetection {
         let mut communication_scores = BTreeMap::new();
         let mut knowledge_scores = BTreeMap::new();
         let mut op_scores = BTreeMap::new();
+        let mut dkg_scores = BTreeMap::new();
 
         let adults = self
             .adults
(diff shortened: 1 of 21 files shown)


sn_dysfunction/src/lib.rs
```diff
@@ -163,6 +171,31 @@ impl DysfunctionDetection {
...
+        trace!("Attempting to remove logged dkg session for {:?}", node_id,);
+        let mut has_removed = false;
+
+        if let Some(entry) = self.dkg_issues.get(node_id) {
+            let v = entry.value();
+
+            // only remove the first instance from the vec
+            let prev = v.write().await.pop_front();
+
+            if prev.is_some() {
```
maqi, line 186: just `has_removed = prev.is_some();` shall be enough?

sn_dysfunction/src/lib.rs
```diff
@@ -224,6 +257,25 @@ pub(crate) fn get_mean_of(data: &[f32]) -> Option<f32> {
     }
 }
 
+fn std_deviation(data: &[f32]) -> Option<f32> {
```
maqi, line 260: better to have a comment explaining what this function aiming for ?

sn_dysfunction/src/lib.rs
```diff
@@ -224,6 +257,25 @@ pub(crate) fn get_mean_of(data: &[f32]) -> Option<f32> {
...
+    match (get_mean_of(data), data.len()) {
+        (Some(data_mean), count) if count > 0 => {
+            let variance = data
+                .iter()
+                .map(|value| {
+                    let diff = data_mean - *value;
```
maqi, line 266: just to confirm `data_mean` is in type `f32` right? and the result of `diff` could be `negative`?

sn_node/src/bin/sn_node.rs
```diff
@@ -110,31 +110,29 @@ fn create_runtime_and_node() -> Result<()> {
...
-            Ok(result) => {
+        match run_node(config).await {
+            Ok(_) => {
                 info!("Node has finished running, no runtime errors were reported");
-                Ok(result)
             }
             Err(error) => {
-                warn!("Node instance finished with an error. Restarting node...");
-                Err(error)
+                warn!("Node instance finished with an error: {error:?}");
```
maqi, line 123: so now this `error` no longer trigger an error to be reported back further?

sn_node/src/node/api/dispatcher/background_tasks.rs
```diff
@@ -25,9 +25,9 @@ const PROBE_INTERVAL: Duration = Duration::from_secs(30);
 #[cfg(feature = "back-pressure")]
 const BACKPRESSURE_INTERVAL: Duration = Duration::from_secs(60);
 const SECTION_PROBE_INTERVAL: Duration = Duration::from_secs(300);
-const LINK_CLEANUP_INTERVAL: Duration = Duration::from_secs(120);
+const LINK_CLEANUP_INTERVAL: Duration = Duration::from_secs(10);
```
maqi, line 28: just to confirm this dramatic reduce of the interval is ok for the real usage as well?

sn_node/src/node/core/comm/mod.rs
```diff
@@ -374,27 +374,32 @@ impl Comm {
                         match watcher.await_change().await {
                             SendStatus::Sent => {
                                 successes += 1;
+                                debug!("successs");
                                 break; // we now move to checking next recipient send task..
                             }
                             SendStatus::Enqueued => {
+                                debug!("qqqqq");
```
maqi, line 381: to be removed? also for other couple of debug logs in this match block

sn_node/src/node/core/comm/mod.rs
```diff
@@ -444,54 +449,33 @@ impl Comm {
...
-            }
-        };
-
-        res
+        debug!("get or create peersesh::: none, so making oneee");
+        let link = Link::new(*peer, self.our_endpoint.clone(), self.msg_listener.clone());
+        let session = PeerSession::new(link);
+        debug!("get or create peersesh::: about to insert");
+        let _ = self.sessions.insert(*peer, session.clone());
+        debug!("get or create peersesh::: insertedddd");
```
maqi, line 465: remove or change the spelling?

sn_node/src/node/core/comm/mod.rs
```diff
@@ -444,54 +449,33 @@ impl Comm {
...
+        session
     }
 
     /// Any number of incoming qp2p:Connections can be added.
     /// We will eventually converge to the same one in our comms with the peer.
     async fn add_incoming(&self, peer: &Peer, conn: qp2p::Connection) {
         {
```
maqi, line 472: seems this extra layer to life span control can now be removed?

sn_node/src/node/core/comm/mod.rs
```diff
@@ -527,9 +512,11 @@ impl Comm {
             msg_id,
             recipient,
         );
-
+        debug!("oooooooother oeer get");
```
maqi, line 515: remove?
same to the couple of following

sn_node/src/node/core/comm/peer_session.rs
```diff
@@ -89,12 +89,14 @@ impl PeerSession {
...
     }
 
+    #[instrument(skip(self, msg_bytes, msg_priority))]
     pub(crate) async fn send(
         &self,
         msg_id: MsgId,
         msg_priority: i32,
         msg_bytes: Bytes,
     ) -> Result<SendWatcher> {
+        debug!("SENNNNDDD");
```
maqi, line 99: remove

sn_node/src/node/core/comm/peer_session.rs
```diff
@@ -160,7 +165,9 @@ impl PeerSession {
                 }
             }
 
+            debug!("sendloooooping");
```
maqi, line 168: remove

sn_node/src/node/core/messaging/handling/dkg.rs
```diff
@@ -280,6 +280,11 @@ impl Node {
     }
 
     pub(crate) async fn handle_dkg_failure(&self, failure_set: DkgFailureSigSet) -> Result<Cmd> {
+        // track those failed participants
+        for name in &failure_set.failed_participants {
+            self.log_dkg_issue(*name).await?;
```
maqi, line 285: wondering whether this `log_dkg_issue` function can take the list directly, to reduce the `async await`?

sn_dysfunction/src/detection.rs
```diff
@@ -133,6 +143,14 @@ impl DysfunctionDetection {
                 };
                 count
             }
+            IssueType::Dkg => {
+                let count = if let Some(entry) = self.dkg_issues.get(node) {
+                    entry.value().read().await.len()
+                } else {
+                    1
```
maqi, line 150: if `self.dkg_issues` doesn't have the record for the node, it shall means `no issue for that node`.
so the `count` shall return with 0 ?
why here return with 1 ?

sn_node/src/node/core/comm/mod.rs
```diff
@@ -88,46 +88,46 @@ impl Comm {
...
             let is_connected = session.is_connected().await;
 
             if !is_connected {
-                peers_to_cleanup.push(*peer);
+                if !retain_peers.contains(peer) {
+                    peers_to_cleanup.push(*peer);
+                }
+
+                dysfunction
+                    .track_issue(peer.name(), sn_dysfunction::IssueType::Communication)
```
Yoga07, line 111: We cleanup when we lose connection with a node right? Not sure if it's bad enough to be tracked? 🤔

sn_node/src/node/core/comm/mod.rs
```diff
@@ -374,27 +374,32 @@ impl Comm {
                         match watcher.await_change().await {
                             SendStatus::Sent => {
                                 successes += 1;
+                                debug!("successs");
```
Yoga07, line 377: Can be cleaned up? (similar logs below too)

sn_node/src/node/core/comm/mod.rs
```diff
@@ -527,9 +512,11 @@ impl Comm {
             msg_id,
             recipient,
         );
-
+        debug!("oooooooother oeer get");
         let peer = self.get_or_create(&recipient).await;
+        debug!("oooooooother oeer send");
```
Yoga07, line 517: More cleanup? 😄

sn_node/src/node/core/comm/peer_session.rs
```diff
@@ -105,11 +107,13 @@ impl PeerSession {
...
             msg_id,
             msg_bytes,
-            retries: 0,
+            retries: 3,
             reporter,
         };
 
+        debug!("msg wueueueuueeeeeeeeeeeeeeeeeeee writeeee");
         let _ = self.msg_queue.write().await.push(job, msg_priority);
+        debug!("msg wueueueuueeeeeeeeeeeeeeeeeee write done");
```
Yoga07, line 116: Clean ups :)

sn_dysfunction/src/detection.rs
```diff
@@ -170,22 +190,50 @@ impl DysfunctionDetection {
...
+
+        // now we store the z-score
+        let mut final_scores = BTreeMap::default();
+
+        for (name, score) in pre_z_scores {
+            let zscore = match (mean, std_dev) {
+                (Some(mean), Some(std_deviation)) => {
+                    let diff = score - mean;
+
+                    if diff < 1.0 {
```
maqi, line 222: could diff being a negative, i.e. score is less than mean ?

sn_dysfunction/src/detection.rs
```diff
@@ -319,6 +379,82 @@ mod tests {
             })
         }
 
+
+        #[test]
+        /// Test that givena  range of nodes and a few bad nodes, that we can reliably detect those nodes
```
maqi, line 384: given a range of
 reliably detect bad nodes ?

sn_dysfunction/src/detection.rs
```diff
@@ -319,6 +379,82 @@ mod tests {
...
+        /// TODO: right now this fails, as it appears our dysfunction barrier is TooDamnHigh
+        /// This _feels_ like a more correct test thanu sing the arbitrary counts we've had. Thoughts?
```
maqi, line 386: than using 

if the testing scenario is more realistic, than the arbitary counts shall be replaced?

sn_dysfunction/src/lib.rs
joshuef: Yeh, std dev goes both ways

sn_node/src/bin/sn_node.rs
joshuef: Yup, we don't throw out from this func here for node fails. We log the error and the async block completes so we go round again and restart the node

sn_node/src/node/api/dispatcher/background_tasks.rs
joshuef: Yeh, Ci has been fine with it. It doesn't seem to be causing any issue, and should keep nodes leaner...

sn_node/src/bin/sn_node.rs
```diff
@@ -110,31 +110,29 @@ fn create_runtime_and_node() -> Result<()> {
 
     // start a new runtime for a node.
     let rt = tokio::runtime::Runtime::new()?;
-    let res = rt.block_on(async move {
+    let _res = rt.block_on(async move {
```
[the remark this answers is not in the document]
joshuef: The result is bogus there. Always Ok cos we just log what actually comes out of the node. And then retry.

sn_node/src/node/core/comm/mod.rs
joshuef: If we're not connected, that's an issue for a node... I think. Won't happen too often. And should happen evenly unless bad, I thnk

sn_node/src/node/core/mod.rs
```diff
@@ -292,12 +294,31 @@ impl Node {
...
+    pub(crate) async fn log_knowledge_issue(&self, name: XorName) -> Result<()> {
+        trace!("Logging Knowledge issue in dysfunction");
+        self.dysfunction_tracking
+            .track_issue(name, IssueType::Knowledge)
+            .await
+            .map_err(Error::from)
+    }
+
+    /// Log a dkg issue (ie, an initialised but unfinished dkg round for a given participant)
+    pub(crate) async fn log_dkg_issue(&self, name: XorName) -> Result<()> {
```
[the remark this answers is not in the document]
joshuef: For me they're convenience issues to not have to import `IssueType` all over the node code

sn_dysfunction/src/detection.rs
joshuef: it doesn't matter as long as it's the same for all nodes, as we're taking an avg. (and 0 can make multplication trickier later)

sn_dysfunction/src/detection.rs
joshuef: yep. But in this case the node is behaving too well, so we don't need to worry

joshuef: ill add a comment

sn_dysfunction/src/detection.rs
joshuef: udpated

Yoga07 (approved):

maqi (approved):

PR #1242, opened 2022-06-15
feat: use node's section_key and own key for register

maqi:


sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -49,7 +50,12 @@ impl DataStorage {
 
     /// Store data in the local store
     #[instrument(skip(self))]
-    pub async fn store(&self, data: &ReplicatedData) -> Result<Option<StorageLevel>> {
+    pub async fn store(
+        &self,
+        data: &ReplicatedData,
+        pk: PublicKey,
+        keypair: Keypair,
```
bochaco, line 57: I think the pk and keypair arguments shall be renamed to something which tells what they are, current names don't tell anthing but their data type, so it's difficult to understand for the caller/user of the API/fn to know what to pass to it.

sn_node/src/node/core/messaging/handling/service_msgs.rs
```diff
@@ -175,8 +175,10 @@ impl Node {
                 // generate and sign spent proof share
                 let spent_proof_share = self.gen_spent_proof_share(&key_image, &tx).await?;
 
+                let keypair = Keypair::Ed25519(self.info.read().await.keypair.clone());
```
bochaco, line 178: Same comment as previous one, variables should help the reader know what the value they hold is rather than just knowing their data type, so in this case it's clear it's a keypair but what does that keypair represents in the context? that's what a variable shall be named after.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -325,6 +340,8 @@ mod tests {
         let used_space = UsedSpace::new(usize::MAX);
         let runtime = Runtime::new()?;
         let storage = DataStorage::new(path, used_space)?;
+        let pk = PublicKey::Bls(bls::SecretKey::random().public_key());
```
bochaco, line 343: Wat is this argument for? is it the owner? if so then this argument shall be named `owner`, otherwise the code is difficult to read/understand.

sn_node/src/node/core/data/storage/mod.rs
```diff
@@ -325,6 +340,8 @@ mod tests {
...
+        let keypair = Keypair::new_ed25519();
```
bochaco, line 344: Wat is this argument for? this argument shall be named with something that the reader can understand what is being passed around and not just data types.

sn_node/src/node/core/messaging/handling/mod.rs
```diff
@@ -659,11 +659,14 @@ impl Node {
                 } else {
                     let mut cmds = vec![];
 
+                    let pk = PublicKey::Bls(self.network_knowledge.section_key().await);
+                    let keypair = Keypair::Ed25519(self.info.read().await.keypair.clone());
```
bochaco, line 663: What are these pk and keypair values? are they the section_key and node's own keypair for signing something? if so then the variables shall be named accordingly so it's easier to read the code.

sn_node/src/node/core/messaging/handling/service_msgs.rs
```diff
@@ -279,11 +282,10 @@ impl Node {
 fn gen_register_cmd(
     key_image: &KeyImage,
     spent_proof_share: &SpentProofShare,
+    keypair: Keypair,
```
bochaco, line 285: Perhaps we can make this function to belong to the struct by passing `&self` so we don't need to pass the keypair around.

maqi: passing self in instead will make this function another `async fn` and has to `await` on it. 
may not a big issue, however just prefer to encapsulating function as `helper function` instead of `member function` when possible.

bochaco: what's the issue with awaiting it, you are anyway awaiting on `self.info.read().await.keypair.clone()` before calling this function....?

bochaco (approved):

PR #1253, opened 2022-06-20
Even slimmer single thread start pr

joshuef:

### DIFF
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -39,8 +39,7 @@ use bls::PublicKey as BlsPublicKey;
 use section_peers::SectionPeers;
 use secured_linked_list::SecuredLinkedList;
 use serde::Serialize;
-use std::{collections::BTreeSet, iter, net::SocketAddr, sync::Arc};
-use tokio::sync::RwLock;
+use std::{cell::RefCell, collections::BTreeSet, iter, net::SocketAddr, rc::Rc};
 use xor_name::{Prefix, XorName};
 
 /// The minimum age a node becomes an adult node.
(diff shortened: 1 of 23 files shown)


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -580,6 +577,9 @@ impl NetworkKnowledge {
             }
         }
 
+        // drop the borrow
+        drop(chain);
```
maqi, line 581: why for the `self.signed_sap`, there is no explicit `drop` after the `borrow`, but here need a drop for the `self.chain` ?
note the correspondent calling place is a `clone` call: `let chain = self.chain.borrow().clone();`

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -593,15 +593,18 @@ impl NetworkKnowledge {
             node_name,
             node_state.state()
         );
+
+        let the_chain = self.chain.borrow();
```
maqi, line 597: what if call with `self.section_chain()` to save the `drop` ?

sn_interface/src/network_knowledge/mod.rs
joshuef: cos we can't hold `chain` across the next `await` point, so we want to drop it.

sn_node/examples/routing_stress.rs
```diff
@@ -227,7 +227,7 @@ impl Network {
         };
 
         // TODO: Fix this: future returned by `add_node` is not `Send`
-        //let _handle = task::spawn(add_node(id, config, event_tx));
+        //let _handle = task::spawn_local(add_node(id, config, event_tx));
```
oetyng, line 230: Maybe this todo can be fixed now, by uncommenting line 230 and removing line 229 and 231?

oetyng: Just noticed its `routing_stress.rs`. Does this stress test even work anymore?

joshuef: Okay, I have a fix that gets it copmiling. I think we can address its relevance / CI worthiness apart from this PR.

sn_interface/src/network_knowledge/mod.rs
joshuef: ah so it is now!

maqi (approved):

davidrusu (approved):

davidrusu:
bors retry

sn_node/examples/routing_stress.rs
oetyng: > Maybe this todo can be fixed now, by uncommenting line 230 and removing line 229 and 231?

And what about this thing?

PR #1254, opened 2022-06-20
feat(dataquery): allow client to select adult

b-zee:
Augment the DataQuery structure to include specifying the adult that
should be queried by an elder. This allows traffic to be reduced
eventually in the context of IDA (Information Dispersal Algorithm or
erasure coding).


sn_client/src/api/queries.rs
```diff
@@ -94,6 +102,13 @@ impl Client {
             }
 
             attempt += 1.0;
+
+            // In the next attempt, try the next adult, further away.
+            query.adult_index += 1;
+            // There should not be more than a certain amount of adults holding copies of the data. Retry the closest adult again.
+            if query.adult_index >= data_copy_count() {
```
iancoleman, line 109: If we've tried 4 (or whatever) times and no result, this is a serious problem, right? It doesn't feel like looping back around to the first adult is appropriate here. There should be some report of 'stuff might be badly wrong' here...?

It could be the duplication has completely failed and the chunk is lost.

Or it could be the data never existed in the first place.

So I'm not exactly sure how to treat this but it feels like if we get to 'too many adults have been queried' the result should not be 'query the closest adult'...?

b-zee: I agree. However, what was the behavior before this? If the elder queried all (4) adults and the response was bad, then the client would just try again... I don't know if this new behavior would be a regression then.
Perhaps a new PR for a more thougtful/holistic approach is necessary to get the behavior right...

dirvine: @grumbach is working on a gossiping DKG. Perhaps grab a quick huddle with him on slack. We could gossip to all adults when we get to n-1 replicant attempts? Gossip failing means chunk gone from this section, including cache!

joshuef: There are situations where there may be more than 4 chunk holders... (with full nodes holding chunks, but also 4 non-full). The client has no real way to know, so would just keep trying until... XYZ. So perhaps there's a need for a "furthest adult to try", and the ability to react to errors from adults if "no adult at this xorindex" ?

joshuef (approved):

sn_client/src/api/queries.rs
joshuef: If we allow to "go further" than `data_copy_count`, we'll need to have such a feedback to errors to limit the index.

Otherwise I don't think there's anything inherently wrong in looping back over them. It could just be poor network conditions too and that clears up over time. So until we hit the general `timeout` for the query, we keep going as we know.

(Perhaps any error would report which indexes were tried?... might be a good one for a follow up PR, along with some tests to check this functionality against an existing network?)

iancoleman: > until we hit the general timeout for the query, we keep going as we know.

Ah ok I was under the impression this loop could be 'endless' but if it's bounded then not such a big deal.

I guess there's probably some details to sort through on this loop (in a followup pr), eg if the failure is due to network errors we'd want to loop back around, but if the failure is because we received an error saying the chunk doesn't exist we probably wouldn't want/need to loop.

My original concern doesn't seem like a blocker on this pr, feel free to resolve.

PR #1257, opened 2022-06-20
refactor(msg_type): move it to its own file

oetyng:
refactor(msg_type): move it to its own file

- Moves priority fns to service- and system msg.
- Moves deserialise of payload to wire_msg fn when getting priority.


sn_client/src/connections/messaging.rs
```diff
@@ -104,7 +104,7 @@ impl Session {
         };
 
         let msg_kind = AuthKind::Service(auth);
-        let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, dst_location)?;
+        let wire_msg = WireMsg::new_msg(msg_id, payload, msg_kind, 0, dst_location)?;
```
joshuef, line 107: Q @oetyng: can this open the door to an attack? Can clients not just put the prio as they want? (and risk getting blacklisted perhaps? , but better if that could be avoided altogether...)

oetyng: Ah you're right, that's not handled here. There's a fix that I can see immediately, but I'm trying to first remember what I had in mind when I did this (~3 months ago).. also not sure I really need this for the cmd ctrl PR..  *thinking*

oetyng: Ok, I solved it by changing the PR main focus :) 
So, I've skipped that change.

joshuef (approved):

PR #1269, opened 2022-06-23
tests(relocate): unit tests for JoiningAsRelocated

RolandSherwin:


sn_node/src/node/core/bootstrap/relocate.rs
```diff
@@ -265,3 +265,229 @@ impl JoiningAsRelocated {
...
+                gen_addr(),
+            );
+            let node_state = NodeState::joined(node.name(), node.addr, None);
+            let _ = from_sap.members.insert(node.name(), node_state.clone());
+            let signed_node_state = section_signed(&from_sk_set.secret_key(), node_state)?;
+            // to_sap
+            let (to_sap, to_sk_set) = generate_sap();
+
+            let bootstrap: Vec<SocketAddr> = to_sap.elders.iter().map(|(_, addr)| *addr).collect();
+            let (mut joining, _) = JoiningAsRelocated::start(
```
Yoga07, line 313: We could probably add an assert for the cmd returned here, to make sure the correct cmds are sent to the foreign section.

Yoga07: Also, could the returned cmd be handled?

sn_node/src/node/core/bootstrap/relocate.rs
```diff
@@ -265,3 +265,229 @@ impl JoiningAsRelocated {
...
+                    joining.dst_section_key
+                );
+            }
+            Ok(())
+        }
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn retry_and_redirect_should_return_none_on_latest_sap() -> Result<()> {
```
Yoga07, line 362: We could probably combine the three tests? I guess it's about sending the correct PK, the incorrect PK and empty senders list? We can deduplicate a lot of code here IMO :)

sn_node/src/node/core/bootstrap/relocate.rs
```diff
@@ -265,3 +265,229 @@ impl JoiningAsRelocated {
...
+                .elders
+                .iter()
+                .next()
+                .ok_or_else(|| eyre!("Elder will be present"))?;
+            let _ = joining
+                .handle_join_response(response.clone(), *sender)
+                .await;
+
+            // with outdated key, but the new_recipients is empty
+            joining.dst_section_key = bls::SecretKey::random().public_key();
```
Yoga07, line 457: It's a bit unclear what our goal here is. Is there a way to differentiate between a failure from outdated keys and empty recipients?

sn_node/src/node/core/bootstrap/relocate.rs
RolandSherwin: Combined it with the other test case! :+1:

sn_node/src/node/core/bootstrap/relocate.rs
RolandSherwin: Added more comments.

sn_node/src/node/core/bootstrap/relocate.rs
RolandSherwin: There is a test called `membership_churn_starts_on_join_request_from_relocated_node` which basically handles the `JoinAsRelocatedRequest` msg. The msg triggers `handle_join_as_relocated_request` which results in membership change and that is verified in the test case.

Yoga07 (approved):

PR #1273, opened 2022-06-24
feat: deposit owned dbcs

jacderida:
- 554c9d688 **chore: upgrade sn_dbc to 3.3.0**

  This version contains an update for converting an owned DBC to a bearer DBC.

- 23802f8e3 **feat!: extend wallet_deposit for owned dbcs**

  BREAKING CHANGE: the `wallet_deposit` function now provides an `Option<bls::SecretKey>` argument for
  providing a secret key when depositing an owned DBC.

  The owned DBC is changed to a bearer by providing the secret key, and the resulting DBC is then
  stored in the wallet. This means you don't need to provide the secret key at reissue time, where
  there is an issue with mapping input DBCs to secret keys.

  A couple of misc changes:
  * Extra test cases were added here to cover assigning the name to the deposit.
  * More instances of "Wallet" were converted to "wallet", because "wallet" is not a proper noun.

  As an unrelated change, this commit also provides more test coverage for `keys show` command. This
  isn't a hugely important command, but I just wanted to ensure we don't break support for BLS keys.

  These are integration tests because you really need to parse the console output, which you can't do
  with unit tests.

  This should probably have been its own commit, but it was accidentally added via an amend, then the
  API changes were also amended on top, so it was too difficult to break out.

- 67006eb2e **feat!: serialize to bls keys in util functions**

  Utility functions were recently added to the API for serializing to the `Keypair` type. This was
  changed to serialize directly to BLS to make it easier for the CLI to deal directly with BLS keys.
  Soon we will be refactoring the `Keypair` type to have a different use case and things like
  `sn_client` would be refactored to directly work with BLS keys. This is a little step in that
  direction.

  There was a utility function added to `sn_interface` to create a `Keypair` from a hex-based BLS key
  because we still need to use the `Keypair` at this point in time.

- f7940c5cd **feat!: remove use of xorurl with keys command**

  The use of BLS keys means the XorUrl that was displayed to the user will no longer be usable, since
  the BLS key is a 48-byte structure, and not 32 bytes like the Dalek key. We will come back and
  address this issue later, possibly re-introducing the `SafeKey` type in some kind of different form.

  The functionality and corresponding test cases for the `cat` and `dog` command that were working
  with the `SafeKey` data type were removed.

  The test cases for the `keys` commands were elaborated here to check the output of both the pretty
  print and the json formats and to make sure the CLI prints out a matching keypair.

- 69079d698 **feat: extend cli wallet deposit for owned dbcs**

  The CLI is now extended to support the deposit of owned DBCs.

  The `deposit` command will check if the supplied DBC is owned, and if it is, it will check to see if
  the `--secret-key` argument is present and use that. If that argument isn't present, it will attempt
  to use the secret key that's configured for use with the CLI, i.e., the `keys create --for-cli`
  command.

  The `reissue` command was also extended to provide an `--owned` flag, which when used, will reissue
  an owned DBC using the public key configured for use with the CLI. This argument is mutually
  exclusive with the `--public-key` argument, which will reissue the DBC using a specified key.

  So we could offer the user a suggestion when a supplied secret key didn't match, this also involved
  making a little extension to the API, to return a specific type of error. We will need to modify
  `sn_dbc` to return a specific error type for this too, so we can avoid checking the string content
  of the error message, but this will be covered on a separate PR.

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -62,7 +62,19 @@ impl Safe {
             dbc.clone()
         } else if let Some(sk) = secret_key {
             let mut owned_dbc = dbc.clone();
-            owned_dbc.to_bearer(&sk)?;
+            match owned_dbc.to_bearer(&sk) {
+                Ok(()) => {}
+                Err(e) => {
+                    // TODO: Add a specific error type to `sn_dbc` for the mismatched key so we can
+                    // avoid this string check.
+                    if e.to_string()
+                        .contains("supplied secret key does not match the public key")
+                    {
+                        return Err(Error::DbcDepositInvalidSecretKey());
+                    }
+                    return Err(Error::DbcDepositError(e.to_string()));
+                }
+            }
             owned_dbc
         } else {
             return Err(Error::DbcDepositError(
@@ -516,6 +528,34 @@ mod tests {
                 Ok(())
             }
             Err(_) => Err(anyhow!("This test should use a DbcDepositError".to_string())),
+        }
+    }
+
+    #[tokio::test]
+    async fn test_wallet_deposit_with_owned_dbc_with_invalid_secret_key() -> Result<()> {
+        let safe = new_safe_instance().await?;
+        let wallet_xorurl = safe.wallet_create().await?;
+        let sk = bls::SecretKey::random();
+        let sk2 = bls::SecretKey::random();
+        let pk = sk.public_key();
+
+        let dbc = new_dbc(DBC_WITH_12_230_000_000)?;
+        safe.wallet_deposit(&wallet_xorurl, Some("my-dbc"), &dbc, None)
+            .await?;
+        let owned_dbc = safe
+            .wallet_reissue(&wallet_xorurl, "2.35", Some(pk))
+            .await?;
+        let result = safe
+            .wallet_deposit(&wallet_xorurl, Some("owned-dbc"), &owned_dbc, Some(sk2))
+            .await;
+        match result {
+            Ok(_) => Err(anyhow!(
+                "This test case should result in an error".to_string()
+            )),
+            Err(Error::DbcDepositInvalidSecretKey()) => Ok(()),
+            Err(_) => Err(anyhow!(
+                "This test should use a DbcDepositInvalidSecretKey error".to_string()
+            )),
         }
     }
 
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -18,15 +18,16 @@ use sn_dbc::{rng, Owner, OwnerOnce, TransactionBuilder};
 use sn_interface::types::Token;
 use std::collections::{BTreeMap, BTreeSet};
 
-// Type tag used for the Wallet
 const WALLET_TYPE_TAG: u64 = 1_000;
 
-/// Set of spendable DBC's mapped to their friendly name
-/// as defined/chosen by the user when depositing DBC's into a Wallet.
+/// Set of spendable DBCs mapped to their friendly name as defined/chosen by the user when
+/// depositing DBCs into a wallet.
 pub type WalletSpendableDbcs = BTreeMap<String, (Dbc, EntryHash)>;
 
 impl Safe {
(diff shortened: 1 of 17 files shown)


sn_cli/tests/cli_wallet.rs
```diff
@@ -176,6 +177,262 @@ fn wallet_deposit_should_fail_with_suggestion_when_path_is_directory() -> Result
...
+        ],
+        Some(0),
+    )?;
+
+    safe_cmd(
+        [
+            "wallet",
+            "deposit",
+            "--name",
+            "my-first-dbc",
```
bochaco, line 223: perhaps name it differently from the first one in tests like this? from a functional point of view it doesn't matter but it may be hiding bugs, so maybe name it `reissued-dbc` or something like that?

sn_api/src/app/wallet.rs
```diff
@@ -36,22 +37,50 @@ impl Safe {
...
         // TODO: check the input DBCs were spent and all other sort of verifications,
         // perhaps all optional, and we may want a separate API to also do these verifications
         // for the user to perform them without depositing the DBC into a wallet.
+        let dbc_to_deposit = if dbc.is_bearer() {
+            if secret_key.is_some() {
+                return Err(Error::DbcDepositError(
+                    "A secret key should not be supplied when depositing a bearer DBC".to_string(),
+                ));
+            }
+            dbc.clone()
```
bochaco, line 62: You don't need to clone it

bochaco (commented):
Awesome job @jacderida

sn_cli/src/subcommands/dog.rs
```diff
@@ -119,16 +119,8 @@ pub async fn dog_commander(cmd: DogCommands, output_fmt: OutputFmt, safe: &Safe)
...
-                    xorname,
-                    resolved_from,
-                } => {
-                    println!("Resolved from: {}", resolved_from);
-                    println!("= SafeKey =");
-                    println!("XOR-URL: {}", xorurl);
-                    println!("XOR name: 0x{}", xorname_to_hex(xorname));
-                    println!("Native data type: SafeKey");
+                SafeData::SafeKey { .. } => {
+                    unimplemented!("The SafeKey data type is not supported at the moment");
```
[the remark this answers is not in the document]
bochaco: ....a panic means your app/code didn't handled an error/situation, which is not the case here where we know what happened and we are just not supporting it.

sn_cli/src/subcommands/wallet.rs
```diff
@@ -119,17 +133,54 @@ pub async fn wallet_commander(
...
+                }
+            } else {
+                None
+            };
+            let name = match safe
+                .wallet_deposit(&wallet_url, name.as_deref(), &dbc, sk)
+                .await
+            {
+                Ok(name) => name,
+                Err(e) => match e {
```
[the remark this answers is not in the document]
jacderida: I'm not sure that having the `match` inside `map_err` is making this code more readable, and I'm having trouble getting it working. Can you show me what you had in mind?

sn_api/src/app/wallet.rs
jacderida: This doesn't compile without the clone.

sn_cli/src/subcommands/wallet.rs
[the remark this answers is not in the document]
bochaco: not a big deal, but something like this:
```
            let name = safe
                .wallet_deposit(&wallet_url, name.as_deref(), &dbc, sk)
                .await
                .map_err(|err| match err {
                    Error::DbcDepositInvalidSecretKey() => {
                        eyre!("The supplied secret key did not match the public key for this DBC.")
                            .suggestion(
                                "Please run the command again with the correct key for the \
                                --secret-key argument.",
                            )
                    }
                    other => {
                        eyre!(other)
                    }
                })?;
```

sn_api/src/app/wallet.rs
bochaco: ah yeah...you are right, nw!

bochaco (approved):

bochaco (approved):

PR #1143, opened 2022-04-26
feat(spentbook): first version of Spentbook messaging and storage

bochaco:
This is a follow up PR (3rd PR) to PR #1097 and PR #1105, introducing first and basic implementation of Spentbook messaging, storage and client API:
- Storage is implemented using Register as the underlying data type. To be changed when actual Spentbook native data type is put in place.
- First version of sn_client API for Spentbook messages.
- sn_client::spent_proof_shares API to fetch spent proof shares.
- sn_client::spend_dbc API to request Elders to store the spent proof shares.
- Serialise SpentProofShare to store it as an entry in the underlying Register.

The CLI `wallet reissue` and `wallet deposit` commands are currently compatible with the DBC's generated and printed out by the [mint-repl](https://github.com/maidsafe/sn_dbc#mint-repl-example) and [DBC Playground](https://github.com/maidsafe/sn_dbc_examples) output, once a DBC is generated by those example apps, it can be deposited in a Wallet and check its balance, e.g. the following commands deposit the genesis DBC into a Wallet on Safe, check its total balance, and reissue a DBC from the Wallet for 3.77 safecoins:
```
$ safe wallet create
Wallet created at: "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y"

$ safe wallet balance safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Wallet at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y" has a total balance of 0.000000000 safecoins

$ safe wallet deposit --name my-first-dbc safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y c4a1953dfa419234b8ab9a...
Spendable DBC deposited with name 'my-first-dbc' in Wallet located at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y"

$ safe wallet balance safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y
Wallet at "safe://hyryynyen8rzu4umhifw6j7chbr5wthr1osfb3qcc78n1498zrpd8wsm4y5my84y" has a total balance of 18446744073.709551615 safecoins

[...]

PR #1274, opened 2022-06-24
feat(node): generate the genesis DBC when launching first node

bochaco:
This is a follow up PR (5th PR) to PR https://github.com/maidsafe/safe_network/pull/1235, PR https://github.com/maidsafe/safe_network/pull/1097, PR https://github.com/maidsafe/safe_network/pull/1105, and PR https://github.com/maidsafe/safe_network/pull/1143.

The Genesis DBC is generated and owned by the genesis PK of the network's section chain, i.e. the very first section key in the chain. 
The first node mints the genesis DBC (as a bearer DBC) and stores it in a file named `genesis_dbc`, located at it's configured root dir, e.g. when using the `sn_launch_tool` the genesis DBC will be stored at `~/.safe/node/local-test-network/sn-node-genesis/genesis_dbc`.

In this implementation the amount owned by the Genesis DBC is set to `4,525,524,120,000,000,000` individual units.
As per current state of discussions (still in progress), at the inception of the Safe network a total supply of 4,525,524,120 whole tokens will be created, each whole token can be subdivided 10^9 times, thus creating a total of 4,525,524,120,000,000,000 available units.

E.g. the following CLI commands deposit the generated genesis DBC into a wallet on Safe, and check its total balance:
```
$ safe wallet create
Wallet created at: "safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o"

$ safe wallet balance safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o
Wallet at "safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o" has a total balance of 0.000000000 safecoins

$ safe wallet deposit safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o --name genesis --dbc ~/.safe/node/local-test-network/sn-node-genesis/genesis_dbc
Spendable DBC deposited with name 'genesis' in wallet located at "safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o"

$ safe wallet balance safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o
Wallet at "safe://hyryynyizz7zfcmeyeu3ety9mp549nh5r1z7ro9t1ebjib9drcmk15d59fyb6o" has a total balance of 4525524120.000000000 safecoins
```

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -53,6 +53,7 @@ impl Safe {
         // TODO: check the input DBCs were spent and all other sort of verifications,
         // perhaps all optional, and we may want a separate API to also do these verifications
         // for the user to perform them without depositing the DBC into a wallet.
+
         let dbc_to_deposit = if dbc.is_bearer() {
             if secret_key.is_some() {
                 return Err(Error::DbcDepositError(
--- a/sn_cli/src/subcommands/wallet.rs
+++ b/sn_cli/src/subcommands/wallet.rs
@@ -14,6 +14,7 @@ use crate::operations::config::Config;
 use bls::{PublicKey, SecretKey};
 use color_eyre::{eyre::eyre, Help, Result};
 use sn_api::{Error, Safe};
+use sn_dbc::Dbc;
 use std::path::Path;
 use structopt::StructOpt;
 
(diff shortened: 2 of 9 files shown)


maqi:
The PR description is very good which explains the scope and aim of the change in detail.
Will it better to put these words into ReadMe of the crate or somewhere in the dbc module as comment?

Also, have two points in doubt:
1, seems according to the change, there is genesis file name of GENESIS_DBC_FILENAME get appended automatically for the first node.Then in the example command of `safe wallet deposit`, this file will be read in, right?
2, looks like the deposit has to deposit ALL tokens into the wallet at once, right?
3, If trying to deposit the same genesis tokens into another wallet, that attempt shall get failed, right?

bochaco:
> The PR description is very good which explains the scope and aim of the change in detail. Will it better to put these words into ReadMe of the crate or somewhere in the dbc module as comment?

Makes sense, I'll see where I can put it, perhaps in the CLI User Guide, will add a commit with it
 
> 1, seems according to the change, there is genesis file name of GENESIS_DBC_FILENAME get appended automatically for the first node.Then in the example command of `safe wallet deposit`, this file will be read in, right? 

Yes correct, that's one off the CLI argument recently implemented, instead of having to paste the DBC content to the command you can have it reaad it from a file.

> 2, looks like the deposit has to deposit ALL tokens into the wallet at once, right? 

Correct, the balance is owned by the DBC your are depositing, if you want to split it you have to hen reissue from it into other output DBCs. Right now there are not many options in CLI to split it in more than two but eventually we can allow that.

> 3, If trying to deposit the same genesis tokens into another wallet, that attempt shall get failed, right?

Depositing it won't fail, depositing DBCs in wallets is just having the DBC content stored in the wallet, which is private data the user owns, so you can deposit it in many places with no issue. What should (eventually) fail is if two wallets are then trying to reissue from the same DBC more than once, but that is not yet supported due to the limitation of our temporary/hacky implementation of spentbook which doesn't prevent double spent yet.

maqi:
>depositing DBCs in wallets is just having the DBC content stored in the wallet, which is private data the user owns

Would like to confirm that this also means when trying to deposit a genesis DBC with genesis_key_A to a wallet which created in a network having the genesis_key_B, the operation will succeed.
However, at the point of trying to spent it (i.e. reissuing), the operation will then fail, right?

bochaco:
> Would like to confirm that this also means when trying to deposit a genesis DBC with genesis_key_A to a wallet which created in a network having the genesis_key_B, the operation will succeed. However, at the point of trying to spent it (i.e. reissuing), the operation will then fail, right?

Correct, the DBC is just a certificate you keep wherever you want, you don't even need to deposit it on a wallet, or even keep it stored in the network, you can keep it on a piece of paper if you want. But when you want to spend it that's when the network needs to check if the certificate and signatures you provide are valid, so it's when you reissue it that the network will perform the verifications to sign a new entry on the spentbook.

maqi (approved):

PR #1278, opened 2022-06-28
chore: Tweak dysf interval, reducing to report on issues more rapidly

joshuef:
If not, we can only ever propose one node for a membership change
(lost), every 30s... which may not succeed under churn...


Yoga07 (approved):

testnet/bin.rs
```diff
@@ -46,7 +46,7 @@ const SAFE_NODE_EXECUTABLE: &str = "sn_node.exe";
 
 const BASE_TRACING_DIRECTIVES: &str = "testnet=info,sn_launch_tool=debug";
 const NODES_DIR: &str = "local-test-network";
-const DEFAULT_INTERVAL: &str = "10000";
+const DEFAULT_INTERVAL: &str = "100";
```
b-zee, line 49: Was this changed on purpose? I know that the interval can generally be lower, but not sure if this is only reasonable on more powerful machines.

maqi (approved):

testnet/bin.rs
joshuef: Yeh that was intentional. For me locally (and on the prior CI runs) this seems to have been working well

PR #1288, opened 2022-06-30
fix(comm): re-enable send job retries on transient errors

davidrusu:
Previous work on PeerSessions had removed SendJob retries, this PR adds that functionality back.

### DIFF
--- a/sn_node/src/comm/peer_session.rs
+++ b/sn_node/src/comm/peer_session.rs
@@ -27,6 +27,7 @@ use tokio::{sync::mpsc, time::Instant};
 // TODO: temporarily disable priority while we transition to channels
 // type Priority = i32;
 
+const MAX_SENDJOB_RETRIES: usize = 10;
 const DEFAULT_DESIRED_RATE: f64 = 10.0; // 10 msgs / s
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


sn_node/src/comm/peer_session.rs
```diff
@@ -184,7 +188,12 @@ impl PeerSessionWorker {
         info!("Finished peer session shutdown");
     }
 
-    async fn send(&mut self, job: SendJob) -> SessionStatus {
+    async fn send(&mut self, mut job: SendJob) -> SessionStatus {
+        if job.retries > MAX_SENDJOB_RETRIES {
+            job.reporter.send(SendStatus::MaxRetriesReached);
+            return SessionStatus::Ok;
```
davidrusu, line 194: Debatable whether we should be terminating this peer connection or if we should just continue.

On one hand the errors may be specific to the SendJob, i.e. suppose the message is just too big to stuff into the pipe.

On the other, we gave this peer connection 10 attempts to complete the job so perhaps it's worth terminating and having higher level code try and re-create the session.

Thoughts?

joshuef: I think a terminate can make sense here. 10 attempts is a good amount. This _shouldnt_ be happening all the time, so it seems fine to me.

PR #1290, opened 2022-06-30
fix(join): do not leave loop running on timeout

oetyng:
Fixes the case when a node start times out and a new node is started
while the previous one is still looping in some task.

This used to cause `address already in use` errors.

Even though a random address would have solved that,
it's better to ensure that the running node actually is dropped, before
starting a new one.

(NB: Might well be better and cleaner solutions to this, that could be discussed in this PR)

### DIFF
--- a/sn_node/src/bin/sn_node.rs
+++ b/sn_node/src/bin/sn_node.rs
@@ -53,6 +53,7 @@ use tracing_subscriber::filter::EnvFilter;
 
 #[cfg(not(feature = "tokio-console"))]
 const MODULE_NAME: &str = "sn_node";
+const JOIN_TIMEOUT_SEC: u64 = 30;
 const BOOTSTRAP_RETRY_TIME_SEC: u64 = 15;
 
 fn main() -> Result<()> {
@@ -310,9 +311,11 @@ async fn run_node(config: Config) -> Result<()> {
 
     let log = format!("The network is not accepting nodes right now. Retrying after {BOOTSTRAP_RETRY_TIME_SEC} seconds");
 
+    let join_timeout = Duration::from_secs(JOIN_TIMEOUT_SEC);
     let bootstrap_retry_duration = Duration::from_secs(BOOTSTRAP_RETRY_TIME_SEC);
+
     let (node, mut event_stream) = loop {
-        match NodeApi::new(&config, bootstrap_retry_duration).await {
+        match NodeApi::new(&config, join_timeout).await {
             Ok(result) => break result,
             Err(NodeError::CannotConnectEndpoint(qp2p::EndpointError::Upnp(error))) => {
                 return Err(error).suggestion(
--- a/sn_node/src/bin/sn_node.rs
+++ b/sn_node/src/bin/sn_node.rs
@@ -53,8 +53,8 @@ use tracing_subscriber::filter::EnvFilter;
 
 #[cfg(not(feature = "tokio-console"))]
 const MODULE_NAME: &str = "sn_node";
-const JOIN_TIMEOUT_SEC: u64 = 30;
-const BOOTSTRAP_RETRY_TIME_SEC: u64 = 15;
+const JOIN_TIMEOUT_SEC: u64 = 10;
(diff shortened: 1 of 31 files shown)


sn_node/src/bin/sn_node.rs
```diff
@@ -310,9 +310,11 @@ async fn run_node(config: Config) -> Result<()> {
 
     let log = format!("The network is not accepting nodes right now. Retrying after {BOOTSTRAP_RETRY_TIME_SEC} seconds");
 
+    let join_timeout = Duration::from_secs(2 * BOOTSTRAP_RETRY_TIME_SEC);
     let bootstrap_retry_duration = Duration::from_secs(BOOTSTRAP_RETRY_TIME_SEC);
+
     let (node, mut event_stream) = loop {
-        match NodeApi::new(&config, bootstrap_retry_duration).await {
+        match NodeApi::new(&config, join_timeout).await {
```
joshuef, line 317: Why have we a new variable that's twice the intended retry time here?

----

Previously:

if `bootstrap_retry_duration` was hit in the internal `timeout`, we errored. We then wait `bootstrap_retry_duration` again before trying.

Here:

If after `2 * bootstrap_retry_duration` we haven't  joined, we error (out of the loop 👍 ). And then wait `bootstrap_retry_duration` once more? Is that correct?

Is there a need for a new variable twice as large? (I appreciate naming can be improved here regardless). What's the reason for `2*` ?

sn_node/src/node/core/bootstrap/join.rs
```diff
@@ -152,14 +162,22 @@ impl<'a> Join<'a> {
...
+
         self.send_join_requests(join_request.clone(), &recipients, section_key, false)
             .await?;
 
         // Avoid sending more than one duplicated request (with same SectionKey) to the same peer.
         let mut used_recipient_saps = UsedRecipientSaps::new();
 
         loop {
-            let (response, sender) = self.receive_join_response().await?;
+            join_timeout = join_timeout
```
joshuef, line 174: Can you add a comment just clarifying the intended function here. That we error out if we've exceeded the `join_timeout` 🙇

sn_node/src/node/core/bootstrap/join.rs
```diff
@@ -434,8 +452,21 @@ impl<'a> Join<'a> {
...
     #[tracing::instrument(skip(self))]
-    async fn receive_join_response(&mut self) -> Result<(JoinResponse, Peer)> {
-        while let Some(event) = self.incoming_msgs.recv().await {
+    async fn receive_join_response(
+        &mut self,
+        mut join_timeout: Duration,
+    ) -> Result<(JoinResponse, Peer)> {
+        let mut timer = Instant::now();
+
+        while let Some(event) = tokio::time::timeout(join_timeout, self.incoming_msgs.recv())
```
joshuef, line 461: same here please w/r/t comment

sn_node/src/bin/sn_node.rs
oetyng: It just seemed as if a varable used for one thing (bootstrap retry interval) had been used for another (join timeout).
As I've seen the timeout error now and then, I thought 30s could be used  and seemed like a valid timeout.
Better as a new const but didn't seem overly important considering the already existing reuse of the const for separate things.

oetyng: Added separate const.

joshuef: Makes sense. I think that's probably my actual confusion here, the initial confused naming.

`bootstrap_retry_duration` sounds like _over this time, we'll retry. That wasn't accurate before... it was twice as long from initial contact attempt to retry, and now it's three times as long.

Which feels off, but the source of this isn't this PR I think, so feel free to ignore this. I'm not sure I have any ideas of to improve the naming / use there (and as `bootstrap_retry_duration` isn't a config option atm... I think it doesn't really matter).

joshuef (approved):

joshuef:
merged via #1292

PR #1303, opened 2022-07-05
chore(client): add binary to query chunk at adults

b-zee:


sn_client/src/bin/query-adult/main.rs
```diff
@@ -0,0 +1,185 @@
...
+fn self_encryptable_file() -> Bytes {
+    let mut file = BytesMut::with_capacity(MIN_ENCRYPTABLE_BYTES);
+    unsafe {
+        file.set_len(MIN_ENCRYPTABLE_BYTES);
+    }
+
+    thread_rng().fill(&mut file[..]);
+
+    file.freeze()
+}
```
davidrusu, line 176: `MIN_ENCRYPTABLE_BYTES` is small enough that we can just allocate the buffer on the stack and copy into a bytes avoid this unsafe block:

```rust

/// Generate random bytes with minimal self-encryption size
fn self_encryptable_file() -> Bytes {
    let mut file = [0u8; MIN_ENCRYPTABLE_BYTES];
    thread_rng().fill(&mut file[..]);

    Bytes::copy_from_slice(&file)
}
```

b-zee: Thanks @davidrusu, should have thought twice about using `unsafe` in the codebase! Still think it's normal from my embedded experience... 😃

davidrusu (approved):
Looks good!

davidrusu:
@b-zee Cargo toml is broken on merge, mind rebasing?

PR #1316, opened 2022-07-06
remove some dependencies and slightly improve build times

davidrusu:
Dependencies dropped:

```
ansi_term 0.11
env_logger 0.7
fuchsia-cprng
num family:
- num-bigint
- num-complex
- num-iter
- num-rational
quickcheck
rand, rand_core 0.4
rand_core 0.3
rdrand
strsim
vec_map
```

But we are still building two versions of clap (v3 and v2), criterion v0.4.0 should be released soon which will allow us to to drop clap v2.


dirvine (approved):
All helps

b-zee:
Nice work.

The test failing on MacOS is the same test I have seen failing with things @joshuef was working on, not sure if related.

davidrusu:
bors cancel

PR #1317, opened 2022-07-06
feat(api): perform verification of input TX and spentproofs when depositing or reissuing a DBC

bochaco:
This is a follow up PR (9th PR) to PR https://github.com/maidsafe/safe_network/pull/1351, PR https://github.com/maidsafe/safe_network/pull/1226, PR https://github.com/maidsafe/safe_network/pull/1301, PR https://github.com/maidsafe/safe_network/pull/1274, PR https://github.com/maidsafe/safe_network/pull/1235, PR https://github.com/maidsafe/safe_network/pull/1097, PR https://github.com/maidsafe/safe_network/pull/1105, and PR https://github.com/maidsafe/safe_network/pull/1143.

- Verify that the DBC to deposit is valid, i.e. there is a matching transaction provided for each SpentProof, and check the SpentProofs were validly signed by a known section key.
- Verify that input TX and SpentProofs are valid in outputs DBCs when reissuing them, this includes checking the SpentProofs were validly signed by a known section key.
- Adapting sn_cli tests to reuse the same pattern implemented in sn_api tests with PR https://github.com/maidsafe/safe_network/pull/1351 .

Note that in all these verifications we don't yet check the DBC was not already spent when depositing into a wallet. Perhaps we want a separate API to do these verifications for the user to perform them even without depositing the DBC into a wallet.


sn_api/src/app/wallet.rs
```diff
@@ -28,6 +30,36 @@ pub const WALLET_TYPE_TAG: u64 = 1_000;
 /// depositing DBCs into a wallet.
 pub type WalletSpendableDbcs = BTreeMap<String, (Dbc, EntryHash)>;
 
+// Verifier required by sn_dbc API to check a SpentProof
```
jacderida, line 33: Minor style issue: could we use doc comments for these?

sn_api/src/app/wallet.rs
```diff
@@ -83,17 +111,29 @@ impl Safe {
...
-        let amount = dbc_to_deposit
-            .amount_secrets_bearer()
-            .map(|amount_secrets| Token::from_nano(amount_secrets.amount()))?;
+        // TODO: check the DBC was not spent already. Perhaps we want a separate API to do these
+        // verifications for the user to perform them without depositing the DBC into a wallet.
 
-        let safeurl = self.parse_and_resolve_url(wallet_url).await?;
+        // Verify that the DBC to deposit is valid. This verifies there is a matching transaction
+        // provided for each SpentProof, although this does not check if the DBC has been spent.
+        let proof_key_verifier = SpentProofKeyVerifier {
```
jacderida, line 119: I feel like we should probably have at least one new test case for the addition of the verification, where the verification would fail. We would have coverage of the passing case for a normal valid.

sn_api/src/app/wallet.rs
```diff
@@ -418,9 +458,12 @@ impl Safe {
                 .add_spent_transaction(tx);
         }
 
-        // TODO: Perform the verification of the transaction and spentproofs for input DBCs,
+        // Perform verifications of input TX and spentproofs,
         // as well as building the output DBCs.
-        let dbcs = dbc_builder.build_without_verifying()?;
+        let proof_key_verifier = SpentProofKeyVerifier {
```
jacderida, line 463: Similarly for this one, I would expect a new test case with the addition of the new code.

sn_cli/tests/cli_cat.rs
```diff
@@ -326,11 +334,11 @@ fn calling_safe_cat_wallet() -> Result<()> {
             wallet_xorurl
         )))
         .stdout(predicate::str::contains("my-first-dbc"))
-        .stdout(predicate::str::contains("12.230000000"))
+        .stdout(predicate::str::contains(balance.to_string()))
         .stdout(predicate::str::contains(format!(
             "{}...{}",
-            &DBC_WITH_12_230_000_000[..8],
-            &DBC_WITH_12_230_000_000[DBC_WITH_12_230_000_000.len() - 8..]
+            &dbc_hex_string[..8],
```
jacderida, line 340: Could these go into a helper function in the command test utilities and the name of the function would explain what it's doing and hide the specific range selection?

sn_cli/tests/cli_wallet.rs
```diff
@@ -7,14 +7,26 @@
...
 use predicates::prelude::*;
+use sn_api::{test_helpers::get_next_bearer_dbc, Token};
 use sn_cmd_test_utilities::util::{
-    parse_keys_create_output, parse_wallet_create_output, safe_cmd, safe_cmd_stdout,
-    DBC_WITH_12_230_000_000,
+    get_random_string, parse_keys_create_output, parse_wallet_create_output, safe_cmd,
+    safe_cmd_stdout,
 };
 
+async fn get_bearer_dbc_on_file(tmp_data_dir: &TempDir) -> Result<(ChildPath, Token)> {
```
jacderida, line 22: We could maybe put this in the test utils crate also, since it appears in the wallet test cases too?

Makefile
```diff
@@ -17,7 +17,7 @@ gha-build-x86_64-unknown-linux-musl:
 	rustup target add x86_64-unknown-linux-musl
 	cargo build --release --target x86_64-unknown-linux-musl --bin sn_node
 	cargo build --release --target x86_64-unknown-linux-musl \
-		--no-default-features --features testing --bin safe
+		--no-default-features --bin safe
```
jacderida, line 20: To be honest I'm not sure what this `testing` feature is/was doing. Let's just see what happens. We can always add it back in if need be.

bochaco: ok, sounds good, I'll also remove thee `--no-default-features` as I doubt we need it.

sn_api/src/app/wallet.rs
bochaco: we could, I usually only do that for public vars/structs/methods as I thought they won't be otherwise published anywhere, i mean if the item is not public?

sn_api/src/app/wallet.rs
bochaco: makes sense, i'll work on a test case for it

sn_cli/tests/cli_wallet.rs
bochaco: this is only used in the wallet test cases, or what do you mean?

bochaco: ...ah you meant in cli_cat tests, I've done that.

jacderida (approved):

PR #1319, opened 2022-07-07
fix: passing the churn test

maqi:
This PR contains the work to passing the churn test.
There are mainly two fixes:
1, Only trigger data reorganization when there is membership update.
   Previously, data reorganzation get undertaken whenever there is
   incoming message. Which result in a looping of messaging among
   nodes.
2, Only broadcast result when the QueryResponse is not an error.
   Previously, this will cause the client thinking the whole query
   is failed whenever an error response received.

### DIFF
--- a/sn_client/src/connections/listeners.rs
+++ b/sn_client/src/connections/listeners.rs
@@ -256,7 +256,16 @@ impl Session {
 
         let _handle = tokio::spawn(async move {
             match msg {
-                ServiceMsg::QueryResponse { response, .. } => {
+                ServiceMsg::QueryResponse {
+                    response,
+                    correlation_id,
+                } => {
+                    trace!(
+                        "ServiceMsg with id {:?} is QueryResponse regarding {:?} with response {:?}",
+                        msg_id,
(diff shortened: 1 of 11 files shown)


davidrusu:
bors +

sn_client/src/connections/listeners.rs
```diff
@@ -265,8 +274,14 @@ impl Session {
                     if let Ok(op_id) = response.operation_id() {
                         if let Some(entry) = queries.get(&op_id) {
                             let all_senders = entry.value();
-                            for (_msg_id, sender) in all_senders {
-                                let res = sender.try_send(response.clone());
+                            // Only valid response shall get broadcast to all
+                            for (ori_msg_id, sender) in all_senders {
+                                let res = if response.is_success() || ori_msg_id == &correlation_id
+                                {
```
b-zee, line 280: Small suggestion, feel free to ignore.

davidrusu:
Looks like bor s crashed, lets cancel and try again

bors cancel

PR #1325, opened 2022-07-13
Clear out more async

davidrusu:
Changes:
- remove unused async
- remove RefCell in NetworkKnowledge
- remove Locks in SignatureAggregator and SectionKeyProvider

### DIFF
--- a/sn_dysfunction/src/lib.rs
+++ b/sn_dysfunction/src/lib.rs
@@ -288,9 +288,7 @@ mod tests {
         let mut dysfunctional_detection = DysfunctionDetection::new(adults.clone());
         let nodes_to_retain = adults[5..10].iter().cloned().collect::<BTreeSet<XorName>>();
 
-        dysfunctional_detection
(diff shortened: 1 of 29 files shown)


joshuef (approved):

joshuef:
I've made https://github.com/maidsafe/safe_network/pull/1328 to try out a larger stack size just now

b-zee:
An `Arc`ed value allocates on the heap. Good to note there is likely some stack increase after taking out `Arc`s.

davidrusu:
@b-zee interesting :thinking: 

Some potential culprits:

LRUCache uses ArrayVec under the hood which looks like it is being stack-allocated:
https://docs.rs/arrayvec/latest/src/arrayvec/arrayvec.rs.html#42-46

We make use of it here: https://github.com/maidsafe/safe_network/pull/1325/files#diff-ec5edf7281e30ebd769a3b98e9bc61ff8a560acd21edc480f7882174bf66dc08R31-R34

I'll box it and see how things look

PR #1331, opened 2022-07-14
chore: use timestamp for log files

joshuef:
This means that once written a log file's name will not change.
This hould make debugging a live network easier (your log file is less likely to change while viewing), and also pulling logs from a network via rsync fafster (as the namess/content aren't changing all the time).


joshuef:
cc @happybeing , as you worke don original impl, and this may impact vdash.

PR #1334, opened 2022-07-14
refactor(prefix_map): move SectionChain into NetworkPrefixMap

RolandSherwin:
- Introduce `NetworkPrefixMap::sections_dag` which is the SectionChain of all the all the sections we are aware of
    -  Added a proptest to make sure `sections_dag` and `sections` stay in sync.
- Remove `NetworkPrefixMap::genesis_key` and `NetworkKnowledge::genesis_key` as it can be obtained from `sections_dag`
- Remove prefix map test files from `./resources/test_prefix_maps` and instead write them during test.


sn_interface/src/network_knowledge/prefix_map/mod.rs
```diff
@@ -26,26 +26,32 @@ use bls::PublicKey as BlsPublicKey;
...
+#[derive(Debug, Clone)]
 pub struct NetworkPrefixMap {
-    /// Map of sections prefixes to their latest signed section authority providers.
-    sections: Arc<DashMap<Prefix, SectionAuth<SectionAuthorityProvider>>>,
     /// The network's genesis public key
     genesis_pk: BlsPublicKey,
+    /// Map of sections prefixes to their latest signed section authority providers.
+    sections: Arc<DashMap<Prefix, SectionAuth<SectionAuthorityProvider>>>,
+    /// A DAG containing all section chains of the whole network that we are aware of
+    pub section_dag: Arc<RwLock<SecuredLinkedList>>,
```
RolandSherwin, line 45: Used in client's `Session` which spawns tasks, hence the struct has to be `Sync`

davidrusu (commented):
**Edit: Just realized that `format!("{:?}", prefix_map.genesis_key())` predates your changes, we can hold off on this suggestion in this PR:**

using `format!("{:?}", prefix_map.genesis_key())` as the filename is a hack and will change each time someone changes the `Display` impl for BLS keys, lets do something more robust.

i.e. try:

`NetworkPrefixMap::genesis_key_hex() -> String`

That returns the genesis key as a hex-encoded string, you can use that as a filename reliably and won't depend on the BLS display impl.

sn_cli/src/operations/config.rs
```diff
@@ -586,22 +586,40 @@ pub mod test_utils {
...
+            prefix_maps.push(prefix_map);
+        }
+        Ok(prefix_maps)
+    }
 
     impl Config {
         pub async fn create_config(
             tmp_dir: &TempDir,
-            serialized_config: Option<&str>,
+            serialized_config: Option<String>,
```
davidrusu, line 617: change this to `config: Option<Settings>` and do the serializing to JSON only inside this function.

sn_cli/src/operations/config.rs
```diff
@@ -778,32 +793,30 @@ mod constructor {
...
+        match network_info {
+            NetworkInfo::Local(_, genesis_key) => {
+                assert_eq!(*genesis_key, Some(*prefix_map.genesis_key()))
+            }
+            _ => {
+                return Err(eyre!(
+                    "The network information should be of type NetworkInfo::Local"
+                ))
+            }
+        };
```
davidrusu, line 819:
```suggestion
        assert!(matches!(
            network_info,
            NetworkInfo::Local(_, Some(genesis_key)) if genesis_key == prefix_map.genesis_key())
        )
```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -499,6 +489,11 @@ impl NetworkKnowledge {
         &self.prefix_map
     }
 
+    // Returns mutable reference to network prefix map
+    pub fn mut_prefix_map(&mut self) -> &mut NetworkPrefixMap {
```
davidrusu, line 493: Convention is for `mut` on the end: `<name>_mut()`, i.e. `iter_mut()`

```suggestion
    pub fn prefix_map_mut(&mut self) -> &mut NetworkPrefixMap {
```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -137,8 +137,8 @@ pub fn section_has_room_for_node(
 /// Container for storing information about the network, including our own section.
 #[derive(Clone, Debug)]
 pub struct NetworkKnowledge {
-    /// Network genesis key
-    genesis_key: BlsPublicKey,
+    // TODO: get current section chain from NetworkPrefixMap::sections_dag once RwLock has been
+    // removed; else have to make a lot of fns as async
```
davidrusu, line 141: Seems like you've resolved this TODO already?

RolandSherwin: I'm writing a follow up PR to make changes to `NetworkKnowledge` struct, so including the changes there.

sn_interface/src/network_knowledge/prefix_map/mod.rs
```diff
@@ -119,74 +128,38 @@ impl NetworkPrefixMap {
...
             .collect()
     }
 
     /// Get `SectionAuthorityProvider` of a known section with the given prefix.
     #[allow(unused)]
     pub fn get(&self, prefix: &Prefix) -> Option<SectionAuthorityProvider> {
-        self.sections
-            .get(prefix)
-            .map(|entry| entry.value().value.clone())
+        self.sections.get(prefix).map(|sap| sap.value.clone())
```
davidrusu, line 148: Use your helpers when available
```suggestion
        self.get_signed(prefix).map(|sap| sap.value)
```

sn_interface/src/network_knowledge/prefix_map/mod.rs
```diff
@@ -96,15 +105,15 @@ impl NetworkPrefixMap {
...
             .iter()
-            .filter(|e| {
-                if let Some(prefix) = exclude {
-                    e.key() != prefix
+            .filter(|&(prefix, _)| {
+                if let Some(p) = exclude {
+                    prefix != p
                 } else {
                     true
                 }
```
davidrusu, line 113:
```suggestion
                Some(prefix) != exclude
```

sn_interface/src/network_knowledge/utils/mod.rs
```diff
@@ -14,35 +14,21 @@ use tokio::io::{AsyncReadExt, AsyncWriteExt};
...
-pub async fn compare_and_write_prefix_map_to_disk(prefix_map: &NetworkPrefixMap) -> Result<()> {
+pub async fn write_prefix_map_to_disk(prefix_map: &NetworkPrefixMap) -> Result<()> {
     // Open or create `$User/.safe/prefix_maps` dir
     let prefix_map_dir = get_prefix_map_dir()?;
     fs::create_dir_all(prefix_map_dir.clone())
         .await
         .map_err(|_| {
             Error::DirectoryHandling("Could not read '.safe/prefix_maps' directory".to_string())
         })?;
+    let tmp_prefix_map_file = prefix_map_dir.join(format!("temp-{:?}", prefix_map.genesis_key()));
```
davidrusu, line 25: You should be using the `tempfile` crate here (we already have it as a dependency). 

You still have over-write corruption issues with this algorithm since the temp file is deterministically named. i.e. all nodes running locally are using the same temp file.

davidrusu (changes requested):
Some good work here, but there's a few changes that are needed still

sn_interface/src/network_knowledge/utils/mod.rs
RolandSherwin: Oh yeah, that's right! Updating it!

sn_cli/src/operations/config.rs
RolandSherwin: This is really helpful! Found a couple more places where I replaced the match statements. Everything is much cleaner now!

sn_interface/src/network_knowledge/prefix_map/mod.rs
RolandSherwin: Aye! Will keep that in mind :+1:

sn_cli/src/operations/config.rs
```diff
@@ -494,10 +495,33 @@ impl Config {
...
+            let temp_file_name: String = thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(30)
+                .map(char::from)
+                .collect();
+            let mut path = path.clone();
+            path.pop();
+            path = path.join(temp_file_name);
+            path
+        };
```
davidrusu, line 510: I assume this will be replaced with `tempfile` once #1433 is in?

RolandSherwin: Aye! I will move this change to #1433

sn_interface/src/network_knowledge/utils/mod.rs
```diff
@@ -23,29 +28,18 @@ pub async fn compare_and_write_prefix_map_to_disk(prefix_map: &NetworkPrefixMap)
...
-    let mut file = fs::File::create(&prefix_map_file)
+    // If /tmp is on a different mount point, rename is not atomic; hence create tempdir within
+    // prefix_map_dir
+    let temp_dir = tempfile::tempdir_in(prefix_map_dir)
+        .map_err(|e| Error::FileHandling(format!("Error creating tempdir: {:?}", e)))?;
+    let tmp_prefix_map_file = temp_dir.path().join("temp_prefix_map");
+
+    let mut file = fs::File::create(&tmp_prefix_map_file)
         .await
         .map_err(|e| Error::FileHandling(e.to_string()))?;
```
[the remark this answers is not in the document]
RolandSherwin: Rename is not atomic if `/tmp` is on a different mount point. So wanted to mitigate that.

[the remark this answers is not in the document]
davidrusu: @RolandSherwin sorry yeah realized /tmp is not what we want, I updated the comment after I wrote it, please check again

davidrusu (approved):
:ship: good stuff

PR #1336, opened 2022-07-14
fix: propose offline against nodes having the same conn info

maqi:

### DIFF
--- a/sn_node/src/node/messaging/agreement.rs
+++ b/sn_node/src/node/messaging/agreement.rs
@@ -9,6 +9,7 @@
 use crate::node::{
     node_api::cmds::Cmd, relocation::ChurnId, Event, MembershipEvent, Node, Proposal, Result,
 };
+use sn_dysfunction::IssueType;
 use sn_interface::{
     messaging::system::{JoinResponse, KeyedSig, MembershipState, SectionAuth, SystemMsg},
     network_knowledge::{
@@ -124,12 +125,14 @@ impl Node {
                 existing_age + 1 < incoming_age
             };
 
+            // When the existing node is valid, just ignore the incoming agreement.
+            // Otherwise, track the existing node as really bad in the dysfunction.
             if is_existing_valid {
                 return Ok(cmds);
             } else {
-                let mut list = BTreeSet::new();
-                let _ = list.insert(existing_peer.name());
-                cmds.push(Cmd::ProposeOffline(list));
+                let _ = self
+                    .dysfunction_tracking
+                    .track_issue(existing_peer.name(), IssueType::ReallyBad);
             }
         }
 
--- a/sn_node/src/node/messaging/agreement.rs
+++ b/sn_node/src/node/messaging/agreement.rs
@@ -103,6 +103,36 @@ impl Node {
             }
         }
 
+        // In case there is a node already having the same connection info,
+        // we need to:
+        //   1, if the existing node is valid, reject the incoming agreement.
+        //   2, if the existing node is invalid, accept the incoming agreement
+        //      AND propose an offline against the existing node.
+        //
+        // The existing node is valid when:
+        //   1, Its age is lower than the incoming node by more than 1.
+        //   2, Its age is higher than the incoming node by just 1.
+        if let Some(existing_peer) = self
+            .network_knowledge
+            .get_peer_by_conn_info(&new_info.addr())
+        {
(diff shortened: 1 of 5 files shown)


sn_node/src/node/messaging/agreement.rs
```diff
@@ -107,6 +108,36 @@ impl Node {
...
 
+        // In case there is a node already having the same connection info,
+        // we need to:
+        //   1, if the existing node is valid, reject the incoming agreement.
+        //   2, if the existing node is invalid, accept the incoming agreement
+        //      AND propose an offline against the existing node.
+        //
+        // The existing node is valid when:
+        //   1, Its age is lower than the incoming node by more than 1.
+        //   2, Its age is higher than the incoming node by just 1.
```
davidrusu, line 119: This validation should be done in the membership validations to prevent even accepting these nodes into the network in the first place.

Consider moving this logic to `NodeState::validate` here:
https://github.com/maidsafe/safe_network/blob/main/sn_interface/src/network_knowledge/node_state.rs#L71

`NodeState::validate` is called by Membership before it is proposed.

You'll need to modify the `Validate` function to take something like `members: BTreeMap<XorName, NodeState>` instead of the current `members: BTreeSet<XorName>`.

sn_node/src/node/messaging/agreement.rs
```diff
@@ -107,6 +108,36 @@ impl Node {
...
+        if let Some(existing_node) = self
+            .network_knowledge
+            .get_name_by_conn_info(&new_info.addr())
+        {
+            let existing_age = existing_node[XOR_NAME_LEN - 1];
+            let incoming_age = new_info.age();
```
davidrusu, line 125: Consider modifying the `get_name_by_conn_info()` helper to return a `Peer` or a `NodeState` so that we can use the `age()` methods.

```suggestion
        if let Some(existing_peer) = self
            .network_knowledge
            .get_peer_by_conn_info(&new_info.addr())
        {
            let existing_age = existing_peer.age();
            let incoming_age = new_info.age();
```

sn_node/src/node/messaging/agreement.rs
```diff
@@ -107,6 +108,36 @@ impl Node {
...
+            } else {
+                existing_age + 1 < incoming_age
+            };
+
+            if is_existing_valid {
+                return Ok(cmds);
+            } else {
+                let mut list = BTreeSet::new();
+                let _ = list.insert(existing_node);
+                cmds.push(Cmd::ProposeOffline(list));
```
joshuef, line 137: I think we should try and keep all offline proposals to be kicked off via dysfunction. That way we have _one_ place that's the decision center for node punishment.

We should here instead (if we cannot avoid this altogether as @davidrusu suggests): write something to sn_dysfunction (a new `IssueType`), which will basically _always_ result in the node being returned as `dysfunctional`.

sn_node/src/node/messaging/agreement.rs
maqi: I checked the code of `NodeState::validate` and the correspondent place calling it.
It seems the validation will be carried out agaisnt all received proposal shares?
If so, this will be quite in-efficient and even need to pass in additional parameter of `connection_info` in ? 

Also, if want to make dyfunction to be responsible of casting the offline vote, move the validate into `node_state` will become tricky.

So, I'd prefer leave the work of move it to `NodeState::validate` as a separte task/PR afterwards, if really decided to do so.

maqi: I checked the code of `NodeState::validate` and the correspondent place calling it.
It seems the validation will be carried out agaisnt all received proposal shares?
If so, this will be quite in-efficient and even need to pass in additional parameter of `connection_info` in ? 

Also, if want to make dyfunction to be responsible of casting the offline vote, move the validate into `node_state` will become tricky.

So, I'd prefer leave the work of move it to `NodeState::validate` as a separte task/PR afterwards, if really decided to do so.

maqi: > would the overall vote to join not fail though (for the duplicate node) if enough nodes have the check

There is still chance the invalid one passed the vote first before the valid one.
With enough nodes have the check, the consequence will be `valid one` get rejected by mistake.

joshuef (changes requested):

joshuef:
closing, as superseded by healthcheck #1375

PR #1339, opened 2022-07-15
chore: refactor WireMsg.serialize to use BytesMut and avoid some Vec …

joshuef:
…allocations


b-zee (approved):
Looks okay! Cargo.lock perhaps in a different PR on the off chance it breaks something?

davidrusu (approved):
Good change :+1:

PR #1340, opened 2022-07-15
chore(client): remove extra bootstrap step

RolandSherwin:
Also remove `Session::genesis_key` as it's unused. Can be obtained from `NetworkPrefixMap` if required.


davidrusu (approved):
Nice

PR #1349, opened 2022-07-18
chore: use SendMsg for send_node_msg_to_nodes

joshuef:
As it always results in a delivery group of 1. Which is pointless

### DIFF
--- a/sn_node/src/node/api.rs
+++ b/sn_node/src/node/api.rs
@@ -130,7 +130,7 @@ impl Node {
     }
 
     // Send message to peers on the network.
-    pub(crate) fn send_msg_to_nodes(&self, mut wire_msg: WireMsg) -> Result<Option<Cmd>> {
+    pub(crate) fn send_msg_on_to_nodes(&self, mut wire_msg: WireMsg) -> Result<Option<Cmd>> {
         let dst_location = wire_msg.dst_location();
         let (targets, dg_size) = delivery_group::delivery_targets(
             dst_location,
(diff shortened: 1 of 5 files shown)


sn_node/src/node/data/records/mod.rs
```diff
@@ -215,40 +214,42 @@ impl Node {
...
-        let close_full_adults = if let Some(closest_empty) = candidates.iter().next() {
+        let candidates_clone = candidates.clone();
+        let close_full_adults = if let Some(closest_empty) = candidates_clone.iter().next() {
             full_adults
                 .iter()
                 .filter_map(|name| {
-                    if target.cmp_distance(name, closest_empty) == Ordering::Less {
-                        Some(*name)
+                    if target.cmp_distance(name, &closest_empty.name()) == Ordering::Less {
+                        Some(closest_empty)
```
Yoga07, line 223: Was this wrong before? :thinking:

joshuef: ah no, i've borked that. Good spot!

joshuef: sorted

Yoga07 (approved):

sn_node/src/node/api.rs
```diff
@@ -130,7 +130,7 @@ impl Node {
     }
 
     // Send message to peers on the network.
-    pub(crate) fn send_msg_to_nodes(&self, mut wire_msg: WireMsg) -> Result<Option<Cmd>> {
+    pub(crate) fn send_msg_on_to_nodes(&self, mut wire_msg: WireMsg) -> Result<Option<Cmd>> {
```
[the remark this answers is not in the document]
joshuef: yeh, that was my thinking, as it talks about `relay` below. Happy for other ideas (or i can revert the name change... but i also wanted to distinguish this from other funcs of similar name)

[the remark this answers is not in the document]
oetyng: Ah, yeah not too important, was just wondering. Might have some naming idea eventually. These send msg fns with similar names are moving towards less confusion bit by bit :)

PR #1352, opened 2022-07-18
Refactor sn_node binary and add OpenTelemetry protocol server

b-zee:
Refactored the binary into multiple files for readability. Tweaked how logging is setup to use more ergonomic mechanisms that are endorsed by `tracing` docs.

The OTLP setup requires it to be run ideally from a Tokio Runtime, which is why the last commit has a refactor that makes the node use a single runtime. This might cause some side effects. Also, compiling the OpenTelemetry related crate dependencies seem to require a build dependency, the ProtoBuf compiler `protoc`, which is also less than ideal.

Start a Jaeger instance (a service that keeps track of the logs and has a web interface):
```
docker run --name jaeger \
  -e COLLECTOR_OTLP_ENABLED=true \
  -p 16686:16686 \
  -p 4317:4317 \
  -p 4318:4318 \
  jaegertracing/all-in-one:1.36
```
To give it a try: `RUST_LOG=sn_node=trace cargo run --release --bin sn_node --features otlp -- --first --skip-auto-port-forwarding --local-addr=127.0.0.1:0`.

Then, at http://localhost:16686, via tags one can filter several things, e.g.the instance id: `service.instance.id=<PID>`, where `PID` is the process ID of the node.

It seems to be a limitation that stuff only gets logged once the 'span' ends. So on spans that take long (async), it doesn't yield any result before it ends.


sn_node/README.md
```diff
@@ -1,9 +1,30 @@
...
+
+By specifying the `otlp` feature for the `sn_node` binary, logs will be sent to an OTLP endpoint. This endpoint can be configured by environment variables. (See [opentelemetry.io/docs/...](https://opentelemetry.io/docs/reference/specification/protocol/exporter/) for more information.)
+
+```sh
+export OTEL_EXPORTER_OTLP_ENDPOINT="http://localhost:4317"
+export RUST_LOG=sn_node=trace
+cargo run --release --bin sn_node --features otlp -- --first --skip-auto-port-forwarding --local-addr=127.0.0.1:0
+```
+
+Before running the node, an OTLP endpoint should be available. An example of an OTLP-supporting endpoint is Jaeger, which can be launched with Docker like this:
```
joshuef, line 17: Can we get a link to jaeger for easy setup 🙇

PR #1358, opened 2022-07-19
chore: remove SendMsgDeliveryGroup Cmd

joshuef:
This is not needed or really utilised in any real way.

We can just use SendMsg cmds

### DIFF
--- a/sn_node/src/node/api.rs
+++ b/sn_node/src/node/api.rs
@@ -132,7 +132,7 @@ impl Node {
     // Send message to peers on the network.
     pub(crate) fn send_msg_on_to_nodes(&self, mut wire_msg: WireMsg) -> Result<Option<Cmd>> {
         let dst_location = wire_msg.dst_location();
-        let (targets, dg_size) = delivery_group::delivery_targets(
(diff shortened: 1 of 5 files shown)


sn_node/src/node/api.rs
```diff
@@ -164,9 +164,8 @@ impl Node {
         let dst_pk = self.section_key_by_name(&target_name);
         wire_msg.set_dst_section_pk(dst_pk);
 
-        let cmd = Cmd::SendMsgDeliveryGroup {
+        let cmd = Cmd::SendMsg {
             recipients: targets.into_iter().collect(),
-            delivery_group_size: dg_size,
```
oetyng, line 169: Seems logging on line 160 will be confusing with the dg_size mentioned.

joshuef: ah yeh, i wonder if i can just strip that whole dg stuff out 🤔

joshuef: hmm, maybe not in this pass. i'll clean that log up 👍

oetyng (approved):

PR #1359, opened 2022-07-19
Simplify node dirs

oetyng:
There was a `NodeApi` and a `api.rs` for `Node`, and I noticed that the `NodeApi` was only used in `routing_minimal` and `routing_stress` examples, so moved it to a `TestNodeApi` used only there. Then we could disambiguate this api thing.

Logic for starting a node was moved to its own file.


sn_node/src/node/mod.rs
```diff
@@ -46,12 +50,15 @@ pub use self::{
...
     data::DataStorage,
     error::{Error, Result},
-    node_api::{
+    flow_ctrl::{
         event::{Event, MembershipEvent, MessagingEvent, NodeElderChange},
         event_channel::EventReceiver,
-        NodeApi,
     },
 };
+//#[cfg(feature = "test-utils")]
```
joshuef, line 58: lets remove these comments if unwanted 👍

joshuef (changes requested):
Excellent. That's been bugging me for a bit, I kept starting some similar refactors and kept getting sidetracked. 

Good stuff 💪 

Only thing is some commented code to be removed, and I see some empty doc blocks

sn_node/src/node/mod.rs
oetyng: Ah yeah that got left. I think they are wanted but then the examples need the feature flagging, and I think it should be its own PR.
I'll remove that.

PR #1361, opened 2022-07-19
High prio sends

joshuef:
atop #1360

### DIFF
--- a/sn_node/src/node/flow_ctrl/cmds.rs
+++ b/sn_node/src/node/flow_ctrl/cmds.rs
@@ -178,6 +178,11 @@ impl Cmd {
     pub(crate) fn priority(&self) -> i32 {
         use Cmd::*;
         match self {
+            // TODO: check if we can pull out node DST here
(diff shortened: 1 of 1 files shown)


oetyng (commented):
Just for docs I'll write down here what we talked about:

This high prio of send msg makes perfect sense.
All the job leading up to the msg is done, and we want to just get rid of the msg from the system asap. The actual sending isn't (I assume) very resource intensive either.
The _communication_ of our _results of spent work_, should not be delayed - only the acceptance of _new_ work to spend time on (delayed as in sorted by prio).

So, even if we have pending high importance stuff like `dkg`, `membership` changes etc, any msgs we're continuously producing should leave the system asap, i.e. have highest priority. (And there, any `node-node` comms > `node>client` comms 👍 )

One possible additional improvement would be that within `SendMsg` cmds, the highest prio underlying msg gets sent first. But, for laters..

oetyng (approved):

PR #1367, opened 2022-07-20
feat(api): expose a public API which allows users to check if a DBC has been already spent on the network

bochaco:
This is a follow up PR (10th PR) to https://github.com/maidsafe/safe_network/pull/1317, PR https://github.com/maidsafe/safe_network/pull/1351, PR https://github.com/maidsafe/safe_network/pull/1226, PR https://github.com/maidsafe/safe_network/pull/1301, PR https://github.com/maidsafe/safe_network/pull/1274, PR https://github.com/maidsafe/safe_network/pull/1235, PR https://github.com/maidsafe/safe_network/pull/1097, PR https://github.com/maidsafe/safe_network/pull/1105, and PR https://github.com/maidsafe/safe_network/pull/1143.

- Expose a public `is_dbc_spent` API which allows users to check if a DBC's `KeyImage` has been already spent on the network.
- Have the CLI `wallet deposit` command to perform a verification is the supplied DBC has been already spent before depositing into a wallet.
- Allow users to provide a `--force` flag with the CLI `wallet deposit` command to skip the verification of DBC already spent and force the deposit into the wallet.

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -17,8 +17,9 @@ use bytes::Bytes;
 use log::{debug, warn};
 use sn_client::Client;
 use sn_dbc::{
-    rng, AmountSecrets, Hash, Owner, OwnerOnce, PublicKey, RingCtTransaction, Signature,
-    SpentProof, TransactionBuilder,
+    rng, AmountSecrets, Error as DbcError, Hash, IndexedSignatureShare, KeyImage, Owner, OwnerOnce,
+    PublicKey, RingCtTransaction, Signature, SpentProof, SpentProofContent, SpentProofShare,
+    TransactionBuilder,
 };
 use sn_interface::types::Token;
 use std::collections::{BTreeMap, BTreeSet};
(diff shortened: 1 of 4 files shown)


sn_api/src/app/wallet.rs
```diff
@@ -1125,4 +1189,19 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_wallet_is_dbc_spent() -> Result<()> {
+        let (safe, unspent_dbc, _) = new_safe_instance_with_dbc().await?;
+
+        // the api shall confirm the genesis DBC's key_image has been spent
+        let is_genesis_spent = safe.is_dbc_spent(GENESIS_DBC.key_image_bearer()?).await?;
+        assert!(is_genesis_spent);
+
+        // and confirm the other DBC has not been spent yet
```
jacderida, line 1201: Minor point, but I think these should be two separate test cases.

sn_cli/src/subcommands/wallet.rs
```diff
@@ -154,6 +159,20 @@ pub async fn wallet_commander(
                 )?)
             };
 
+            if force {
+                println!(
+                    "\nWARNING: --force flag set, hence skipping verification to check if \
+                supplied DBC has been already spent.\n"
+                );
+            } else if safe.is_dbc_spent(dbc.key_image_bearer()?).await? {
```
jacderida, line 167: Sorry if this is incorrect, but would you not need to accommodate owned DBCs here too?

bochaco: ah yes, good point, will try to fix that

jacderida (approved):

PR #1368, opened 2022-07-21
feat(traceroute): trace a message's flow in the network

Yoga07:
Implements a feature to trace a message's flow through the network. `WireMsgHeader` piggybacks a list to which entities(Clients, Elders, Adults) attach their identities(PublicKey) wherever they process the message.

This is currently implemented for Client's queries, commands, and commandACKs(all of which fall under `ServiceMsg`). This will be extended to support inter-node messages(`SystemMsg`) in the near future.

Build and run with `--features=traceroute` to log traceroutes.

### DIFF
--- a/sn_interface/src/messaging/serialisation/wire_msg.rs
+++ b/sn_interface/src/messaging/serialisation/wire_msg.rs
@@ -19,6 +19,11 @@ use custom_debug::Debug;
 use serde::Serialize;
 use xor_name::XorName;
 
+#[cfg(feature = "traceroute")]
+use crate::types::PublicKey;
+#[cfg(feature = "traceroute")]
+use serde::Deserialize;
+
 /// In order to send a message over the wire, it needs to be serialized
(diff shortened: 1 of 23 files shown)


sn_interface/src/messaging/serialisation/wire_msg.rs
```diff
@@ -38,6 +43,15 @@ pub struct WireMsg {
     pub payload_debug: Option<std::sync::Arc<dyn std::fmt::Debug + Send + Sync>>,
 }
 
+#[cfg(feature = "traceroute")]
+/// PublicKey of the entity that created/handled it's associated WireMsg
```
b-zee, line 47:
```suggestion
/// PublicKey of the entity that created/handled its associated WireMsg
```

b-zee (approved):

joshuef (changes requested):
Let enable the feature to be passed into `testnet` bin so we can easily use it locally. Otherwise LGTM 👍

joshuef (approved):

PR #1371, opened 2022-07-21
Add health check

joshuef:
Close https://github.com/maidsafe/safe_network/pull/1336

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -15,16 +15,18 @@ pub(super) mod event_channel;
 pub(crate) mod tests;
 
 pub(crate) use self::cmd_ctrl::CmdCtrl;
-
 use crate::comm::MsgEvent;
 use crate::node::{flow_ctrl::cmds::Cmd, messages::WireMsgUtils, Error, Node, Result};
-
+use ed25519_dalek::Signer;
 use sn_interface::{
     messaging::{
+        data::{DataQuery, DataQueryVariant, ServiceMsg},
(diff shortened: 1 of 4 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -110,6 +116,57 @@ impl FlowCtrl {
...
+                    public_key: PublicKey::Ed25519(keypair.public),
+                    signature: Signature::Ed25519(signature),
+                };
+
+                let proofed_auth = AuthorityProof::verify(auth, payload)?;
+                let msg_id = MsgId::new();
+                let our_info = node.info();
+                let origin = our_info.peer();
+
+                // generate the cmds, and ensure we go through dysfunction tracking
```
maqi, line 155: we could just push the send cmd here and log it in the dysfunction.
the issue of calling `handle_valid_service_msg` is that:
1, it will then query the 4 Adults in one go? if so, may need to adjust the interval and update the comment?
2, if there is other `un-expected or un-wanted` check or operation happens inside, this `manually composed` query may trigger some other un-defined behaviour?

joshuef: 1. going through `handle_valid_service_msg`gets us dysfunction detection on both ends of the request as normal. I think it makes sense to go through that "standard" flow for this.

2. Not that I'm aware of. It's just setting up a fake client msg at this stage.

davidrusu (approved):
Looks solid :+1:

davidrusu:
Looks like the traceroute change is causing your changes to fail to build

joshuef:
closing in favour of #1375

PR #1372, opened 2022-07-21
CI: Add check for checking docs

b-zee:
add a PR check for linting docs

E.g. checks broken links in the documentation


davidrusu:
image
Looks like the check is being skipped?

b-zee:
@davidrusu Well spotted... But, in this case that is correct. There is no source code changed in this PR. I actually had a commit in this PR that checked the failure, but removed the commit again after I verified the CI did the right thing. I might have better done this differently, but CI is not my specialty. :grin:

The second commit in this PR actually adds a few paths that will decide whether these tools are run. But, those paths have not been changed in this case.

davidrusu:
Makes sense! :ship:

PR #1374, opened 2022-07-21
chore: simplify data replication msg sending

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -340,35 +340,33 @@ impl FlowCtrl {
                         let system_msg =
                             SystemMsg::NodeCmd(NodeCmd::ReplicateData(vec![data_to_send]));
 
-                        for recipient in recipients {
-                            let name = recipient.name();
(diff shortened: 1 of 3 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -276,35 +276,33 @@ impl FlowCtrl {
...
-
-                            debug!(
-                                "{:?} Data {:?} to: {:?} w/ {:?} ",
-                                LogMarker::SendingMissingReplicatedData,
-                                address,
-                                recipient,
-                                wire_msg.msg_id()
-                            );
+                        let name = recipients[0].name();
+                        let dst = sn_interface::messaging::DstLocation::Node {
```
davidrusu, line 280: Is Node the correct destination here? Would `DstLocation::Section` fit better?

joshuef: No, it's only going to the specific nodes that need this data.

oetyng:
fat fingered the rerun of jobs.. was supposed to rerun failed

PR #1381, opened 2022-07-24
refactor: expose known keys on network knowledge

oetyng:
The method can be called directly instead of passing known keys in the
cmds.

### DIFF
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -614,6 +614,14 @@ impl NetworkKnowledge {
         self.chain.has_key(key)
     }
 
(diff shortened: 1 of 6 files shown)


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -614,6 +614,14 @@ impl NetworkKnowledge {
         self.chain.has_key(key)
     }
 
+    /// Return weather current section chain has the provided key
```
joshuef, line 617:
```suggestion
    /// Return a vec of all known keys
```

But also this doc doesn't seem t

joshuef (approved):

PR #1384, opened 2022-07-24
fix(joins): check reachability before sending msgs

oetyng:
We tried to send a msg to the peer that failed the check. That wouldn't work though.
Also, other branches in same flow were sending msgs before doing that check.

### DIFF
--- a/sn_node/src/node/messaging/system_msgs.rs
+++ b/sn_node/src/node/messaging/system_msgs.rs
@@ -390,7 +390,14 @@ impl Node {
             SystemMsg::HandoverAE(gen) => self.handle_handover_anti_entropy(sender, gen),
             SystemMsg::JoinRequest(join_request) => {
                 trace!("Handling msg: JoinRequest from {}", sender);
-                self.handle_join_request(sender, *join_request, comm).await
+
+                // First do reachability check (we can't respond to the node if it fails).
+                if comm.is_reachable(&sender.addr()).await.is_err() {
+                    debug!("Ignoring JoinRequest from {sender} - not reachable.");
+                    return Ok(vec![]);
+                };
+
+                self.handle_join_request(sender, *join_request).await
             }
             SystemMsg::JoinAsRelocatedRequest(join_request) => {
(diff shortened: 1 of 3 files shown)


joshuef:
> We tried to send a msg to the peer that failed the check. That wouldn't work though.

I think we'd be able to send it back on the initial connection. (The comm check attempts to open a _new_ conn I think, to ascertain reachability)

sn_node/src/node/messaging/system_msgs.rs
```diff
@@ -400,7 +407,13 @@ impl Node {
                     return Ok(vec![]);
                 }
 
-                self.handle_join_as_relocated_request(sender, *join_request, comm)
+                // First do reachability check (we can't respond to the node if it fails).
+                if comm.is_reachable(&sender.addr()).await.is_err() {
+                    debug!("Ignoring JoinAsRelocatedRequest from {sender} - not reachable.");
```
Yoga07, line 412: Would nodes on relocation stall if they don't get a response back? Think we should add a timeout there alongside this change maybe?

It won't be the same case for normal joins as they have a timeout and exit the process if there's no response, but the flow is a bit different here for relocations.

Yoga07: Even without this change, I think we should have a timeout during relocations(or at least a warning of the time it takes) as they are the same as regular joins.

oetyng: > I think we'd be able to send it back on the initial connection. (The comm check attempts to open a _new_ conn I think, to ascertain reachability)

Mm, that's right, the incoming conn is (most likely) still there, and there msgs _can_ be received. Well, then there's no need to move the check.

oetyng:
> I think we'd be able to send it back on the initial connection. (The comm check attempts to open a new conn I think, to ascertain reachability)

Mm, that's right, the incoming conn is (most likely) still there, and there msgs can be received. Well, then there's no need to move the check.

oetyng:
The incoming conn is (most likely) still there, and there msgs can be received.

PR #1393, opened 2022-07-25
chore: make Sign of SystemMsgs higher prio as this could hold them ba…

joshuef:
…ck before sending


davidrusu (approved):
:+1:

joshuef:
Tests I've run look dodgy. will dig in the morn

joshuef:
Testnet fails were due to failed init of the network. All good

PR #1396, opened 2022-07-26
chore(node): simplify the send funcitonality

joshuef:
Previously we've had a send to client and a send ot nodes, the latter
which would only send to a subset (with the delivery group setting), and
reported errors.

We've stripped out the delivery group stuff already as unnecessary. So here
we just have one function to send from nodes, any errors are handled
appropriately if the recipient is in our section, otherwise ignored.

### DIFF
--- a/sn_node/src/comm/mod.rs
+++ b/sn_node/src/comm/mod.rs
@@ -33,7 +33,6 @@ use sn_interface::{
 
 use bytes::Bytes;
 use dashmap::DashMap;
-use futures::stream::{FuturesUnordered, StreamExt};
 use qp2p::{Endpoint, IncomingConnections};
 use std::{collections::BTreeSet, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::{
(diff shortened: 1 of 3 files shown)


sn_node/src/comm/mod.rs
```diff
@@ -353,66 +251,69 @@ impl Comm {
...
                         recipient
                     );
-                    try_next(error, recipient, &mut tasks);
+                    error!(
+                        "Sending message (msg_id: {:?}) to {:?} (name {:?}) failed as we have disconnected from the peer. (Error is: {})",
+                        msg_id,
+                        addr,
+                        name,
+                        error,
+                    );
```
davidrusu, line 276: These log lines can be compressed down to ~1-2 lines by inlining the variables:
```suggestion
                    error!(
                        "Sending message (msg_id: {msg_id:?}) to {addr:?} (name {name:?}) failed as we have disconnected from the peer. (Error is: {error})"
                    );
```

davidrusu (approved):
Looks good

PR #1402, opened 2022-07-27
chore: refactor FlowCtrl/CmdCtrl

joshuef:
- removes Cmd retries as unused.
- Removes clone from `FlowCtrl`, 
- unifies periodic Cmd checks into ont loop

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -52,7 +52,8 @@ const DATA_BATCH_INTERVAL: Duration = Duration::from_millis(50);
 const DYSFUNCTION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
 // 30 adult nodes checked per minute., so each node should be queried 10x in 10 mins
 // Which should hopefully trigger dysfunction if we're not getting responses back
-const HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(2);
+const ADULT_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(2);
+const ELDER_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(5);
 // to prevent cpu racing
 const LOOP_SLEEP_INTERVAL: Duration = Duration::from_millis(10);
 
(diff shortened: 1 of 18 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -111,327 +216,218 @@ impl FlowCtrl {
...
-                            wire_msg,
-                            recipients,
-                        };
+                let mut recipients = vec![];
 
-                        if let Err(e) = self.cmd_ctrl.push(cmd).await {
-                            error!("Error in data replication loop: {:?}", e);
-                        }
-                    }
+                for peer in data_recipients.iter() {
```
oetyng, line 368: this loop, if it's to get owned peers within the collection, then a copied/cloned thingie instead should work?

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -48,29 +48,137 @@ const DYSFUNCTION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
...
+
+            if last_dysfunction_check.elapsed() > DYSFUNCTION_CHECK_INTERVAL {
+                last_dysfunction_check = now;
+                let dysf_cmds = Self::check_for_dysfunction(self.node.clone()).await;
+                cmds.extend(dysf_cmds);
+            }
+
+            // Finally, handle any incoming conn messages
+            // this requires mut self
+            if let Some(msg) = self.incoming_msg_events.recv().await {
```
oetyng, line 170: It seems adults are not listening to incoming msgs when this block is down here?
It would need to go at line 100 to be executed by adults as well.

Another thing though, is that the periodic checks are then driven by us receiving msgs.
Probably that is okay. But could cause some strange gaps.

An option would be to use `try_recv` api on the receiver, and do a tiny sleep (to not have the cpu race) if returning `Err(TryRecvError::Empty)`. So, the sleep length would be around the interval we'd expect to check for new msgs (I don't know what value we want there).

oetyng (changes requested):

sn_node/src/node/flow_ctrl/mod.rs
joshuef: yeh good catch!

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -48,29 +48,137 @@ const DYSFUNCTION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
...
+                debug!("msggggeventtt");
```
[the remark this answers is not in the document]
joshuef: nop!

joshuef:
Closing as merged main changes in via #1465

PR #1406, opened 2022-07-27
Lock tweaks2

joshuef:

### DIFF
--- a/sn_interface/src/types/cache/mod.rs
+++ b/sn_interface/src/types/cache/mod.rs
@@ -144,8 +144,7 @@ where
             let excess = len - self.capacity;
             let excess_keys: Vec<_>;
             {
-                let read_items = &self.items;
-                let mut items = read_items.iter().collect_vec();
+                let mut items = self.items.iter().collect_vec();
 
                 // reversed sort
                 items.sort_by(|(_, item_a), (_, item_b)| item_b.elapsed().cmp(&item_a.elapsed()));
--- a/sn_interface/src/types/cache/mod.rs
+++ b/sn_interface/src/types/cache/mod.rs
@@ -13,7 +13,6 @@ use itertools::Itertools;
 use std::collections::BTreeMap;
 use std::hash::Hash;
 use std::time::Duration;
-use tokio::sync::RwLock;
 
 /// A [`BTreeMap`]-backed cache supporting capacity- and duration-based expiry.
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


dirvine (approved):

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -104,19 +109,33 @@ impl Dispatcher {
                 auth,
                 #[cfg(feature = "traceroute")]
                 traceroute,
-            } => {
-                let mut node = self.node.write().await;
+            } => match msg {
```
oetyng, line 112: It's `self.dysfunction_tracking.track_issue(` in `fn read_data_from_adults` that requires the write lock.
But now there is a new `TrackNodeIssueInDysfunction` for it, so no need to write lock on a read, and no need for the handle_valid_query_msg` fn, just issue that cmd and it will be dispatched to the write locked node.

Replace line 131 in node/data/records/mod.rs with

cmds.push(Cmd::TrackNodeIssueInDysfunction {
                                name: target.name(),
                                issue: IssueType::PendingRequestOperation(Some(operation_id)),
                            });

oetyng (changes requested):

sn_interface/src/types/cache/mod.rs
```diff
@@ -138,18 +133,18 @@ where
...
-        let len = self.len().await;
+    fn drop_excess(&mut self) {
+        let len = self.len();
         if len > self.capacity {
             let excess = len - self.capacity;
             let excess_keys: Vec<_>;
             {
-                let read_items = self.items.read().await;
+                let read_items = &self.items;
                 let mut items = read_items.iter().collect_vec();
```
davidrusu, line 148: No need for a let binding for the reference anymore.
```suggestion
                let mut items = self.items.iter().collect_vec();
```

sn_node/examples/routing_stress.rs
```diff
@@ -752,55 +752,51 @@ struct ProbeTracker {
...
         }
     }
 
     async fn receive(&mut self, dst: &XorName, received_public_key: &PublicKey) {
-        let result = &self
+        let result = self
             .sections
-            .iter()
-            .find_map(|(prefix, cache)| block(cache.get(dst)).map(|pk| (prefix, pk, cache)));
+            .iter_mut()
```
davidrusu, line 769: Does this need to be `iter_mut()`

PR #1412, opened 2022-07-28
Extend traceroute for system msgs and improve reability

Yoga07:
Example of the new logging format: 
`"Traceroute: Elder(Ed25519(PublicKey(b70f..ff5a))) => Adult(Ed25519(PublicKey(1797..4750)))"`

### DIFF
--- a/sn_interface/src/messaging/serialisation/wire_msg.rs
+++ b/sn_interface/src/messaging/serialisation/wire_msg.rs
@@ -21,6 +21,9 @@ use crate::types::PublicKey;
 #[cfg(feature = "traceroute")]
 use serde::Deserialize;
 
+#[cfg(feature = "traceroute")]
+use std::fmt::{Debug as StdDebug, Display, Formatter};
+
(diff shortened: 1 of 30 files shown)


sn_interface/src/messaging/serialisation/wire_msg.rs
```diff
@@ -278,11 +315,15 @@ impl WireMsg {
...
+    pub fn add_trace(&mut self, traceroute: &mut Traceroute) {
+        self.header
+            .msg_envelope
+            .traceroute
+            .0
+            .append(&mut traceroute.0)
     }
 
-    pub fn show_trace(&self) -> Vec<Entity> {
+    pub fn show_trace(&self) -> Traceroute {
```
davidrusu, line 326: `show_trace()` implies this function would print or somehow display a traceroute, but it's just accessing the trace
```suggestion
    pub fn trace(&self) -> Traceroute {
```

davidrusu (approved):
Just one small nit

joshuef (approved):

PR #1386, opened 2022-07-24
WIP - simplify send msg

oetyng:
WIP

This places signing and wire msg instantiation in one location, and
removes lots of old variables that aren't used in the flow anymore.

PR #1413, opened 2022-07-28
Do not require node write lock on query

oetyng:
This creates the `AddToPendingQieries` cmd, which adds asyncly to the
list.
Also cleans up the `read_data_from_adults` fn a bit.

**NB**: Draft while on top of pending PR #1386 (simplify send msg).


oetyng:
Superseded by #1417

PR #1417, opened 2022-07-29
Simplify messaging + cleanup the code (-1.5k lines)

oetyng:
- Places signing and wire msg instantiation in one location
- More reuse of methods to replace duplication of code.
- Simplifies comms module.
- Removes lots of old code and variables that aren't used in the flow anymore:
-> Many instances of `WireMsg` plumbing removed.
-> `SrcLocation` and `DstLocation` are removed (replaced with `Dst`).
-> `delivery_group` mod is removed.


oetyng:
Superseded by #1422, #1423 and #1424.

PR #1419, opened 2022-07-30
chore(ci): increase node startup interval to 5s

oetyng:
This not only removes the ci join issues, but also the fairly common other test errors (not enough spendproof shares, failed to obtain response, not enough chunks etc - seems this could be due to ghost nodes).

With this we do not need to override our ci all the time due to these tests failing.

Also it removes the risk of hiding new bugs introduced, behind what we assume were those join related errors.

***

Interval could probably be lowered some. But that'd optimally be done while working on fixing the underlying problem.

Goal for that would be to not have any join-rate that is capable of throwing off the network like that.


jacderida (approved):

willief:
Bit late, but from my own attempts at scripting a stable baby-fleming with different verbosity on certain nodes I found even a 5s delay was not always enough. Consider making this 7 or even 8s if there are still issues.

PR #1422, opened 2022-08-01
refactor(send): simplify send msg

oetyng:
Removes 512 lines.

This places signing and wire msg instantiation in one location, and
removes lots of old variables that aren't used in the flow anymore.


sn_node/src/node/bootstrap/relocate.rs
```diff
@@ -44,7 +44,6 @@ impl JoiningAsRelocated {
     // shall be fed back with `handle_join_response` function.
     pub(crate) fn start(
         node: NodeInfo,
-        genesis_key: BlsPublicKey,
```
joshuef, line 47: Do we not need the genesis key here to know we're joining the correct network?

sn_node/src/node/data/records/mod.rs
```diff
@@ -79,91 +70,61 @@ impl Node {
...
+        });
+
+        let msg = SystemMsg::NodeQuery(NodeQuery::Data {
+            query: query.variant,
+            auth: auth.into_inner(),
+            origin: EndUser(origin.name()),
+            correlation_id: msg_id,
+        });
+
+        cmds.push(self.trace_system_to_one(
```
joshuef: I find this naming confusing. What's the trace part here? vs just sending a system msg and maybe traceroutes are on or off? It's not clear from the docs on the func to me either `Send a (`SystemMsg`) message to a node`

sn_node/src/node/flow_ctrl/cmds.rs
```diff
@@ -216,6 +217,8 @@ impl Cmd {
 
             Comm(_) => 7,
```
joshuef: I think `AddToPendingQueries` always needs to be higher than `HandleValidSystemMsg` for node query respones, right? Can we note that here (it looks like it _is_ currently, doing this)

(An aside... I was wondering how we can explicitly link all the priority values across Cmds and the messages... I didn't have a good way. Not sure if you've any ideas, or if perhaps we should just look to declare _all_ of the prios in one place to avoid mistakes adjusting things here? )
```suggestion
// AddToPendingQueries needs to _always_ be handled before incoming query responses, otherwise we may have a response with no client recipients eg...
```

sn_node/src/node/data/records/mod.rs
oetyng: This is a method that takes a feat flagged traceroute arg. But we don't need that everywhere (mostly, we do not actually), so there is a method that doesn't take it.

This method is changed in the other PR.

sn_node/src/node/flow_ctrl/cmds.rs
oetyng: Move Q to PR #1426

oetyng:
Superseded by PR #1423.

PR #1423, opened 2022-08-01
refactor(send): reuse methods

oetyng:
[refactor(send): simplify send msg](https://github.com/maidsafe/safe_network/pull/1423/commits/a619fc7aea975c6aa3201dfd257132195f062ee1)
This places signing and wire msg instantiation in one location, and
removes lots of old variables that aren't used in the flow anymore.
(Removes 512 lines together with [bc6c012](https://github.com/maidsafe/safe_network/pull/1423/commits/bc6c012417ae13e3692000c25364551730fee010).)

[refactor(messaging): remove more unused code](https://github.com/maidsafe/safe_network/pull/1423/commits/716f731c088db5e7fb3e2e008943c0d29f4d8141)
More reuse of methods to replace duplication of code.
Deprecates delivery group, since it is no longer used.
Also, `DstLocation` and `SrcLocation` are removed (replaced with `Dst`).
(Removes 985 lines.)


sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -306,6 +288,84 @@ impl Dispatcher {
     }
 }
 
+fn get_msgs(
```
joshuef, line 291: Wee doc comment here would be good. 

Perhaps renaming to `generate_wire_msgs`?

PR #1430, opened 2022-08-01
feat(cli): wallet CLI cmd to allow storing output DBC to a file upon reissuing it

bochaco:
Allows to pass a file path to write the output DBC when reissuing with `wallet reissue` CLI cmd. If there is a problem writing to the file the CLI falls back to print the DBC data out. This arg can be extended to also support a wallet xorurl, so the user can reissue and get the output DBC either written to a file, or automatically deposited to another wallet the user owns.

This PR also renames the `--public-key` argument of `wallet reissue` command to `--to` as agreed in the review herein.

Example outputs:
```
$ safe wallet reissue 3.5 --save ./reissued_dbc.txt --from safe://hyryynywwgpr4yofnhpsm8kkcskjmtu6bq8m8834j88qia9djdzxfhy7mdhb6o
DBC content written at './reissued_dbc.txt'.
Reissued DBC with 3.5 safecoins.
This is a bearer DBC that can be spent by anyone.
```

```
$ safe wallet reissue 1.25 --save /invalid/file/path --from safe://hyryynywwgpr4yofnhpsm8kkcskjmtu6bq8m8834j88qia9djdzxfhy7mdhb6o
Error: Unable to write DBC at '/invalid/file/path': No such file or directory (os error 2).
Reissued DBC with 1.25 safecoins.
-------- DBC DATA --------
4c34acb4552a7a...e1de4400000000
--------------------------
This is a bearer DBC that can be spent by anyone.
```

### DIFF
--- a/sn_cli/src/subcommands/wallet.rs
+++ b/sn_cli/src/subcommands/wallet.rs
@@ -17,6 +17,7 @@ use color_eyre::{eyre::eyre, eyre::Error, Help, Result};
 use sn_api::{Error as ApiError, Safe};
 use sn_dbc::{Dbc, Error as DbcError};
 use std::path::Path;
+use tokio::fs;
 
 #[derive(Subcommand, Debug)]
 pub enum WalletSubCommands {
@@ -54,7 +55,7 @@ pub enum WalletSubCommands {
         force: bool,
     },
     #[clap(name = "reissue")]
-    /// Reissue a DBC from a wallet to a SafeKey.
+    /// Reissue a DBC from a wallet.
     Reissue {
         /// The amount to reissue
         amount: String,
@@ -64,12 +65,15 @@ pub enum WalletSubCommands {
         /// To reissue the DBC to a particular owner, provide their public key. This should be a
         /// hex-encoded BLS key. Otherwise the DBC will be reissued as bearer, meaning anyone can
         /// spend it. This argument and the --owned argument are mutually exclusive.
-        #[clap(long = "public-key")]
-        public_key_hex: Option<String>,
+        #[clap(long = "to")]
+        to: Option<String>,
         /// Set this flag to reissue as an owned DBC, using the public key configured for use with
         /// safe. This argument and the --public-key argument are mutually exclusive.
         #[clap(long = "owned")]
         owned: bool,
+        /// A file path to store the content of the reissued DBC.
+        #[clap(long = "save")]
+        save: Option<String>,
     },
 }
 
@@ -200,18 +204,19 @@ pub async fn wallet_commander(
         WalletSubCommands::Reissue {
             amount,
             from,
-            public_key_hex,
+            save,
+            to,
             owned,
         } => {
-            if owned && public_key_hex.is_some() {
+            if owned && to.is_some() {
                 return Err(eyre!(
-                    "The --owned and --public-key arguments are mutually exclusive."
+                    "The --owned and --to arguments are mutually exclusive."
                 )
                 .suggestion(
                     "Please run the command again and use one or the other, but not both, of these \
                     arguments."));
             }
-            let pk = if let Some(pk_hex) = public_key_hex {
+            let pk = if let Some(pk_hex) = to {
                 Some(PublicKey::from_hex(&pk_hex)?)
             } else if owned {
                 let sk = read_key_from_configured_credentials(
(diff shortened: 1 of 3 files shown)


sn_cli/tests/cli_wallet.rs
```diff
@@ -925,3 +925,149 @@ async fn wallet_deposit_spent_owned_dbc_force() -> Result<()> {
...
+    .stdout(predicate::str::contains(format!(
+        "DBC content written at '{}'.",
+        to_dbc_file_path.display()
+    )))
+    .stdout(predicate::str::contains("-------- DBC DATA --------").not())
+    .stdout(predicate::str::contains(
+        "Reissued DBC with 0.987654321 safecoins.",
+    ))
+    .success();
+
```
joshuef, line 971: I think we should perhaps assert the files contents here too?

sn_cli/tests/cli_wallet.rs
```diff
@@ -925,3 +925,149 @@ async fn wallet_deposit_spent_owned_dbc_force() -> Result<()> {
...
+            "0.111",
+            "--from",
+            &wallet_xorurl,
+            "--to",
+            &to_dbc_file_path.display().to_string(),
+        ],
+        Some(0),
+    )?;
+
+    safe_cmd(
```
joshuef, line 1051: ah wait, is that what this cmd is doing (checking the issued file)? if so ignore my other comment!

sn_cli/src/subcommands/wallet.rs
```diff
@@ -229,17 +234,43 @@ pub async fn wallet_commander(
             let dbc = safe.wallet_reissue(&from, &amount, pk).await?;
             let dbc_hex = dbc.to_hex()?;
 
+            // Write the DBC to a file is the user required it, but we'll fall
+            // back to print it out if that fails.
```
davidrusu, line 238:
```suggestion
            // Write the DBC to a file if the user requests it, otherwise print to stdout.
```

sn_cli/tests/cli_wallet.rs
```diff
@@ -925,3 +925,149 @@ async fn wallet_deposit_spent_owned_dbc_force() -> Result<()> {
...
+    let to_dbc_file_path = tmp_data_dir.child(get_random_string());
+
+    safe_cmd(
+        [
+            "wallet",
+            "reissue",
+            "0.987654321",
+            "--from",
+            &wallet_xorurl,
+            "--to",
```
davidrusu, line 956: `--to` seems like the wrong word here. I would think `--to` would set the owner of the output DBC.

How about `--save` or `--save-to`

bochaco: makes sense to me, perhaps I can then rename `--public-key` to `--to` then, cc: @jacderida

sn_cli/tests/cli_wallet.rs
bochaco: yeah, i was gonna have a single testcase for both checks but I know @jacderida would be suggesting to split them up, so I thought we'd go for two separate testcases

sn_cli/src/subcommands/wallet.rs
bochaco: I don't think it's the same as I was trying to also highlight that it falls back to print it out if there was a problem trying to write the file, i'll use some of the wording you used though

sn_cli/tests/cli_wallet.rs
jacderida: Ah, in general, I like to have separate test cases when different code paths are being exercised. I think it makes the test case easier to understand and also pinpoints the failure more accurately when the test fails. In our code I often seen large test cases where a whole bunch of stuff is being lumped together.

In this particular case though, I don't quite understand what the distinction is between `wallet_deposit_dbc_reissued_to_file` and `wallet_reissue_to_file_succeeds`?

jacderida: In my opinion, if you're testing the functionality of two separate commands, that would merit having two different test cases.

sn_cli/src/subcommands/wallet.rs
jacderida: To be honest, I think the code describes this fine on its own without the comment.

sn_cli/tests/cli_wallet.rs
jacderida: Yeah, that does make sense. The thing is though, it makes me question whether `--from` would be the right choice of word for the source wallet, given that from/to conjures up thoughts of like an email sender and recipient. Probably a discussion for another day though.

jacderida (approved):

PR #1431, opened 2022-08-01
fix(file_store): use correct data address type

oetyng:
The type including `SafeKey` had been incorrectly used (since it is not
a network side concept), which caused a lot of `Result` return values
bloating the upstream call tree unnecessarily.


joshuef (approved):
bors r+

davidrusu (approved):
Nice

PR #1432, opened 2022-08-01
chore: cleanup unnecessary options and results

oetyng:
These were remaining in places where there was no path for None or
Error.
Cleaning these up removes accidental complexity.


joshuef (approved):
bors r+

sn_dysfunction/src/detection.rs
```diff
@@ -286,6 +286,12 @@ impl DysfunctionDetection {
     }
 }
 
+fn rand_op_id() -> OperationId {
+    use rand::Rng;
+    let mut rng = rand::thread_rng();
+    OperationId(rng.gen())
+}
+
```
davidrusu, line 294: This feels like it should be a constructor on `OperationId`. i.e. `OperationId::random(rand::thread_rng())`

davidrusu (approved):
Nice find

PR #1433, opened 2022-08-02
chore(cli): make writes atomic

RolandSherwin:
Fixes the write corruption when running CLI tests using multiple threads.

### DIFF
--- a/sn_cli/src/operations/config.rs
+++ b/sn_cli/src/operations/config.rs
@@ -19,10 +19,12 @@ use std::{
     collections::BTreeMap,
     default::Default,
     fmt,
+    io::Write,
     path::{Path, PathBuf},
     thread,
     time::Duration,
 };
+use tempfile::NamedTempFile;
 use tokio::fs;
 use tracing::debug;
 use url::Url;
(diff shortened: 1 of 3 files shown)


joshuef (approved):

sn_cli/src/operations/config.rs
```diff
@@ -551,15 +552,37 @@ impl Config {
...
+            let temp_file_name: String = thread_rng()
+                .sample_iter(&Alphanumeric)
+                .take(30)
+                .map(char::from)
+                .collect();
+            let mut path = self.cli_config_path.clone();
+            path.pop();
+            path = path.join(temp_file_name);
+            path
+        };
```
davidrusu, line 570: Any reason why we aren't using the `tempfile` crate here?

davidrusu (changes requested):
Should use the `tempfile` crate instead of generating random names by hand.

davidrusu:
bors cancel

sn_cli/src/operations/config.rs
RolandSherwin: Did not want to introduce a new dep, will make changes to use `tempfile` if `thread_rng()` has a possibility to generate the same value.

davidrusu: It's actually already a dependency of `sn_cli` you can see it with:

    cargo tree -p sn_cli -i tempfile

It's pulled in through multiple other dependencies.

RolandSherwin: Aha! That's quite useful! Making the changes right now..

sn_cli/src/operations/config.rs
```diff
@@ -494,12 +495,31 @@ impl Config {
...
+            .parent()
+            .ok_or_else(|| eyre!("Path {:?} should be inside a folder", path.display()))?;
+        let temp_file = NamedTempFile::new_in(prefix_map_dir)
+            .wrap_err_with(|| "Error creating temp file".to_string())?;
+
         let serialized = Self::serialise_prefix_map(prefix_map)?;
-        fs::write(path, serialized).await.wrap_err_with(|| {
-            format!("Unable to write NetworkPrefixMap to '{}'", path.display())
+        fs::write(temp_file.path(), serialized)
+            .await
```
davidrusu, line 507: We have the file opened when `temp_file` was instantiated, so we can just write to it directly with `write_all`. `fs::write` will re-open the file and then write.

```suggestion
        temp_file.write_all(serialized)
```

RolandSherwin: I have made the changes, but my thought was to make the write async, `tempfile` is sync only. So I assume the cost of opening the file > sync writes?

davidrusu: >So I assume the cost of opening the file > sync writes?

Possibly, but it's so infrequent that I don't think it will even register.

davidrusu (approved):
Thanks! good changes throughout

PR #1436, opened 2022-08-02
feat(api): expose a public API to reissue many output DBCs from a wallet

bochaco:


sn_api/src/app/wallet.rs
```diff
@@ -1221,4 +1249,65 @@ mod tests {
...
+        // let's partition the total amount to reissue in a few amounts
+        let output_amounts = vec![
+            dbc1_balance.as_nano() - 700,
+            dbc2_balance.as_nano() - 700,
+            150,
+            100,
+            60,
+            90,
+        ];
+        assert_eq!(reissued_amount, output_amounts.iter().sum::<u64>());
```
jacderida, line 1275: Why would you need to assert something here? Has any depositing related code changed in this PR? If not, wouldn't we expect `wallet_deposit` to be covered and working from those tests?

Also, just a more general question. I take it we only require one test case for this new code because most of it is going through the single reissue code that we already tested?

bochaco: > Why would you need to assert something here?
this is just to make sure if we change the amounts above they still are what they are meant to be, i.e. they all should sum up to the same value determined to be reissued, remember this is just testing code, and it's just to make sure that if anyone modifies the test, the values are valid for it.

>  take it we only require one test case for this new code because most of it is going through the single reissue code that we already tested?
yes, I'm not really adding new functionality but just refactoring our private API to reissue DBCs which was already supporting to reissue to many, so now we just use that internal/private api to expose the two public apis, both go through the same flow/logic for reissuing dbcs.

jacderida (approved):

PR #1439, opened 2022-08-03
feat(churnid): make the content explicit

oetyng:
We previously allowed any length vec, while we only based it on
signatures. Therefore, we could restrict it to the signature contents
boundary.


davidrusu (approved):
:+1: looks good

PR #1450, opened 2022-08-08
Log Sleep with unique id

maqi:


sn_node/src/node/flow_ctrl/cmd_ctrl.rs
```diff
@@ -29,7 +30,7 @@ use tokio::{sync::RwLock, time::Instant};
 
 type Priority = i32;
 
-const SLEEP_TIME: Duration = Duration::from_millis(10);
+const SLEEP_TIME: u64 = 10;
```
joshuef, line 33: lets note this is millis

joshuef (approved):

PR #1452, opened 2022-08-08
feat: add AeProbe dysfunction. Refactor score calculation

joshuef:


sn_dysfunction/src/detection.rs
```diff
@@ -73,100 +81,111 @@ impl DysfunctionDetection {
         let mut knowledge_scores = BTreeMap::new();
         let mut op_scores = BTreeMap::new();
         let mut dkg_scores = BTreeMap::new();
+        let mut probe_scores = BTreeMap::new();
 
-        let nodes = self.nodes.to_vec();
+        let nodes = &self.nodes;
         for node in nodes.iter() {
```
maqi, line 87: you can then use `for node in &self.nodes` directly here?

sn_dysfunction/src/detection.rs
```diff
@@ -73,100 +81,111 @@ impl DysfunctionDetection {
...
+    /// get the node's score, relaative to the average for all nodes being tracked
+    fn calculate_node_score_for_type(&self, node: &XorName, issue_type: &IssueType) -> f32 {
+        let node_issue_count = self.get_node_issue_count_for_type(node, issue_type);
+
+        // we can shortcircuit here
+        if node_issue_count == 0 {
+            return 0.0;
+        }
+
+        debug!("node {node} issue count: {:?}", node_issue_count);
```
maqi, line 131: for easier tracking, better log the type of the issue together here?

sn_dysfunction/src/detection.rs
```diff
@@ -962,7 +1006,7 @@ mod knowledge_tests {
 
         // Add a new nodes
         let new_node = random_xorname();
-        dysfunctional_detection.add_new_node(new_node);
+        // dysfunctional_detection.add_new_node(new_node);
```
maqi, line 1009: no longer to call the separate `add_new_node` ? 
if so, just remove the line?

sn_dysfunction/src/lib.rs
```diff
@@ -170,6 +177,24 @@ impl DysfunctionDetection {
             }
         }
     }
+    /// Removes a probe tracker from the node liveness records. Returns true if a record was removed
```
maqi, line 180: no boolean returned.
comment need to be changed.

sn_dysfunction/src/detection.rs
```diff
@@ -216,23 +229,16 @@ impl DysfunctionDetection {
...
-            let zscore = match (mean, std_dev) {
-                (Some(mean), Some(std_deviation)) => {
-                    let diff = score - mean;
-                    let threshold = if (mean / 2.0) > DIFF_THRESHOLD {
-                        mean / 2.0
-                    } else {
-                        DIFF_THRESHOLD
-                    };
+            let zscore = {
+                let threshold = std_dev.unwrap_or(1.0);
```
maqi, line 233: the std_dev is measuring the `diff to the average` 
so I think the `threshold` is actually shall be something like `average/mean + std_dev` or `average/mean + std_dev/2` ?

maqi (approved):

PR #1454, opened 2022-08-08
refactor(anti-entropy): reduce AE msgs to one msg with a kind field

davidrusu:
This reduces the `SystemMsg::AntiEntopy{Update|Retry|Redirect}` msgs to a single `SystemMsg::AntiEntropy {.. kind }` msg to allow for de-duplication of logic.

### DIFF
--- a/sn_node/src/node/messaging/anti_entropy.rs
+++ b/sn_node/src/node/messaging/anti_entropy.rs
@@ -241,8 +241,49 @@ impl Node {
     ) -> Result<Vec<Cmd>> {
         let snapshot = self.state_snapshot();
 
-        self.update_network_knowledge(section_auth.clone(), section_signed, proof_chain, sender)
-            .await?;
+        info!("Anti-Entropy: message received from peer: {sender}");
+
+        let signed_sap = SectionAuth {
+            value: section_auth.clone(),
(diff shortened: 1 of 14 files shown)


sn_node/src/node/messaging/anti_entropy.rs
```diff
@@ -179,211 +186,78 @@ impl Node {
...
-                section_auth,
-                section_signed,
-                section_chain,
-                bounced_msg,
-                sender,
-            )
-            .await?;
+        // Check if we need to resend any messsages and who should we send it to.
+        let (bounced_msg, response_peer) = match kind {
+            AntiEntropyKind::Update { .. } => return Ok(cmds), // Nope, bail early
```
grumbach, line 207: Do we need the field in `AntiEntropyKind::Update`?

grumbach (approved):
LGTM!

sn_node/src/node/messaging/anti_entropy.rs
joshuef: `bounced_msg` you mean @grumbach ? No, the Update is purely about network knowledge. It's sent proactively or in response to a probe

joshuef (approved):

PR #1458, opened 2022-08-09
Add ae probe as health check

joshuef:
atop #1452

### DIFF
--- a/sn_interface/src/messaging/system/mod.rs
+++ b/sn_interface/src/messaging/system/mod.rs
@@ -13,6 +13,7 @@ mod msg_authority;
 mod node_msgs;
 mod node_state;
 mod signed;
+use bls::PublicKey as BlsPublicKey;
 
 pub use agreement::{DkgFailureSig, DkgFailureSigSet, DkgSessionId, Proposal, SectionAuth};
 pub use join::{JoinRejectionReason, JoinRequest, JoinResponse, ResourceProof};
@@ -70,7 +71,9 @@ pub enum SystemMsg {
         kind: AntiEntropyKind,
     },
     /// Probes the network by sending a message to a random or chosen dst triggering an AE flow.
-    AntiEntropyProbe,
+    /// Sends the current section key of target section which we know
+    /// This expects a response, even if we're up to date.
+    AntiEntropyProbe(BlsPublicKey),
     #[cfg(feature = "back-pressure")]
     /// Sent when a msg-consuming node wants to update a msg-producing node on the number of msgs per s it wants to receive.
     /// It tells the node to adjust msg sending rate according to the provided value in this msg.
@@ -196,7 +199,7 @@ impl SystemMsg {
         use super::msg_type::BACKPRESSURE_MSG_PRIORITY;
         use super::msg_type::{
             ANTIENTROPY_MSG_PRIORITY, DKG_MSG_PRIORITY, JOIN_RELOCATE_MSG_PRIORITY,
-            JOIN_RESPONSE_PRIORITY, NODE_DATA_MSG_PRIORITY,
+            JOIN_RESPONSE_PRIORITY, MEMBERSHIP_PRIORITY, NODE_DATA_MSG_PRIORITY,
         };
         match self {
             // DKG messages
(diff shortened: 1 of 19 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -171,6 +173,45 @@ impl FlowCtrl {
...
+                        name: elder.name(),
+                        issue: sn_dysfunction::IssueType::AwaitingProbeResponse,
+                    });
+                }
+
+                // Send a probe message to an elder
+                cmds.push(node.generate_section_probe_msg());
+
+                for cmd in cmds {
+                    info!("Sending healthcheck elder probe {:?}", cmd);
```
maqi, line 206: the cmd pushed is `Cmd::TrackNodeIssueInDysfunction` , which only `log` the issue but not `send the probe msg` ?

joshuef: it's sent below `cmds.push(node.generate_section_probe_msg());`

maqi: well, I mean the below `self.cmd_ctrl.push(cmd).await` only insert the cmd into the cmd_queue, and returns with a `watcher` . i.e. it's just pushing the cmd into the queue, but not actually `carry out the sending action now` . Hence the comment could be little bit misleading.
Unless we wait on that `watcher.await` to complete, then we can call the `completion of that pushing is really sending`

sn_interface/src/messaging/system/mod.rs
```diff
@@ -70,7 +71,9 @@ pub enum SystemMsg {
         kind: AntiEntropyKind,
     },
     /// Probes the network by sending a message to a random or chosen dst triggering an AE flow.
-    AntiEntropyProbe,
+    /// Sends the current section key of target section which we know
+    /// This expects a response, even if we're up to date.
+    AntiEntropyProbe(BlsPublicKey),
```
maqi, line 76: I am little bit concered with this as the SectionKey could remain same for a long time during which many adults joined or left.
Shall we use `members` here or a `hash of members` to reduce msg size as a probing body to make the probing more updated reflecting the section status ?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -484,11 +484,17 @@ impl NetworkKnowledge {
...
 
-    // Try to merge this `NetworkKnowledge` members with `peers`.
+    /// Try to merge this `NetworkKnowledge` members with `peers`.
+    /// Checks if we're already up to date before attempting to verify and merge members
     pub fn merge_members(&self, peers: BTreeSet<SectionAuth<NodeState>>) -> Result<bool> {
         let mut there_was_an_update = false;
+        let our_current_members = self.section_peers.members();
 
         for node_state in peers.iter() {
+            if our_current_members.contains(node_state) {
```
maqi, line 494: just to double confirm this `contains` will not block an update from `Joined` to `Left` .
i.e. the check is against the combination of `name + node_status` , not by name only

sn_node/src/node/flow_ctrl/mod.rs
joshuef: Hmm, i'm not sure that really matters? It's all about averaging out over time. It seems the extra complexity to have that only go after send is not worth it? 

Otherwise we have this same pattern for chunk queries eg, so we'd need to change it across the board, so perhaps not for this PR?

sn_interface/src/messaging/system/mod.rs
joshuef: > I am little bit concered with this as the SectionKey could remain same for a long time during which many adults joined or left.
I'm not sure i see the concern? It's a PK only. `members` would be bigger, `hash of members` more or less the same size?

Here we're just piggybacking upon functionality we have. So `AeProbe` should actually just trigger a normal Ae response _if the sending PK info is out of date_.

We _then_ *if it's not out of date*, handle the msg at elders and respond. This response is not about the content (the sender's info was valid), it's about saying "i exist". 

We _could_ add more msgs for this... specifically, but I thought it was overkill... So maybe we should do that?

sn_node/src/node/flow_ctrl/mod.rs
maqi: ok.
yeah it won't affect any pratical execution.
just may cause some mis-impression during tracking/debugging the logs.

I am OK with just `ignore` it currently.

sn_interface/src/messaging/system/mod.rs
maqi: well, Why I am bit concerned with the `same content body of section key` is that there will be higher chance that response to probe A could be counted as response to probe B, which may cause some trouble.

If we use the `msg_id` as well to track the resonse, then shall not be a problem.
just maybe an `outdate` may not be detected in time. which shall be ok, as we always reponse with ALL section members I know, right?

maqi (approved):

PR #1464, opened 2022-08-10
refactor(client): sn_client to only read a default prefix map file, updates to be cached on disk by user

bochaco:
- CLI to cache the up to date PrefixMap after all commands were executed and right before exiting.
- Refactoring sn_cli::Config to remove some redundant code.

### DIFF
--- a/sn_client/src/api/mod.rs
+++ b/sn_client/src/api/mod.rs
@@ -200,7 +200,7 @@ impl Client {
     /// This is updated by the client as it receives Anti-Entropy/update messages from the network.
     /// Any user of this API is responsible for caching it so it can use it for any new `Client`
     /// instance not needing to obtain all this information from the network all over again.
-    pub async fn get_prefix_map(&self) -> NetworkPrefixMap {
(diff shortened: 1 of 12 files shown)


sn_client/src/api/mod.rs
```diff
@@ -192,6 +195,15 @@ impl Client {
             .has_key(section_key)
     }
 
+    /// NetworkPrefixMap used to bootstrap the client on the network.
+    ///
+    /// This is updated by the client as it receives Anti-Entropy/update messages from the network.
+    /// Any user of this API is responsible for caching it so it can use it for any new `Client`
+    /// instance not needing to obtain all this information from the network all over again.
+    pub async fn get_prefix_map(&self) -> NetworkPrefixMap {
```
b-zee, line 203: Rust convention (https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter):

```suggestion
    pub async fn prefix_map(&self) -> NetworkPrefixMap {
```

b-zee (approved):
Looks good to me, @bochaco!

PR #1465, opened 2022-08-10
Simplified cmd ctrl refactor

joshuef:
Removes a lot of node cmd looping, and the `Clone` requirement on `CmdCtrl`, which will hopefully let us remove `RwLock` inside there in a follow up PR.

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -13,6 +13,7 @@ pub(super) mod event;
 pub(super) mod event_channel;
 #[cfg(test)]
 pub(crate) mod tests;
+use crate::log_sleep;
 
 pub(crate) use self::cmd_ctrl::CmdCtrl;
 
@@ -39,7 +40,7 @@ use tokio::{
         mpsc::{self, error::TryRecvError},
         RwLock,
     },
-    time::{sleep, Instant},
+    time::Instant,
 };
 
 const PROBE_INTERVAL: Duration = Duration::from_secs(30);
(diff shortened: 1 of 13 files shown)


testnet/bin.rs
```diff
@@ -49,7 +49,7 @@ const SAFE_NODE_EXECUTABLE: &str = "sn_node.exe";
 
 const BASE_TRACING_DIRECTIVES: &str = "testnet=info,sn_launch_tool=debug";
 const NODES_DIR: &str = "local-test-network";
-const DEFAULT_INTERVAL: &str = "5000";
+const DEFAULT_INTERVAL: &str = "100";
```
b-zee, line 52: I think I have seen this a few times before... Isn't there some middle ground?

b-zee (approved):

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -51,30 +55,206 @@ const DYSFUNCTION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
...
+                };
+            }
+
+            if last_dysfunction_check.elapsed() > DYSFUNCTION_CHECK_INTERVAL {
+                last_dysfunction_check = now;
+                let dysf_cmds = Self::check_for_dysfunction(self.node.clone()).await;
+                cmds.extend(dysf_cmds);
+            }
+
+            let no_cmds = cmds.is_empty();
```
maqi, line 243: you can do
```rust
if cmds.is_empty() {
    sleep(LOOP_SLEEP_INTERVAL).await;
}
```
here directly? no need to have an extra `no_cmds` ?

same comment for the above place for the adults

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -112,331 +292,214 @@ impl FlowCtrl {
...
-    fn start_running_health_checks(self) {
+    /// Initiates and generates all the subsequent Cmds to perform a healthcheck
+    async fn perform_health_checks(node: Arc<RwLock<Node>>) -> Result<Vec<Cmd>> {
         info!("Starting to check the section's health");
-        let _handle: JoinHandle<Result<()>> = tokio::task::spawn_local(async move {
-            let mut interval = tokio::time::interval(ADULT_HEALTH_CHECK_INTERVAL);
-            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+        let node = node.read().await;
+        let our_prefix = node.network_knowledge.prefix();
+        let _our_info = node.info();
```
maqi, line 300: for debugging? may can now be removed?

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -112,331 +292,214 @@ impl FlowCtrl {
...
+        }
     }
 
-    fn start_section_probing(self) {
-        info!("Starting to probe section");
-        let _handle = tokio::task::spawn_local(async move {
-            let mut interval = tokio::time::interval(SECTION_PROBE_INTERVAL);
-            interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
+    /// Generates a probe msg, which goes to a random section in order to
+    /// passively maintain network knowledge over time
```
maqi, line 354: this comment sounds very similar to the one for the `probe_the_network` function?

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -112,331 +292,214 @@ impl FlowCtrl {
...
-                    let cmd = self.node.read().await.generate_section_probe_msg();
-                    info!("Sending section probe msg");
-                    if let Err(e) = self.cmd_ctrl.push(cmd).await {
-                        error!("Error sending section probe msg: {:?}", e);
-                    }
-                }
-            }
-        });
+        // Send a probe message to an elder
+        if !prefix.is_empty() {
```
maqi, line 364: like the probing to the network knowledge, when in first section, no need to carry out probing own section?

testnet/bin.rs
joshuef: Ah, that's me testing our this PR w/ CI. That's not intended to stay.

sn_node/src/node/flow_ctrl/mod.rs
joshuef: I don't think so. I haven't changed the internal logic (i think). It's just moved about.

PR #1466, opened 2022-08-10
chore: remove wiremsg.priority as uneeded

joshuef:
Remove the Wiremssg priority as it's not actually used in sending from the nodes and can be quite CPU intensive. (accounting for 35% of CPU time here)

The Cmds themselves are prioritised in the node, so they should all get sent out in that order anyway


davidrusu (approved):
Nice find

PR #1467, opened 2022-08-10
refactor(tests): reorganise flow control unit tests

jacderida:
The unit tests in the `flow_ctrl` module provide coverage for messaging handling in the node. To run
each test, a `Node` must constructed, and this involves a lot of tedious setup code.

A `network_utils` testing module is introduced to organise the code related to this setup and to
also provide a `TestNodeBuilder` to reduce duplication and hopefully make it easier to provide more
coverage for message handlers. The doc comments should hopefully make clear how the struct can be
used in various different testing contexts. I will also be looking to extend its functionality a bit
further when I come to unit testing the message handlers for the DBC spentbook commands.

There were a few tests whose setup was too complex to use the builder because they require too much
customisation and seem best left alone.

A `cmd_utils` module is also introduced to organise the code for processing commands. I again also
plan on extending this when considering the DBC tests.


joshuef (approved):

joshuef:
nice cleanup 💪

PR #1470, opened 2022-08-11
feat: SecuredLinkedList using Merkle registers

RolandSherwin:
- Use Merkle reigsters to implement SecuredLinkedList

### DIFF
--- a/sn_interface/src/network_knowledge/errors.rs
+++ b/sn_interface/src/network_knowledge/errors.rs
@@ -51,7 +51,7 @@ pub enum Error {
     #[error("No matching Section")]
     NoMatchingSection,
     #[error(
-        "A JoinResponse was reeived after we've already joined the network. It has been ignored."
+        "A JoinResponse was received after we've already joined the network. It has been ignored."
     )]
     ExistingMemberConflict,
     #[error("The node in question is not a member of the section")]
@@ -62,4 +62,10 @@ pub enum Error {
     Consensus(#[from] sn_consensus::Error),
     #[error("An archived node attempted to rejoin the section")]
     ArchivedNodeRejoined,
+    #[error("The provided signature cannot be verified")]
+    InvalidSignature,
+    #[error("Key not found in the SectionsDAG: {}", hex::encode(_0.to_bytes()))]
+    KeyNotFound(bls::PublicKey),
+    #[error("The 'to' or 'from' key is not present in the same branch")]
+    InvalidBranch,
 }
--- a/sn_interface/src/network_knowledge/sections_dag.rs
+++ b/sn_interface/src/network_knowledge/sections_dag.rs
@@ -1,0 +1,883 @@
+// Copyright 2022 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+#![allow(dead_code)]
+
+use super::errors::{Error, Result};
+use crdts::{
+    merkle_reg::{Hash, MerkleReg, Node, Sha3Hash},
+    CmRDT,
+};
+use itertools::Itertools;
+use serde::{Deserialize, Serialize};
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt::{self, Debug, Formatter},
+    iter, mem,
+};
+use tiny_keccak::{Hasher, Sha3};
+
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+pub struct SectionInfo {
+    key: bls::PublicKey,
+    sig: bls::Signature,
+}
+
+impl Debug for SectionInfo {
+    fn fmt(&self, formatter: &mut Formatter) -> std::fmt::Result {
+        let bytes: Vec<u8> = self
+            .key
+            .to_bytes()
+            .into_iter()
+            .chain(self.sig.to_bytes().into_iter())
+            .collect();
+        let hex = hex::encode(bytes);
+        let hex: String = hex.chars().into_iter().take(10).collect();
+        write!(formatter, "SectionInfo({})", hex)
+    }
+}
+
+impl Sha3Hash for SectionInfo {
+    fn hash(&self, hasher: &mut Sha3) {
+        hasher.update(&self.key.to_bytes());
+        hasher.update(&self.sig.to_bytes());
+    }
+}
+
+/// A Merkle DAG of BLS keys where every key is signed by its parent key, except the genesis one.
+#[derive(Clone, PartialEq, Eq, Deserialize, Serialize)]
+pub struct SectionsDAG {
+    genesis_key: bls::PublicKey,
+    dag: MerkleReg<SectionInfo>,
+    dag_root: BTreeSet<bls::PublicKey>,
+    hashes: BTreeMap<bls::PublicKey, Hash>,
+}
+
+impl Debug for SectionsDAG {
+    fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
+        write!(formatter, "{:?}", self.keys().format(", "))
+    }
+}
+
+impl SectionsDAG {
+    /// Creates a new DAG containing only the `genesis_key`
+    pub fn new(genesis_key: bls::PublicKey) -> Self {
+        Self {
+            genesis_key,
+            dag: MerkleReg::new(),
+            dag_root: BTreeSet::new(),
+            hashes: BTreeMap::new(),
+        }
+    }
+
+    /// Insert new key into the DAG. `parent_key` must exist in the DAG and must validate
+    /// `signature`, otherwise error is returned.
+    pub fn insert(
+        &mut self,
+        parent_key: &bls::PublicKey,
+        key: bls::PublicKey,
+        signature: bls::Signature,
+    ) -> Result<()> {
+        if !self.verify_sig(parent_key, &key, &signature) {
+            return Err(Error::InvalidSignature);
+        }
+
+        let parent = if *parent_key == self.genesis_key {
+            self.dag_root.insert(key);
+            BTreeSet::new()
+        } else {
+            let hash = self
+                .hashes
+                .get(parent_key)
+                .ok_or(Error::KeyNotFound(*parent_key))?;
+            BTreeSet::from([*hash])
+        };
+
+        let info = SectionInfo {
+            key,
+            sig: signature,
+        };
+        // dag's children arg = parent
+        let child = self.dag.write(info, parent);
+        self.hashes.insert(key, child.hash());
+        self.dag.apply(child);
+
+        Ok(())
+    }
+
+    /// Insert new key into the DAG. `parent_key` must exist in the DAG and must validate
+    /// `signature`, otherwise error is returned.
+    fn insert_node(&mut self, parent_key: &bls::PublicKey, node: Node<SectionInfo>) -> Result<()> {
+        self.insert(parent_key, node.value.key, node.value.sig)
+    }
+
+    /// Get a partial `SectionsDAG` from the `from` key to the `to` key
+    /// Returns `Error::KeyNotFound` if the `to` key is not present in the DAG;
+    /// Returns `Error::InvalidBranch` if the `from` key is not found or is not a direct ancestor of the `to` key
+    pub fn partial_dag(&self, from: &bls::PublicKey, to: &bls::PublicKey) -> Result<Self> {
+        // start from the "to" key (bottom of the tree) and traverse to the root
+        let mut crdt_ops: Vec<Node<SectionInfo>> = Vec::new();
+        // if "to" key is genesis, returns error
+        let mut node = self.get_node(to)?;
+        loop {
+            if node.value.key == *from {
+                // we have reached the end
+                break;
+            }
+            // we don't insert the `from` node
+            crdt_ops.push(node.clone());
+            match self.parent_node(node.hash()) {
+                Some(parent) => node = parent,
+                // reached root of the dag
+                None => {
+                    // after reaching dag_root, return the partial_dag if from == genesis
+                    if *from == self.genesis_key {
+                        break;
+                    }
+                    // from key was apparently not a direct ancestor
+                    return Err(Error::InvalidBranch);
+                }
+            }
+        }
+        // create the SectionsDAG and apply the CRDT ops
+        let mut dag = SectionsDAG::new(*from);
+        let mut parent = *from;
+        for node in crdt_ops.into_iter().rev() {
+            let key = node.value.key;
+            dag.insert_node(&parent, node)?;
+            parent = key;
+        }
+        Ok(dag)
+    }
+
+    /// Update our current `SectionsDAG` with the keys from another `SectionsDAG`
+    /// Returns `Error::InvalidSignature` if the provided DAG fails signature verification
+    /// Returns `Error::KeyNotFound` if the genesis_key of either of the DAGs is not present in the
+    /// other
+    pub fn merge(&mut self, mut other: Self) -> Result<()> {
+        if !other.self_verify() {
+            return Err(Error::InvalidSignature);
+        }
+        // find which DAG is the parent
+        if !self.has_key(other.genesis_key()) {
+            if other.has_key(self.genesis_key()) {
+                mem::swap(self, &mut other);
+            } else {
+                return Err(Error::KeyNotFound(self.genesis_key));
+            }
+        }
+        // insert all the nodes from `other`
+        let mut need_to_visit: Vec<(bls::PublicKey, Node<SectionInfo>)> = Vec::new();
+        for key in other.dag_root.iter() {
+            need_to_visit.push((*other.genesis_key(), other.get_node(key)?));
+        }
+        while let Some((parent_key, current_node)) = need_to_visit.pop() {
+            need_to_visit.extend(
+                other
+                    .child_nodes(current_node.hash())
+                    .into_iter()
+                    .map(|child_node| (current_node.value.key, child_node)),
+            );
+            self.insert_node(&parent_key, current_node)?;
+        }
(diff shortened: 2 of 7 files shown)


sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+
+impl Sha3Hash for SectionInfo {
+    fn hash(&self, hasher: &mut Sha3) {
+        let bytes: Vec<u8> = self
+            .key
+            .to_bytes()
+            .into_iter()
+            .chain(self.sig.to_bytes().into_iter())
+            .collect();
+        hasher.update(bytes.as_slice());
```
davidrusu, line 35: You don't want to build up a bytes buffer and pass it all at once to hasher, instead call hasher update with each chunk of data you want to include in the hash.

```suggestion
        hasher.update(&self.key.to_bytes());
        hasher.update(&self.sig.to_bytes());
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+        };
+
+        let info = SectionInfo {
+            key,
+            sig: signature,
+        };
+        // children argument (DAG terminology) is the actual parent
+        let child = self.dag.write(info, parent);
+        self.dag.apply(child.clone());
+        self.hashes.insert(key, child.hash());
```
davidrusu, line 98: Reverse these two lines to avoid the clone on `child`

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+
+        Ok(())
+    }
+
+    /// Returns the list of the leaf sections. A section is considered a leaf if it has not
+    /// gone through any churn or split. Empty list if we only hold the genesis.
+    pub fn leaf_sections(&self) -> Vec<SectionInfo> {
+        self.dag.read().values().cloned().collect()
+    }
```
davidrusu, line 107: The semantics of this function is a bit unintuitive, can we make it private and only use it inside the `sections_dag` module?

My issue with it is that the genesis section would would be included in the result, I don't think callers out side of this module would expect that.

I would even give it an uglier name like: `non_genesis_leaf_sections()` to make the semantics more clear.

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+        let leaf_sections = self.leaf_sections();
+        if leaf_sections.is_empty() {
+            vec![self.genesis_key]
+        } else {
+            leaf_sections.iter().map(|info| info.key).collect()
+        }
+    }
+
+    /// Returns the list of all the keys in the SectionDAG
+    pub fn keys(&self) -> Vec<bls::PublicKey> {
```
davidrusu, line 121: These functions return `Vec`'s, is this to match the existing prefix map implementation?

I'm imagining the more common operation would be set membership checks. i.e. do we recognize a key, so would it be more useful to return a BTreeSet / HashSet from these functions?

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+                    return Err(e);
+                }
+            }
+        };
+        // if parent_node is None if we're provided the dag_root; hence parent = genesis key
+        Ok(self
+            .parent_node(hash)
+            .map(|node| node.value.key)
+            .or(Some(self.genesis_key)))
+    }
```
davidrusu, line 272: You can do the same kind of move as my suggestion for `get_child_keys()`, check for genesis first, then you don't need to `match`

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+                let parent = match self.get_parent_key(&key) {
+                    Ok(parent) => {
+                        match parent {
+                            Some(parent) => parent,
+                            // cannot reach? check
+                            None => break false,
+                        }
+                    }
+                    Err(_) => break false,
+                };
```
davidrusu, line 227:
```suggestion
                let parent = match self.get_parent_key(&key) {
                    Ok(Some(parent)) => parent,
                    Err(_) => break false,
                };
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+        let mut parent_key = match self
+            .get_parent_key(&proof_chain[0].value.key)
+            .unwrap_or(Some(self.genesis_key))
+        {
+            Some(key) => key,
+            None => {
+                error!("SectionsDAG::join() Unreachable since we provide Some(genesis)");
+                return Err(Error::KeyNotFound(self.genesis_key));
+            }
+        };
```
davidrusu, line 200: How about we take the `parent_key` as an argument to `join`?

a proof chains can be: `struct ProofChain { known_key: PublicKey, new_nodes: Vec<Node<..>> }`

`fn join(&mut self, proof_chain: ProofChain)`

This way the first entry of  `new_nodes` is not special.

This is similar to what you've done with the genesis key being separate from the merkle reg dag

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+                    error!(
+                        "SectionsDAG::keys() Unreachable since key is obtained from self Can be error only if hashes go out of sync"
+                    );
+                    return Vec::new();
+                }
+            };
+            // if children is empty, the below is not executed
+            children.iter().for_each(|node| {
+                need_to_visit.push(*node);
+            });
```
davidrusu, line 141:
```suggestion
            need_to_visit.extend(children);
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+        while let Some(key) = need_to_visit.pop() {
+            keys.push(key);
+            let children = match self.get_child_keys(&key) {
+                Ok(children) => children,
+                // unreachable since keys are obtained from self
+                Err(_) => {
+                    error!(
+                        "SectionsDAG::keys() Unreachable since key is obtained from self Can be error only if hashes go out of sync"
+                    );
+                    return Vec::new();
```
davidrusu, line 135: This is a good candidate for a panic, by returning a vec, we're masking a bug.

Either we panic, or we need to return a Result

```suggestion
                    panic!("SectionsDAG::keys() Unreachable since key is obtained from self Can be error only if hashes go out of sync");
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+            None => {
+                // parent set is empty if we are genesis's child
+                if *parent_key == self.genesis_key {
+                    self.dag_root.insert(key);
+                    BTreeSet::new()
+                } else {
+                    return Err(Error::KeyNotFound(*parent_key));
+                }
+            }
+        };
```
davidrusu, line 89: Can be written simpler if you check for genesis upfront:

```suggestion
        let parent = if parent_key == &self.genesis_key {
            self.dag_root.insert(key);
            BTreeSet::new()
        } else {
            let hash = self.hashes.get(parent_key).ok_or(Error::KeyNotFound(*parent_key))?;
            BTreeSet::from([*hash])
        };
```

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Got it! Making the change!

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Yeah, that makes sense, it's too quirky for an user to deal with..

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Ah! yeah, `BTreeSet` will be the best fit for that.. Will make changes to use `BTreeSet` for other similar functions as well..

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,888 @@
...
+                    return Err(e);
+                }
+            }
+        };
+        Ok(self
+            .child_nodes(hash)
+            .iter()
+            .map(|node| node.value.key)
+            .collect())
+    }
```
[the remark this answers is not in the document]
RolandSherwin: But for most cases, the provided key won't be the `genesis_key`, so it's better to check for it only when required right?

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Thanks for the input, this is much cleaner! Implementing it!

sn_interface/src/network_knowledge/sections_dag.rs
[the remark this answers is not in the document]
davidrusu: Are you thinking for perf? I'd prioritize keeping the logic tight / easy to follow with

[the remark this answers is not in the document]
RolandSherwin: Noted! will keep that in mind..

sn_interface/src/network_knowledge/errors.rs
```diff
@@ -62,4 +62,10 @@ pub enum Error {
     Consensus(#[from] sn_consensus::Error),
     #[error("An archived node attempted to rejoin the section")]
     ArchivedNodeRejoined,
+    #[error("The provided signature cannot be verified")]
+    InvalidSignature,
+    #[error("Key not found in the SectionsDAG")]
+    KeyNotFound(bls::PublicKey),
+    #[error("The 'to' and 'from' key are not present in the same branch")]
```
davidrusu, line 69:
```suggestion
    #[error("The 'to' or 'from' key is not present in the same branch")]
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+        }
+        // create the SectionsDAG and apply the CRDT ops
+        let mut dag = SectionsDAG::new(*from);
+        let mut parent = *from;
+        for node in crdt_ops.into_iter().rev() {
+            let key = node.value.key;
+            dag.insert_node(&parent, node)?;
+            parent = key;
+        }
+        Ok(dag)
```
davidrusu, line 154: That's pretty cool :+1:

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+    }
+
+    /// Update our current `SectionsDAG` with the keys from another `SectionsDAG`
+    /// Returns `Error::InvalidSignature` if the provided DAG fails signature verification
+    /// Returns `Error::KeyNotFound` if the genesis_key of either of the DAGs is not present in the
+    /// other
+    pub fn join(&mut self, mut other: Self) -> Result<()> {
```
davidrusu, line 161: I think `merge` is more clear
```suggestion
    pub fn merge(&mut self, mut other: Self) -> Result<()> {
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+        if !self.keys().contains(other.genesis_key()) {
+            if other.keys().contains(self.genesis_key()) {
+                mem::swap(self, &mut other);
+            } else {
+                return Err(Error::KeyNotFound(self.genesis_key));
+            }
+        }
+        if other.dag_root.is_empty() {
+            return Ok(());
+        }
```
davidrusu, line 175: This branch doesn't look necessary, the code below should be able to handle an empty dag_root

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+        while let Some(node) = need_to_visit.pop() {
+            // cannot be None since we don't provide the genesis_key
+            let parent = match other.get_parent_key(&node.value.key)? {
+                Some(parent) => parent,
+                None => {
+                    panic!("SectionsDAG::join() Unreachable since key can never be `genesis_key`")
+                }
+            };
+            need_to_visit.extend(other.child_nodes(node.hash()).into_iter());
```
davidrusu, line 189: vec.extend() takes an `IntoIterator` so you don't need to call `into_iter()`

```suggestion
            need_to_visit.extend(other.child_nodes(node.hash()));
```

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+        let leaf_sections = self.non_genesis_leaf_sections();
+        if leaf_sections.is_empty() {
+            BTreeSet::from([self.genesis_key])
+        } else {
+            leaf_sections.iter().map(|info| info.key).collect()
+        }
+    }
+
+    /// Returns the list of all the keys in the `SectionsDAG`
+    pub fn keys(&self) -> BTreeSet<bls::PublicKey> {
```
davidrusu, line 213: If you modify the `merkle-reg` implementation in `rust-crdt` to add the method "all_nodes() -> impl Iterator<Item=&Node>`, you can reduce this algorithm down to just calling `all_nodes()` map that to keys and chain on the `genesis_key`.



I'll leave it up to you if you want to make that change.

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+    /// ```ascii-art
+    /// D->E->F
+    ///    |
+    ///    +->G
+    /// ```
+    ///
+    /// Then such DAG is impossible to forge because the adversary would have to have access to
+    /// the secret key corresponding to `D` in order to validly sign `E`. Thus such DAG can be
+    /// safely considered trusted.
+    pub fn check_trust<'a, I>(&self, trusted_keys: I) -> bool
```
davidrusu, line 315: This function is odd... we trust a DAG iff we trust it's genesis key. I don't understand the need for a "main branch".

I think we can leave this and all code related to the "main_branch" out of this implementation, it doesn't make much sense to me why we'd need it.

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+
+    /// Returns the keys in the main branch
+    /// Consider the following DAG
+    /// ```ascii-art
+    /// A->B->C->E
+    ///       |
+    ///       +->F
+    /// ```
+    /// Then [A,B,C] are the keys in the main branch;
+    fn main_branch_keys(&self) -> Vec<bls::PublicKey> {
```
davidrusu, line 376: this idea seems odd to me, lets not include it in this implementation

davidrusu (commented):
It's looking really good, just a few small nits and maybe one semi-big one (the `check_trust` comment)

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+                mem::swap(self, &mut other);
+            } else {
+                return Err(Error::KeyNotFound(self.genesis_key));
+            }
+        }
+        if other.dag_root.is_empty() {
+            return Ok(());
+        }
+        // insert all the nodes from `other`
+        let mut need_to_visit: Vec<Node<SectionInfo>> = Vec::new();
```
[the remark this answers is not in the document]
RolandSherwin: That's a pretty neat idea! Making the changes!

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Would love to do that! Will raise a PR..

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,1020 @@
...
+                }
+            };
+            need_to_visit.extend(children);
+        }
+        keys
+    }
+
+    /// Returns whether `key` is present in this `SectionsDAG`.
+    pub fn has_key(&self, key: &bls::PublicKey) -> bool {
+        self.keys().iter().any(|existing_key| existing_key == key)
```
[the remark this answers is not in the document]
RolandSherwin: This is great! noted!

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Yeah, you're right, if the genesis key of either of the chain is not present in the other while merging, `merge` returns an error. I just mirrored the API from SLL without giving it a second thought. Will remove it!

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: The PR is up, will make changes as soon as its approved. https://github.com/rust-crdt/rust-crdt/pull/131

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -0,0 +1,883 @@
...
+    /// Returns `Error::InvalidSignature` if the provided DAG fails signature verification
+    /// Returns `Error::KeyNotFound` if the genesis_key of either of the DAGs is not present in the
+    /// other
+    pub fn merge(&mut self, mut other: Self) -> Result<()> {
+        if !other.self_verify() {
+            return Err(Error::InvalidSignature);
+        }
+        // find which DAG is the parent
+        if !self.keys().contains(other.genesis_key()) {
+            if other.keys().contains(self.genesis_key()) {
```
davidrusu, line 168: Can you use `self.has_key()` and `other.has_key()` here instead of `keys().contains()`?

sn_interface/src/network_knowledge/sections_dag.rs
davidrusu: Ok, lets keep this question in mind as we do the integration work: 

> Is `check_trust()` needed?

I suspect it's not needed since, as you say, you are doing all these checks in the `merge()` already, but we'll see as you do the work to swap out the implementation.

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Oops, sorry, overlooked it.

davidrusu (approved):
:100: nice work!

PR #1471, opened 2022-08-11
chore: further tweak dysf, reduce score by std dev for better avg.

joshuef:
Also adjusts tests to this, which now feels a bit saner too


sn_dysfunction/src/detection.rs
```diff
@@ -230,15 +230,12 @@ impl DysfunctionDetection {
 
         for (name, score) in pre_z_scores {
             let zscore = {
-                // avg score + 1/2 std dev
-                let threshold = (std_dev.unwrap_or(1.0) / 2.0) + mean.unwrap_or(1.0);
-
-                if score <= threshold {
+                if score <= 0.0 {
```
maqi, line 233: this `score` is from the previous calculated `final_score`, which is a sum of weighted scores, and actually will never be less than 0 ?

sn_dysfunction/src/detection.rs
```diff
@@ -230,15 +230,12 @@ impl DysfunctionDetection {
...
                     // small or negative diff mean nodes are doing well
                     None
                 } else {
                     // make the score relative to std_dev of nodes...
-                    Some(score)
+                    Some(score - std_dev)
```
maqi, line 238: this could be less than 0

maqi (approved):

PR #1478, opened 2022-08-12
Client/node cleanup and refactoring

b-zee:
Various tweaks that are chore/refactor. Also removes the `closest_or_opposite` method, as that wasn't used like it was designed. It actually is always used to find anything closest. In one case it's used to find a section closer than its own section, in which case an exclusion is passed in.

- chore: remove unused Session member
- chore(client): leave out unnecessary Arc<RwLock>
- chore: remove unnecessary unwrap
- refactor: circumvent clone to use reference
- refactor: SAP reference instead of clone
- refactor: clean up unused functionality


joshuef (approved):

joshuef (changes requested):
node unit tests are having a bad time

joshuef (approved):

PR #1481, opened 2022-08-15
Use bytes mut in seriliase

joshuef:
An effort to reduce allocations during messaging. Not sure of the impact here, so this is draft for now. Will report back w/ some heap tracing.

This is atop #1465


joshuef:
Closing as nothing significant gained here.

PR #1484, opened 2022-08-15
chore(node): increase data query limit

joshuef:
Now we differentiate queries per adult/index, we may need more queries.

### DIFF
--- a/sn_interface/src/messaging/system/mod.rs
+++ b/sn_interface/src/messaging/system/mod.rs
@@ -56,7 +56,7 @@ pub enum AntiEntropyKind {
 }
 
 #[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
-#[allow(clippy::large_enum_variant)]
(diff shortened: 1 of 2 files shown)


sn_interface/src/messaging/system/mod.rs
```diff
@@ -56,7 +56,7 @@ pub enum AntiEntropyKind {
 }
 
 #[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
-#[allow(clippy::large_enum_variant)]
+#[allow(clippy::large_enum_variant, clippy::derive_partial_eq_without_eq)]
```
davidrusu, line 59: Any reason not to derive the `Eq` trait as well?

davidrusu (approved):

sn_interface/src/messaging/system/mod.rs
b-zee: It can't be derived because there's an `f64` in that type... @davidrusu

PR #1487, opened 2022-08-16
chore(client): associated functions to methods

b-zee:


sn_client/src/connections/messaging.rs
```diff
@@ -513,7 +513,7 @@ pub(super) fn send_msg_in_bg(
 }
 #[instrument(skip_all, level = "trace")]
 pub(super) async fn send_msg(
```
davidrusu, line 515: Any reason to keep this as a standalone function? Could we moved into an `impl Session` as well?

PR #1489, opened 2022-08-16
feat(logging): simplify log format to '<timestamp> [module] LEVEL <msg>'

davidrusu:
I've found our logging format to be very difficult to parse, IMO it's cluttered with a ton of unnecessary information making command line tools ineffective following what's going on.

The format proposed in this PR is:

     [<timestamp> LEVEL module::path/span1/span2/span3] <msg>

e.g.
```
[2022-08-16T13:58:25.869826Z TRACE sn_node::node::messaging/validate_msg/apply_ae] Entropy check skipped for MsgId(d6aa..ccd7), handling message directly
```

The entire log is on one line making it simple to filter down a large collection of logs using the unix tools.

e.g. `rg -v` can be used to prune out messages we don't care for:

```
tail -f .../sn-node-*/*.log | rg -v 'qp2p|quinn|sn_node::comm'
```
This will filter out logs related to `qp2p`, `quinn`, `sn_node::comm`, you can add to this set interactively as you explore the logs.

### DIFF
--- a/sn_interface/src/lib.rs
+++ b/sn_interface/src/lib.rs
@@ -71,7 +71,7 @@ use tracing_subscriber::{
     fmt::{
         format::Writer,
         time::{FormatTime, SystemTime},
-        FmtContext, FormatEvent, FormatFields, FormattedFields,
+        FmtContext, FormatEvent, FormatFields,
     },
     registry::LookupSpan,
 };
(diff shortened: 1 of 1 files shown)


davidrusu:
Thanks for the pointer, switched the format to follow the env logger default, it seems like a good starting point.

I've also pushed a change to include the spans after the module path. I've updated the OP comment to show the examples

joshuef (approved):

PR #1490, opened 2022-08-16
Cmd clone gone

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
+++ b/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
@@ -159,7 +159,7 @@ impl CmdCtrl {
         let _ = tokio::task::spawn_local(async move {
             match dispatcher.process_cmd(cmd).await {
                 Ok(cmds) => {
-                    monitoring.increment_cmds().await;
+                    // monitoring.increment_cmds().await;
 
                     for cmd in cmds {
                         match cmd_process_api.send((cmd, Some(id))).await {
@@ -192,7 +192,7 @@ impl CmdCtrl {
                         .await;
                 }
             }
-            throughpout.increment(); // both on fail and success
+            // throughpout.increment(); // both on fail and success
         });
         Ok(())
     }
--- a/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
+++ b/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
@@ -30,8 +30,7 @@ use tokio::{sync::RwLock, time::Instant};
 
 type Priority = i32;
 
-// In milliseconds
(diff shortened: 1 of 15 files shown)


sn_node/src/node/flow_ctrl/cmd_ctrl.rs
```diff
@@ -45,167 +44,162 @@ const ORDER: Ordering = Ordering::SeqCst;
...
 
     pub(crate) fn node(&self) -> Arc<RwLock<crate::node::Node>> {
         self.dispatcher.node()
     }
 
-    pub(crate) async fn push(&self, cmd: Cmd) -> Result<SendWatcher> {
+    pub(crate) async fn push(&mut self, cmd: Cmd) -> Result<()> {
         self.push_internal(cmd, None).await
     }
+    pub(crate) fn q_len(&self) -> usize {
```
davidrusu, line 75: cute, but perhaps just spell queue out fully?

sn_node/src/node/flow_ctrl/cmd_ctrl.rs
```diff
@@ -45,167 +44,162 @@ const ORDER: Ordering = Ordering::SeqCst;
...
-#[derive(Clone)]
 pub(crate) struct CmdCtrl {
-    cmd_queue: Arc<RwLock<PriorityQueue<EnqueuedJob, Priority>>>,
+    cmd_queue: PriorityQueue<CmdJob, Priority>,
     attempted: CmdThroughput,
     monitoring: RateLimits,
-    stopped: Arc<RwLock<bool>>,
+    stopped: bool,
     pub(crate) dispatcher: Arc<Dispatcher>,
     id_counter: Arc<AtomicUsize>,
```
davidrusu, line 53: Do these two still need to be behind an `Arc`? Also I don't think you need the AtomicUsize anymore, you could reduce that to `usize` as well

joshuef: Aye, I wasn't sure if i should dive about more there or keep it slim. I can get those cleaned up

sn_node/src/node/flow_ctrl/cmd_ctrl.rs
joshuef: ah, this may actually not be needed now. was more of a debug sitch. I'll srt it

sn_node/src/node/flow_ctrl/cmd_ctrl.rs
joshuef: atomic can be removed easily, dispatcher is harder due to being pulled into the threaded process_cmd. I think we can look at simplifying that down the line 👍

joshuef:
#1502 fixes and advances this. closing this branch

PR #1492, opened 2022-08-17
fix: reintroduce Arc<RwLock> for section tree

b-zee:
The RwLock was mistakenly removed by me. This meant that network updates
to the section tree were not propagated back to the client's session.


bochaco (approved):

davidrusu (approved):
:ship:

PR #1495, opened 2022-08-17
ci: simplify PR flows, increase BORS intensity.

joshuef:
Makes initial PR checks ubuntu only for speed,

moves other checks into bors, including the droplet testnet checks.
This would mean all merges to main have to pass the same, real world scrutiny as the release flow prior.


.github/workflows/bors.yml
```diff
@@ -19,11 +11,118 @@ env:
...
+        timeout-minutes: 25
+        with:
+          test-run-name: cli-unit-tests-${{ matrix.os }}
+          profile: ci
+          junit-path: junit.xml
+          package: sn_cli
+          bin: safe
+          release: true
+
   cli-install-tests:
```
jacderida, line 121: Does the Bors run determine whether we get a green tick on the PR? If so, we may want to ditch the CLI install tests. They fail quite regularly because of Github rate limiting and I wouldn't want us not getting a green tick due to that.

We've had a few bugs in the installs and these tests are useful, but we may need to just run these on a nightly run or something.

.github/workflows/bors.yml
```diff
@@ -1,16 +1,8 @@
-name: Nightly Release Run
+name: Bors Merge Flow. All Tests
 
 on:
-  schedule:
-    # * is a special character in YAML so you have to quote this string
-    - cron:  '30 4 * * *'
-  workflow_dispatch:
```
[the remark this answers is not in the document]
joshuef: yeh, deduplicating some of this would be graaaand. I agree though we can look at it later.

.github/workflows/bors.yml
joshuef: yeh it would be the green tick gatekeeper.

joshuef:
bors try

PR #1512, opened 2022-08-24
feat: return error to client on unknown section key

jacderida:
- 50f3d94f3 **tests(messaging): unit tests for spentbook handler**

  Provide unit test coverage for the `SpentbookCmd::Spent` message handler.

  It's important to note that at this point, the failure cases only assert that no commands were
  returned from the handler, because this is the way we deal with failures at the moment.
  Unfortunately this means it's easy for there to be false positives because you can't check the error
  type or message. I will look into changing this as a separate PR.

  Most of the changes here are related to testing infrastructure:
  * Support setting a threshold when a secret key set is generated for the section. For use with the
    genesis DBC generation, the threshold had to be set to 0.
  * Support adults in the test section. The spent message generates data to be replicated on adults,
    so the mechanisms for creating a test section were extended for this. There are now
    `create_section` and `create_section_with_elders` functions, because some existing tests require
    the condition where only elders have been marked as members.
  * The genesis DBC is needed for these tests, so the scope of the function for generating it was
    changed to `pub(crate)`.
  * The `Cmd` struct was extended in the test module to provide utils to get at the content of
    messages, which are used for test verification.
  * Provide util function for wrapping a `ServiceMsg` inside a `WireMsg` and so on. Keeps the testing
    code cleaner.
  * Provide util function for extracting the spent proof share from the replicated data so that we can
    verify the message handler assigned the correct values to its fields.
  * Various util functions related to the use of DBCs were provided in a `dbc_utils` module. The doc
    comments on the functions should hopefully make clear what they are for.

  A couple of superficial changes were also made to the message handler code:
  * The key image sent by the client is validated (along with a test case for that).
  * Change the format of debugging messages and comments to be more uniform.
  * Move some code into functions scoped at `pub(crate)`. This is so they can be shared for use with
    test setup. For further explanation, see the doc comments on these functions in the diff.

- fdcb0f5b0 **feat: return error to client on unknown section key**

  If one of the spent proofs sent by the client have been signed with a key this section is not
  currently aware of, return an error back to the client.

  This introduces a new SpentProofUnknownSectionKey variant to the messaging data errors, because none
  of the existing variants seemed appropriate for this scenario.

### DIFF
--- a/sn_client/src/connections/messaging.rs
+++ b/sn_client/src/connections/messaging.rs
@@ -708,7 +708,7 @@ mod tests {
         let elders_len = 5;
 
         let prefix = prefix("0")?;
-        let (section_auth, _, secret_key_set) = random_sap(prefix, elders_len);
+        let (section_auth, _, secret_key_set) = random_sap(prefix, elders_len, 0, None);
         let sap0 = section_signed(secret_key_set.secret_key(), section_auth)?;
         let (mut network_contacts, _genesis_sk, _) = new_network_network_contacts();
         assert!(network_contacts.insert_without_chain(sap0));
--- a/sn_interface/src/messaging/data/errors.rs
+++ b/sn_interface/src/messaging/data/errors.rs
@@ -71,6 +71,8 @@ pub enum Error {
     /// Destination is either outdated or incorrect
     #[error("Destination is either outdated or wrong")]
     WrongDestination,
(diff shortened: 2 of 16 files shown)


joshuef (approved):
Nice one @jacderida 💪

bochaco (approved):

PR #1513, opened 2022-08-24
feat: Rename Auth types

dirvine:
BREAKING CHANGE: Using client node and elder as these are more specifically the use case

This is up for discussion. I am claiming it adds to readability, but interested to see what folk think. Also not clear why SectionAuth is not an AuthKind

### DIFF
--- a/sn_node/benches/data_storage.rs
+++ b/sn_node/benches/data_storage.rs
@@ -63,15 +63,21 @@ fn random_vector(length: usize) -> Vec<u8> {
     bytes
 }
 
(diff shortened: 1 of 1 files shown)


joshuef:
I think the auth changes here work. 

Some background: `Service` _messages_ are named such as they may (with health checks eg) originate from `Nodes`. Essentially it's a bit more generic than client w/r/t the messaging. 

Though they are really then acting as `Clients` on the network, though... 

I guess I'm wondering if the messaging naming needs to change too? Or is it okay with a disconnect in auth?

PR #1515, opened 2022-08-25
fix(client): always send_msg_in_bg and log error

joshuef:
never use send_msg directly as it can bubble up errors
unexpectedly. ANy send, should be retrying over errors anyway,
so this should not be an issue


joshuef:
this si actually the wrong direction, we want to expose more errors and be surer of whats going on

PR #1519, opened 2022-08-26
chore(client): send less msgs per second by default

joshuef:
This substantially reduces the load on nodes.
NOTE: Nodes will still need to mitigate high sending clients


joshuef:
rebased atop #1531

PR #1521, opened 2022-08-26
fix(sn_node): avoid testing data collision during bench test

maqi:


sn_node/benches/data_storage.rs
```diff
@@ -63,15 +63,28 @@ fn random_vector(length: usize) -> Vec<u8> {
...
     let mut bytes = BytesMut::from(seed);
     let mut rng = OsRng;
     bytes[0] = rng.gen::<u8>();
     bytes[1] = rng.gen::<u8>();
     bytes[2] = rng.gen::<u8>();
     bytes[3] = rng.gen::<u8>();
+    bytes[4] = rng.gen::<u8>();
+    bytes[5] = rng.gen::<u8>();
+    bytes[6] = rng.gen::<u8>();
+    bytes[7] = rng.gen::<u8>();
```
bochaco, line 87: wouldn't be cleaner to do something like `bytes.iter_mut().take(8).for_each(|b| *b = rng.gen::<u8>());` ?

PR #1524, opened 2022-08-26
feat(statemaps): some minimal infrastructure for generating statemaps

davidrusu:
usage:

0. install `statemap`

       git clone git@github.com:TritonDataCenter/statemap.git
       cd statemap
       cargo install --path .

2. startup a network as usual
3. extract the states file from a log using:

        ./resources/scripts/statemap-preprocess.sh

4. run `statemap` on the generated `safe_states.out` file:

        # vary -c to change the fidelity (max-number of rectangles) larger produces larger SVG's
        statemap -c 300000 safe_states.out > safe.svg

5. open `safe.svg` in a browser

issues with this approach:
  - `statemap` wants all state entry timestamps to be monotonically increasing, my method of extracting states from logs can mess that up when you combine logs form multiple nodes and/or log rotation kicks in
  - would be nicer if we have some state constants instead of using magic numbers throughout the code.

### DIFF
--- a/resources/scripts/statemap-preprocess.sh
+++ b/resources/scripts/statemap-preprocess.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -x
+
+if ! command -v rg &> /dev/null
+then
+    echo "ripgrep could not be found and is required"
+    exit 1
+fi
+
+if ! command -v jq &> /dev/null
+then
+    echo "jq could not be found and is required"
+    exit 1
+fi
+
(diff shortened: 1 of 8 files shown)


b-zee:
Interesting graphs! Is this something to feature gate to prevent too much logging noise?

davidrusu:
@b-zee I've feature flagged the logging, and enabled it by default if you are running through the testnet binary

joshuef (approved):

PR #1527, opened 2022-08-29
feat(sections_dag): replace `SecuredLinkedList` with `SectionsDAG`

RolandSherwin:


sn_node/src/node/error.rs
```diff
@@ -63,6 +63,8 @@ pub enum Error {
     MissingSecretKeyShare(bls::PublicKey),
     #[error("Failed to send a message to {0}")]
     FailedSend(Peer),
+    #[error("Link to peer has been dropped {0}")]
+    PeerLinkDropped(Peer),
```
bochaco, line 67: It seems we don't need this error type?

sn_node/src/node/dkg/session.rs
```diff
@@ -34,7 +34,7 @@ use xor_name::XorName;
 const DKG_PROGRESS_INTERVAL: Duration = Duration::from_secs(6);
 
 // Retry sending messages when they timeout
-const MAX_TIMEOUT_RETRIES: usize = 3;
+const MAX_TIMEOUT_RETRIES: usize = 25;
```
bochaco, line 37: I guess we don't need this change in this PR as they changes shouldn't affect our DKG process?

sn_node/src/node/error.rs
RolandSherwin: Sorry, I might have messed up a rebase.. removed it!

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -185,6 +186,22 @@ impl SectionTree {
...
+                "invalid proof chain: {:?}",
+                partial_dag
+            )));
+        }
+
+        // Check the SAP's key is the last key of the proof chain
+        let is_section_key_last = partial_dag.leaf_keys().contains(&signed_sap.section_key())
+            && partial_dag.leaf_keys().len() == 1;
+        if !is_section_key_last {
+            return Err(Error::MultipleBranchError);
```
bochaco, line 202: I think this error should be returned **only** when the condition `partial_dag.leaf_keys().len() == 1` is not met, the other part of the condition shouldn't be throwing this error type

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -225,62 +242,45 @@ impl SectionTree {
...
-        if !proof_chain.self_verify() {
-            return Err(Error::UntrustedProofChain(format!(
-                "invalid proof chain: {:?}",
-                proof_chain
-            )));
-        }
-
-        // Check the SAP's key is the last key of the proof chain
-        if *proof_chain.last_key() != signed_sap.section_key() {
-            return Err(Error::UntrustedSectionAuthProvider(format!(
```
bochaco, line 257: We still want this check and the error returned can still be the same

bochaco (changes requested):

sn_interface/src/network_knowledge/section_tree/mod.rs
RolandSherwin: Moved this check above the part where we query the SectionTree

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -156,15 +155,15 @@ impl NetworkKnowledge {
...
         genesis_key: bls::PublicKey,
-        chain: SecuredLinkedList,
+        dag: SectionsDAG,
         signed_sap: SectionAuth<SectionAuthorityProvider>,
         passed_section_tree: Option<SectionTree>,
     ) -> Result<Self, Error> {
         // Let's check the section chain's genesis key matches ours.
-        if genesis_key != *chain.root_key() {
-            return Err(Error::UntrustedProofChain(format!(
+        if genesis_key != *dag.genesis_key() {
```
davidrusu, line 163: We have a lot of genesis key validations in this constructor and I think it's an artifact of these 3 parameters we are taking in:
- genesis_key
- SectionsDag
- passed_section_tree

Each of these have a "genesis key" and we need to ensure that they all agree.

It would be better if we had a single parameter that has a genesis key so that we don't have to validate.

It seems that the single param would be the `SectionTree`. Instead of taking an option `passed_section_tree: Option<SectionTree>`, we can force the caller to construct the `SectionTree` before calling the `NetworkKnowledge` constructor.

Don't do this in this PR, we can look at it next.

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -318,73 +325,48 @@ impl NetworkKnowledge {
...
+        // This prevent us from updating our NetworkKnowledge based on an AE message where
+        // we don't have the key share for the new SAP, making this node unable to sign section
+        // messages and possibly being kicked out of the group of Elders.
+        if we_should_become_an_elder && !we_have_a_share_of_this_key {
+            warn!("We should be an elder, but we're missing the keyshare!, ignoring update");
+            return Ok(false);
+        };
+        if we_have_a_share_of_this_key && !we_should_become_an_elder {
+            warn!("We have the keyshare but we are not an elder");
+        };
```
davidrusu, line 356: I don't think it's worth having this check here, it should never happen and we can detect if it happens from the `trace` log above that prints both booleans

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -568,8 +552,8 @@ impl NetworkKnowledge {
     }
 
     /// Return current section chain length
```
davidrusu, line 554:
```suggestion
    /// Return the number of keys in our section dag
```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -578,11 +562,12 @@ impl NetworkKnowledge {
     }
 
     /// Return a vec of all known keys
```
davidrusu, line 564:
```suggestion
    /// Return the set of known keys
```

sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -280,12 +280,19 @@ pub mod test_utils {
...
     use eyre::{eyre, Result};
     use itertools::Itertools;
+    use rand::RngCore;
     use serde::Serialize;
     use sn_consensus::{Ballot, Consensus, Decision, Proposition, Vote, VoteResponse};
     use std::collections::BTreeMap;
     use std::{cell::Cell, net::SocketAddr};
     use xor_name::Prefix;
 
+    // Get `Prefix` from st
```
davidrusu, line 290:
```suggestion
    // Parse `Prefix` from string
```

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -167,24 +168,45 @@ impl SectionTree {
...
+                "Invalid partial_dag: {:?}",
+                partial_dag
+            )));
+        }
+
+        // Check the SAP's key is the last key of the proof chain
+        if partial_dag.leaf_keys().len() != 1 {
+            return Err(Error::MultipleBranchError);
+        }
+        if !partial_dag.leaf_keys().contains(&signed_sap.section_key()) {
```
davidrusu, line 202: I think you can combine these last two checks with
```suggestion
        if partial_dag.leaf_keys() != BTreeSet::from_iter([signed_sap.section_key()]) {
```
Also make sure to log `partial_dag.leaf_keys()` in the error msg string as well.

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -225,62 +247,45 @@ impl SectionTree {
...
+        if self.insert(signed_sap) {
+            // update our sections_dag with the proof_chain. Cannot be an error, since in cases where we
+            // have outdated SAP (aware of prefix)/ not aware of the prefix, we have the proof_chains's
+            // root/child key in our sections_dag. Checked in the above match statement.
+            self.sections_dag.merge(partial_dag.clone())?;
+            for (prefix, section_key) in &self.sections {
+                debug!(
+                    "Known prefix,section_key after update: {:?} = {:?}",
+                    prefix, section_key
+                );
```
davidrusu, line 281: nit. save some vertical space
```suggestion
                debug!("Known prefix, section_key after update: {prefix:?} = {section_key:?}");
```

sn_interface/src/network_knowledge/section_tree/mod.rs
RolandSherwin: Since `partial_dag` can contain a tree with more than one branch, the `MultipleBranchError` was introduced to specify that we expect a dag with just 1 branch. So, is it ok to separate out the errors?

sn_interface/src/network_knowledge/mod.rs
RolandSherwin: Yeah, will try to include it in the following cleanup PR.

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -50,17 +53,87 @@ impl Sha3Hash for SectionInfo {
...
+impl<'de> Deserialize<'de> for SectionsDAG {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct Intermediate {
+            genesis_key: bls::PublicKey,
+            sections: Vec<(bls::PublicKey, SectionInfo)>,
+        }
```
davidrusu, line 102: Do you want to move this struct out, derive Serialize for it and use it for the serializer as well at https://github.com/maidsafe/safe_network/pull/1527/files#diff-600741ddfc9d4fdd3347d88539faa1a20b920bab3d6349e726c67ca851c075e8R86-R89?

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -50,17 +53,87 @@ impl Sha3Hash for SectionInfo {
...
+                None => parent_indices.push(usize::MAX),
+                Some(parent_key) => {
+                    let idx = keys
+                        .iter()
+                        .position(|key| *key == parent_key)
+                        .ok_or(fmt::Error)?;
+                    parent_indices.push(idx)
+                }
+            }
+        }
```
davidrusu, line 129: That parent key position() call will make this debug impl quadratic as well, and I imagine we may be logging this quite frequently. Lets pre-index the keys to avoid any slowdown.
```suggestion
        let keys: Vec<_> = self.keys().cloned().collect();
        let key_positions: BTreeMap<&PublicKey, usize> = keys.iter().enumerate().collect();
        let mut parent_indices: Vec<usize> = Vec::new();
        for key in keys.iter() {
            match self.get_parent_key(key).map_err(|_| fmt::Error)? {
                None => parent_indices.push(usize::MAX),
                Some(parent_key) => {
                    let idx = key_positions.get(&parent_key).ok_or(fmt::Error)?;
                    parent_indices.push(idx)
                }
            }
        }
```

sn_node/src/node/messaging/mod.rs
```diff
@@ -182,7 +182,7 @@ impl Node {
     /// based on our current knowledge of the network and sections chains.
     #[instrument(skip_all)]
     async fn verify_section_key(&self, msg_authority: &NodeMsgAuthority, msg: &SystemMsg) -> bool {
-        let known_keys = self.network_knowledge.known_keys();
+        let known_keys: Vec<_> = self.network_knowledge.known_keys().into_iter().collect();
         NetworkKnowledge::verify_node_msg_can_be_trusted(msg_authority, msg, &known_keys)
```
davidrusu, line 186: Consider modifying `verify_node_msg_can_be_trusted` to take the set rather than the vec. It should improve validation perf. (and remove the vec conversion)

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -50,17 +53,87 @@ impl Sha3Hash for SectionInfo {
...
+        let mut need_to_visit: Vec<(bls::PublicKey, Node<SectionInfo>)> = Vec::new();
+        for key in self.dag_root.iter() {
+            need_to_visit.push((
+                *self.genesis_key(),
+                self.get_node(key).map_err(S::Error::custom)?,
+            ));
+        }
+        while let Some((parent_key, current_node)) = need_to_visit.pop() {
+            need_to_visit.extend(
+                self.child_nodes(current_node.hash())
```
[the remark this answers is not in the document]
RolandSherwin: The order is important because we perform `insert()` while deserializing, and we want to make sure we insert the parent before the child. 

Traversing from the root makes sense, but we will be adding a lot of duplicate elements to the `Vec` as we might visit the same parent multiple times. Do you think it's worth keeping a separate `BTreeSet` which we can query to check if a parent `SectionInfo` has been inserted into the Vec, if so we can go ahead with the next leaf.

We have the same logic in `self_verify`
```rust
// we stop when parent is verified; meaning all ancestors are verified.
if verified_keys.contains(&parent) {
    break true;
}
```

sn_node/src/node/mod.rs
```diff
@@ -704,7 +699,21 @@ mod core {
...
                 // us back their whole section chain. However, this situation should be rare.
-                *self.network_knowledge.our_section_dag().prev_key()
+
+                // our_section_dag produces a DAG with only 1 leaf
+                let leaf_key = self
+                    .network_knowledge
+                    .our_section_dag()
+                    .leaf_keys()
+                    .into_iter()
+                    .collect::<Vec<_>>()[0];
```
davidrusu, line 709: This seems to be a common pattern, lets make it a method

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Yeah, that would make more sense! Changed it!

sn_interface/src/network_knowledge/sections_dag.rs
RolandSherwin: Updated it!

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -50,17 +49,100 @@ impl Sha3Hash for SectionInfo {
...
+                sections.push(section);
+            })
+        }
+        let inter = Intermediate {
+            genesis_key: self.genesis_key,
+            sections,
+        };
+        inter.serialize(serializer)
+    }
+}
```
davidrusu, line 104: Good stuff 👍

davidrusu (approved):
👍 looks good

sn_interface/src/network_knowledge/section_tree/mod.rs
davidrusu: ok, yeah makes sense, I saw you testing for that error variant in some test cases.

sn_interface/src/network_knowledge/sections_dag.rs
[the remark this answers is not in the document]
davidrusu: The new implementation you did looks good 👍

PR #1535, opened 2022-09-01
Statemap: Align timelines of all nodes in the network

davidrusu:
- also breakdown the SystemMsg state into the various types of system messages
- move the meat of the statemap module into sn_interface so that we have access to the States enum from more modules.

This is a 15 node network starting up - the gray bars at the beginning of each timeline are from before the node had joine the network, the white is idle time after the node had joined the network:
aligned-statemap


joshuef (approved):
Awesome. 

It'll be good to get this looped into CI runs and uploaded. Could be a handy initial indicator of what is off w/  a network

PR #1536, opened 2022-09-01
Ensure adults are probing for AE updates before first split

davidrusu:
Adults were skipping AE probes before splits for some reason

Also included in this PR:
- move AE probe building to network knowledge
- remove the unused `stopped` state from cmd ctrl which removes some error cases for us.

### DIFF
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -312,6 +312,10 @@ impl NetworkKnowledge {
         true
     }
 
+    pub fn anti_entropy_probe(&self) -> SystemMsg {
+        SystemMsg::AntiEntropyProbe(self.section_key())
+    }
+
     /// Update our network knowledge if the provided SAP is valid and can be verified
     /// with the provided proof chain.
     /// If the '`update_sap`' flag is set to 'true', the provided SAP and chain will be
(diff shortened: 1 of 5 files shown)


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -312,6 +312,10 @@ impl NetworkKnowledge {
         true
     }
 
+    pub fn anti_entropy_probe(&self) -> SystemMsg {
+        SystemMsg::AntiEntropyProbe(self.section_key())
```
davidrusu, line 316: The AE probe will grow in the future to include the membership generation, just moving it here in prep for that.

sn_node/src/node/flow_ctrl/cmd_ctrl.rs
```diff
@@ -78,32 +76,13 @@ impl CmdCtrl {
         !self.cmd_queue.is_empty()
     }
 
-    // consume self
-    // NB that clones could still exist, however they would be in the disconnected state
```
davidrusu, line 82: This used to be true when we were doing multi-threading, no longer true now. There is just a single `CmdCtrl` reference alive for the duration of the node process.

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -235,20 +226,12 @@ impl FlowCtrl {
...
-
-        let prefix = node.network_knowledge().prefix();
-
-        // Send a probe message to an elder
-        if !prefix.is_empty() {
-            Some(node.generate_section_probe_msg())
-        } else {
-            None
-        }
+        node.generate_section_probe_msg()
```
davidrusu, line 234: For some reason we were not performing probes before the first split

joshuef (approved):

PR #1545, opened 2022-09-05
fix: always resend query to adults

joshuef:
Remove a potential source of errors. We _always_ send on queries now. Regardless if the client already exists in `pending_queries`. There's a reason they're re-requesting.

When we have cache implemented at elders we can short circuit this there... but otherwise we should treat each request equally.

### DIFF
--- a/sn_node/src/node/data/records.rs
+++ b/sn_node/src/node/data/records.rs
@@ -103,10 +103,6 @@ impl Node {
             if peers.len() > MAX_WAITING_PEERS_PER_QUERY {
                 warn!("Dropping query from {source_client:?}, there are more than {MAX_WAITING_PEERS_PER_QUERY} waiting already");
                 return Ok(vec![]);
-            } else {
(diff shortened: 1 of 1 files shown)


sn_node/src/node/data/records.rs
```diff
@@ -103,10 +103,6 @@ impl Node {
             if peers.len() > MAX_WAITING_PEERS_PER_QUERY {
                 warn!("Dropping query from {source_client:?}, there are more than {MAX_WAITING_PEERS_PER_QUERY} waiting already");
                 return Ok(vec![]);
```
bochaco, line 105: Shouldn't we be returning the `cmds` here in order to always resend the queries?

joshuef: I think we're alright here. As because we've hit the limit, we don\t want to query again.

bochaco (approved):

PR #1546, opened 2022-09-05
chore: sign query msg outside of send loop

joshuef:
This should reduce client cpu load qhen attempting to retrieve data.
We sign only once, instead of on each attempt.


bochaco (changes requested):
I think we are breaking the change in adult index L127 ? 
```
query.adult_index += 1;
```

PR #1550, opened 2022-09-06
SDKG integration

grumbach:
- integrates `sn_sdkg` and removes `bls_dkg`


sn_node/src/node/mod.rs
```diff
@@ -554,6 +553,32 @@ mod core {
...
 
+            // clean up DKG sessions 2 generations older than current
+            // `session_id.section_chain_len + 2 < current_chain_len`
+            // we voluntarily keep the previous DKG rounds
+            // so lagging elder candidates can still get responses to their gossip.
+            // At generation+2, they are not going to be elders anymore so we can safely discard it
+            let current_chain_len = self.network_knowledge.section_chain_len();
+            let mut old_hashes = vec![];
+            for (hash, session_info) in self.dkg_sessions_info.iter() {
+                if session_info.session_id.section_chain_len + 5 < current_chain_len {
```
grumbach, line 564: Ideally we should only need +2, but for now, before we fix the AE issue after selfish terminations +5 is a safer bet.

sn_interface/src/messaging/system/agreement.rs
```diff
@@ -66,6 +65,11 @@ impl DkgSessionId {
         hash
     }
 
+    /// A small checksum of the session id's hash used for logging as it is very short
+    pub fn sum(&self) -> u16 {
+        self.hash().iter().map(|i| *i as u16).sum::<u16>()
+    }
+
```
davidrusu, line 72: Give this a more descriptive name, something like `dbg_short_hash`. `sum` makes it seem like something general / important.

Also, you'll get fewer collisions if you just take the prefix of the hash, i.e. if you want 16 bits, take the first two bytes of the hash.
```suggestion
    /// A small checksum of the session id's hash used for logging as it is very short
    pub fn dbg_short_hash(&self) -> u16 {
        u16::from_le_bytes(self.hash()[..2])
    }

```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -94,6 +94,12 @@ pub const fn supermajority(group_size: usize) -> usize {
     1 + group_size * 2 / 3
 }
 
+/// `Threshold` of a given group
+#[inline]
+pub const fn threshold(group_size: usize) -> usize {
+    supermajority(group_size) - 1
+}
```
davidrusu, line 101: I don't like seeing subtraction with `usize`, Let's define `supermajority` in terms of `threshold` to get rid of it.

Currently `supermajority = 1 + group_size * 2 / 3` and `threshold = supermajority - 1`
Instead do `supermajority = threshold + 1` and `threshold = group_size * 2 / 3`

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+    // last dkg message timestamp
+    last_received_dkg_msg_time: Option<Instant>,
+}
+
+/// Helper that creates a dkg state
+fn create_dkg_state(
+    session_id: &DkgSessionId,
+    participant_index: usize,
+    secret_key: BlsSecretKey,
+    pub_keys: DkgPubKeys,
```
davidrusu, line 53: There's a few things named some variation of `public-keys` in this function and it's kind of hard to track. Can we rename some by functionality?

```suggestion
    node_key_upgrades: DkgPubKeys,
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+) -> Result<DkgState> {
+    let mut rng = bls::rand::rngs::OsRng;
+    let threshold = threshold(session_id.elders.len());
+    let mut public_keys: BTreeMap<NodeId, BlsPublicKey> = BTreeMap::new();
+    for (xorname, (pubkey, _)) in pub_keys.iter() {
```
davidrusu, line 58:
```suggestion
    for (xorname, (ephemeral_bls_pk, _)) in pub_keys.iter() {
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+        participant_index as u8,
+        secret_key,
+        public_keys,
+        threshold,
+        &mut rng,
+    )?)
+}
+
+// Helper that checks an ephemeral pubkey
+pub(crate) fn check_key(
```
davidrusu, line 75:
```suggestion
pub(crate) fn check_ephemeral_dkg_key(
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+    }
+
+    // check sig
+    let sender_pubkey = pub_key(&key_owner).map_err(|_| Error::InvalidXorname(key_owner))?;
+    debug!(
+        "Checking dkg ephemeral key s{} from {:?}",
+        session_id.sum(),
+        key_owner
+    );
+    let serialized = bincode::serialize(&key)?;
```
davidrusu, line 93:
```suggestion
    let serialized_key = bincode::serialize(&key)?;
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+    ) -> Result<(BlsPublicKey, Signature)> {
+        // error out if we already have a key
+        if self.dkg_ephemeral_keys.get(&session_id_hash).is_some() {
+            return Err(Error::DkgEphemeralKeyAlreadyGenerated);
+        }
+
+        // gen new key
+        let new_secret_key: BlsSecretKey = bls::rand::random();
+        let new_pub_key = new_secret_key.public_key();
+        let serialized = bincode::serialize(&new_pub_key)?;
```
davidrusu, line 126:
```suggestion
        let serialized_key = bincode::serialize(&new_pub_key)?;
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+        let sig = types::keys::ed25519::sign(&serialized, keypair);
```
davidrusu, line 127:
```suggestion
        let key_sig = types::keys::ed25519::sign(&serialized, keypair);
```

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+use std::sync::Arc;
+use std::time::Instant;
+use xor_name::XorName;
+
+pub(crate) type DkgPubKeys = BTreeMap<XorName, (BlsPublicKey, Signature)>;
+
+pub(crate) struct DkgEphemeralKeys {
+    secret_key: BlsSecretKey,
+    pub_keys: DkgPubKeys,
+}
```
davidrusu, line 34: Why do we store the `pub_keys` as a map? I'd imagine this struct would be:
```suggestion
pub(crate) struct DkgEphemeralKeys {
    owner: XorName,
    secret_key: BlsSecretKey,
    signature: Signature
}
```
You don't need to hold the public key since it can be re-computed from the secret_key with `secret_key.public_key()`

sn_node/src/node/dkg.rs
```diff
@@ -0,0 +1,362 @@
...
+        let sig = types::keys::ed25519::sign(&serialized, keypair);
+        let ephemeral_keys = DkgEphemeralKeys {
+            secret_key: new_secret_key,
+            pub_keys: BTreeMap::from_iter([(our_name, (new_pub_key, sig))]),
+        };
+
+        // insert the key if we don't already have it
+        let _did_insert = self
+            .dkg_ephemeral_keys
+            .insert(session_id_hash, ephemeral_keys);
```
davidrusu, line 136: Is the comment outdated? This code always inserts the new ephemeral key.

grumbach: That comment line is probably not very useful, let me remove it. It's redundant with the bail out above if we already have an entry for this hash.

sn_interface/src/messaging/system/agreement.rs
grumbach: or maybe `sh()`

sn_node/src/node/messaging/dkg.rs
```diff
@@ -107,176 +150,567 @@ impl Node {
...
+        );
+        let peers = dkg_peers(our_id, session_id);
+        let node_msg = SystemMsg::DkgVotes {
+            session_id: session_id.clone(),
+            pub_keys,
+            votes: vec![vote],
         };
-        self.send_system_msg(msg, Peers::Single(sender))
+        let cmd = self.send_system_msg(node_msg, Peers::Multiple(peers));
+        Ok(vec![cmd])
```
davidrusu, line 369: Does this initial vote also get processed by us at some point?

sn_node/src/node/messaging/dkg.rs
```diff
@@ -107,176 +150,567 @@ impl Node {
...
+            session_id.sh()
+        );
+        let peers = dkg_peers(our_id, session_id);
+        let node_msg = SystemMsg::DkgVotes {
+            session_id: session_id.clone(),
+            pub_keys,
+            votes: vec![vote],
         };
-        self.send_system_msg(msg, Peers::Single(sender))
+        let cmd = self.send_system_msg(node_msg, Peers::Multiple(peers));
```
davidrusu, line 368: Can this be a call to `broadcast_dkg_votes(..)`?

davidrusu (approved):
nice work :1st_place_medal:

sn_node/src/node/messaging/dkg.rs
grumbach: Oh good one!

davidrusu:
@grumbach you've included a very large `save.svg` statemap! lets remove that before merging.

bors cancel

PR #1556, opened 2022-09-06
chore(client): retry DataNotFound errors for data_copy_count

joshuef:

### DIFF
--- a/sn_client/src/api/queries.rs
+++ b/sn_client/src/api/queries.rs
@@ -6,6 +6,8 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
+use std::collections::BTreeSet;
+
 use super::Client;
 use crate::{connections::QueryResult, errors::Error};
 
@@ -67,6 +69,8 @@ impl Client {
         let _ = span.enter();
         let mut attempts = 1;
         let dst = query.variant.dst_name();
+
+        let mut data_not_found_count = BTreeSet::default();
         loop {
             let msg = ServiceMsg::Query(query.clone());
             let serialised_query = WireMsg::serialize_msg_payload(&msg)?;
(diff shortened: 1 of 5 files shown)


sn_interface/src/messaging/data/mod.rs
```diff
@@ -247,6 +247,24 @@ impl QueryResponse {
         }
     }
 
+    /// Returns true if the result returned is DataNotFound
```
bochaco, line 250: I think this is cleaner?
```
matches!(
            self,
            GetChunk(Err(ErrorMsg::DataNotFound(_)))
                | GetRegister((Err(ErrorMsg::DataNotFound(_)), _))
                | GetRegisterEntry((Err(ErrorMsg::DataNotFound(_)), _))
                | GetRegisterOwner((Err(ErrorMsg::DataNotFound(_)), _))
                | ReadRegister((Err(ErrorMsg::DataNotFound(_)), _))
                | GetRegisterPolicy((Err(ErrorMsg::DataNotFound(_)), _))
                | GetRegisterUserPermissions((Err(ErrorMsg::DataNotFound(_)), _))
                | SpentProofShares((Err(ErrorMsg::DataNotFound(_)), _))
        )
```

sn_client/src/api/queries.rs
```diff
@@ -93,7 +95,23 @@ impl Client {
             .await;
 
             match res {
-                Ok(Ok(query_result)) => break Ok(query_result),
+                Ok(Ok(query_result)) => {
+                    if !query_result.is_data_not_found() {
+                        data_not_found_count += 1;
```
bochaco, line 100: The var name seems to imply the condition is incorrectly negated?

sn_client/src/api/queries.rs
```diff
@@ -93,7 +95,23 @@ impl Client {
...
+                        // if we've tested enough adults, we can return failure
+                        // we do this twice to be sure connection issues were not involved
```
bochaco, line 102: Do we mean here connection issues between Elders and Adults? I assume connection issues from client to Elders won't return a data-not-found error, correct?

joshuef: yeh, issues communicating to elders / a timeout/cleanup issue eg. I'll reword here

sn_client/src/api/queries.rs
joshuef: ups, yeh i shuffled the order but forgot to remove the `!`. good one!

sn_client/src/api/queries.rs
joshuef: Actually this is wrong (I think i said this yesterday in the meet, and then forgot about it 🤦 ). 

We're _only_ tracking `DataNotFound` here. Connection errors dont come into it. I've updated it to track each index's DataNotFound... and we'd loop only until we have DataNotFound at each of them

sn_interface/src/messaging/data/mod.rs
bochaco: @joshuef my suggestion was for the `is_data_not_found` function rather than the `is_success` which was ok but now will return the opposite value

bochaco (approved):

PR #1557, opened 2022-09-07
feat: retry dbc spend on unknown section key

jacderida:
- bdcbeefd6 **feat: dbc spend can update network knowledge**

  When a DBC is submitted for spending, there are a set of spent proofs sent along with the request,
  one for each input for the DBC being spent. It's possible that these proofs were signed with a
  section key that the section processing the request is not aware of. In this case, the request
  processing section needs to be updated with new network knowledge. Therefore, we add a couple of
  fields to the spend command to give it the option to send new network knowledge, namely the updated
  proof chain and the signed section authority provider.

  If the node detects the spend command has been sent with these, it will create a new type of
  internal command, `UpdateNetworkAndHandleValidServiceMsg` which will be returned for processing.
  This command will contain the original spend request, which will be run after the network knowledge
  is updated. The reason for issuing the new command is because the original `HandleValidServiceMsg`
  command does not have mutable access to the node.

- 14c82ca14 **feat: retry dbc spend on unknown section key**

  This is the client side for the scenario where the spent proofs are signed by section keys that the
  processing section is not aware of.

  Several retries will be attempted because it's possible for there to be multiple section keys that
  are not known, but the network will only return back one key at a time.

- 9e8bcae5d **tests: spend with updated network knowledge**

  Previously I had a placeholder in for this case, but now have something working.

  The test requires having two network sections and one of the input DBCs for a transaction being
  signed by the other section key.

  The `TestNodeBuilder` was extended with a function that creates a section without a creating a node,
  and this included being able to provide a section chain and tree.

- 8422307a7 **chore: remove spend retry on client**

  The spend retry depends on providing new network knowledge. We will be using another mechanism to
  obtain this knowledge, which is not available at the moment. Once it's available, we'll add the
  retry again.

  For now we decided it's best to remove it and only merge the node-side changes.

### DIFF
--- a/sn_client/src/api/spentbook_apis.rs
+++ b/sn_client/src/api/spentbook_apis.rs
@@ -12,10 +12,7 @@ use crate::Error;
 
 use sn_dbc::{KeyImage, RingCtTransaction, SpentProof, SpentProofShare};
 use sn_interface::{
-    messaging::data::{
-        DataCmd, DataQueryVariant, Error as NetworkDataError, QueryResponse, SpentbookCmd,
-        SpentbookQuery,
-    },
+    messaging::data::{DataCmd, DataQueryVariant, QueryResponse, SpentbookCmd, SpentbookQuery},
     types::SpentbookAddress,
 };
 
@@ -45,66 +42,14 @@ impl Client {
         spent_proofs: BTreeSet<SpentProof>,
         spent_transactions: BTreeSet<RingCtTransaction>,
     ) -> Result<(), Error> {
-        let mut cmd = SpentbookCmd::Spend {
+        let cmd = SpentbookCmd::Spend {
             key_image,
             tx: tx.clone(),
             spent_proofs: spent_proofs.clone(),
             spent_transactions: spent_transactions.clone(),
             network_knowledge: None,
         };
-        let mut attempts = 1;
-        debug!("Attempting DBC spend request.");
-        debug!(
-            "Will reattempt if spent proof was signed with a section key that is unknown to the \
-            processing section."
-        );
-        loop {
-            let result = self.send_cmd(DataCmd::Spentbook(cmd)).await;
-            match result {
-                Ok(()) => {
-                    return Ok(());
-                }
-                Err(err) => match err {
-                    Error::ErrorCmd {
-                        source:
-                            NetworkDataError::SpentProofUnknownSectionKey(
-                                current_section_key,
-                                unknown_section_key,
-                            ),
-                        ..
-                    } => {
-                        debug!("Encountered unknown section key during spend request.");
-                        debug!(
-                            "Will obtain updated network knowledge and retry. \
-                                Attempts made: {attempts}"
-                        );
-                        attempts += 1;
-                        if attempts > 5 {
-                            error!("DBC spend request failed after 5 retry attempts");
-                            return Err(Error::DbcSpendRetryAttemptsExceeded);
-                        }
-                        let network = self.session.network.read().await;
-                        let proof_chain = network
-                            .get_sections_dag()
-                            .get_proof_chain(&current_section_key, &unknown_section_key)?;
-                        let signed_sap = network
-                            .get_signed_by_key(&unknown_section_key)
-                            .ok_or(Error::SignedSapNotFound(unknown_section_key))?;
-                        cmd = SpentbookCmd::Spend {
-                            key_image,
-                            tx: tx.clone(),
-                            spent_proofs: spent_proofs.clone(),
-                            spent_transactions: spent_transactions.clone(),
-                            network_knowledge: Some((proof_chain, signed_sap.clone())),
-                        };
-                        continue;
-                    }
-                    _ => {
-                        return Err(err);
-                    }
-                },
-            }
-        }
+        self.send_cmd(DataCmd::Spentbook(cmd)).await
     }
 
     //----------------------
--- a/sn_client/src/api/spentbook_apis.rs
+++ b/sn_client/src/api/spentbook_apis.rs
@@ -12,7 +12,10 @@ use crate::Error;
 
 use sn_dbc::{KeyImage, RingCtTransaction, SpentProof, SpentProofShare};
 use sn_interface::{
-    messaging::data::{DataCmd, DataQueryVariant, QueryResponse, SpentbookCmd, SpentbookQuery},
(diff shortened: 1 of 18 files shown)


sn_client/src/api/spentbook_apis.rs
```diff
@@ -33,13 +45,69 @@ impl Client {
...
+            processing section."
+        );
+        loop {
+            let result = self.send_cmd(DataCmd::Spentbook(cmd)).await;
+            match result {
+                Ok(()) => {
+                    break;
+                }
+                Err(err) => match err {
+                    Error::ErrorCmd {
```
bochaco, line 69: You can simply have a matching arm like `Err(Error::ErrorCmd { source: ... })` to avoid  the inner `match`

sn_client/src/api/spentbook_apis.rs
```diff
@@ -33,13 +45,69 @@ impl Client {
...
+                            tx: tx.clone(),
+                            spent_proofs: spent_proofs.clone(),
+                            spent_transactions: spent_transactions.clone(),
+                            updated_proof_chain: Some(proof_chain),
+                            signed_sap: Some(signed_sap.clone()),
+                        };
+                        continue;
+                    }
+                    _ => {
+                        return Err(err);
```
bochaco, line 105: I think we can do in both cases where we stop the loop `break Ok(())` / `break Err(err)`, just for consistency?

sn_interface/src/messaging/data/errors.rs
```diff
@@ -53,7 +53,10 @@ pub enum Error {
...
     InvalidQueryResponseErrorForOperationId,
-    /// Failed to verify a spent proof since it's signed by unknown section key
-    #[error("Spent proof was signed with unknown section key: {0:?}")]
-    SpentProofUnknownSectionKey(bls::PublicKey),
+    /// A DBC spend request could not be processed because the processing section was unaware of
+    /// the section that signed one of the input spent proofs.
+    #[error(
+        "Spent proof is signed by section key {0:?} that is unknown to the current section {1:?}"
+    )]
+    SpentProofUnknownSectionKey(bls::PublicKey, bls::PublicKey),
```
bochaco, line 61: The error msg seems to suggest the known key belongs to the section processing the request rather than the known section key of the remote section which signed the proof. I'm not sure we can even know that from the processing nodes, if a key is unknown to the Elders I don't think they can find out which section or prefix it's supposed to belong to, ....we may need the client to send the whole proof chain from genesis key to the latest key of the section that signed the spent proof.....

sn_interface/src/messaging/data/spentbook.rs
```diff
@@ -36,6 +39,8 @@ pub enum SpentbookCmd {
         tx: RingCtTransaction,
         spent_proofs: BTreeSet<SpentProof>,
         spent_transactions: BTreeSet<RingCtTransaction>,
+        updated_proof_chain: Option<SecuredLinkedList>,
+        signed_sap: Option<SectionAuth<SectionAuthorityProvider>>,
```
bochaco, line 43: Perhaps we can have a single field with both values as I don't think providing only one them is a possibility, it's either both of them or none, so e.g.: `sap_proof: Option<SecuredLinkedList, SectionAuth<SectionAuthorityProvider>>`

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -161,6 +161,19 @@ impl SectionTree {
...
+        &self,
+        section_key: &bls::PublicKey,
+    ) -> Option<&SectionAuth<SectionAuthorityProvider>> {
+        for (_, signed_sap) in self.sections.iter() {
+            if signed_sap.public_key_set().public_key().to_hex() == section_key.to_hex() {
+                return Some(signed_sap);
+            }
+        }
+        None
+    }
```
bochaco, line 175: We don't need to compare the hex values but just the public keys ?
We may need something more complex here since the key we are trying to get the SAP and chain for may not be the latest of the section branch it belongs to. E.g. let's say the key belonged to a section with prefix `0101`, but at the moment we are querying the SAP (and chain) that section split into `01010` and `01011`, we won't have any key in our SAPs (`sections`) list matching the section key we are looking for.

sn_client/src/api/spentbook_apis.rs
jacderida: Hey, sorry, can I just clarify what you mean here please? Are you saying to `break` here rather than return the error?

The break would stop the loop and then return `Ok`. The intention here was that if there was any error other than `SpentProofUnknownSectionKey` that we would just return that error back as we would have done anyway. Do you think we should break out and return `Ok` instead?

sn_interface/src/messaging/data/spentbook.rs
jacderida: Yeah thanks, that makes sense.

sn_interface/src/messaging/data/errors.rs
jacderida: As discussed, will change this error back to returning one key.

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1607,67 +1610,129 @@ async fn spentbook_spend_with_random_key_image_should_return_spentbook_error() -
...
                 )?,
                 &dispatcher,
             )
             .await?;
 
+            // The commands returned here should include the new command to update the network
+            // knowledge and also the other two commands to replicate the spent proof shares and
+            // the ack command, but we've already validated the other two as part of another test.
             assert_eq!(cmds.len(), 3);
             let update_cmd = cmds[0].clone();
```
bochaco, line 1706: Perhaps we need to assume it may not be the first in the list of commands

sn_node/src/node/messaging/service_msgs.rs
```diff
@@ -393,7 +419,10 @@ impl Node {
                     key {:?}",
                     pk
                 );
-                return Err(Error::SpentProofUnknownSectionKey(*pk));
+                return Err(Error::SpentProofUnknownSectionKey(
+                    *pk,
+                    self.network_knowledge.section_key(),
```
bochaco, line 424: Let's remove this second key from the error type as it's not required and not needed.

bochaco (approved):

PR #1577, opened 2022-09-14
test: verify entry_hash generation

maqi:

### DIFF
--- a/sn_client/src/api/register_apis.rs
+++ b/sn_client/src/api/register_apis.rs
@@ -486,6 +486,62 @@ mod tests {
     }
 
     #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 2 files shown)


sn_client/src/api/register_apis.rs
```diff
@@ -485,6 +485,58 @@ mod tests {
         }
     }
 
+    #[tokio::test(flavor = "multi_thread")]
+    async fn register_write_without_publish() -> Result<()> {
```
joshuef, line 489: register_hashes_dont_clash perhaps?

Some comments throughout the test to explain what's going on would be worthwhile here, too 🙇

joshuef (approved):

PR #1583, opened 2022-09-15
Remove rustflags

grumbach:


davidrusu:
Even though we're not seeing a built time improvement, I think it's still useful to remove this flag. We have clippy checks to detect unaddressed warnings already, sometimes it's useful to just push something up to see how CI reacts before fixing all the compiler warnings

joshuef (approved):

PR #1590, opened 2022-09-19
Refactor op id in elders from msg payload bytes

bochaco:
- Generate query/cmd `OperationId` from the msg serialised payload at Elders
- Redefine system::NodeQueryResponse type just as an alias to data::QueryResponse
- Removing `OperationId` from `ServiceMsg` msgs
- Removing `OperationId` from sn_client queries/responses tracker
- Removing `correlation_id` from `SystemMsg` node query/response msgs
- Returning error to client when an Elder reaches a maximum number of waiting peers (currently 100) for the same query response


bochaco:
Superseded by #1591

PR #1591, opened 2022-09-20
Generate query/cmd OperationId from the query msg id when received at Elders

bochaco:
- Generate query/cmd `OperationId` from the query msg id when received at Elders
- Redefine system::NodeQueryResponse type just as an alias to data::QueryResponse
- Removing `OperationId` from `ServiceMsg` msgs
- Removing `OperationId` from sn_client queries/responses tracker
- Removing `correlation_id` from `SystemMsg` node query/response msgs
- Remove peers from pending data queries in Elders when response was obtained from Adults
- Returning error to client when an Elder reaches a maximum number of waiting peers (currently 100) for the same query response


joshuef:
closing, https://github.com/maidsafe/safe_network/pull/1595 is this but atop latest ConnQStability branch after rebase w/ main

PR #1592, opened 2022-09-20
Debugging single threaded client tests no retry

bochaco:


joshuef:
#1596 is this updated atop latest CQS

PR #1594, opened 2022-09-21
Move "will-be-elder" check from `network_knowledge` to `sn_node`

davidrusu:
By moving these checks into `sn_node`, we no longer need a `SectionKeysProvider` to update `NetworkKnowledge`.

This makes it easier to use network_knowledge from outside of `Node`. For example, in bootstrapping, we can create a `NetworkKnowledge` instance and bring it up to date before we construct a full `Node` instance.

### DIFF
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -122,7 +122,6 @@ impl<'a> Joiner<'a> {
             "Bootstrap run, network contacts as we have it: {:?}",
             self.network_contacts
         );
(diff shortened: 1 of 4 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -148,7 +143,6 @@ impl<'a> Joiner<'a> {
     #[tracing::instrument(skip(self))]
     async fn join(
         mut self,
-        network_genesis_key: BlsPublicKey,
```
davidrusu, line 151: This was unused.

joshuef (approved):

PR #1606, opened 2022-09-22
Remove bootstrap node from join process; remove unneeded error handling in Comms

davidrusu:
Two changes
1. remove some unnecessary error handling in Comm
2. Remove the `bootstrap_node` argument to the join process.

The `bootstrap_node: SocketAddr` was there as a backup in case we did not have any peer entries in our SectionTree, but if you trace the logic all the way through, you can see that the `bootstrap_node` was being pulled from our `SectionTree`! So the fallback address is a fallback to itself?

### DIFF
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -48,7 +48,7 @@ pub(crate) async fn join_network(
     network_contacts: SectionTree,
     join_timeout: Duration,
 ) -> Result<(NodeInfo, NetworkKnowledge)> {
-    let (outgoing_msgs_sender, outgoing_msgs_receiver) = mpsc::channel(1);
+    let (outgoing_msgs_sender, outgoing_msgs_receiver) = mpsc::channel(100);
 
     let span = trace_span!("bootstrap");
     let joiner = Joiner::new(node, outgoing_msgs_sender, incoming_msgs, network_contacts);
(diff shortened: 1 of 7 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -115,39 +115,39 @@ impl<'a> Joiner<'a> {
...
-        .map_err(|_| Error::JoinTimeout)?
+        }
     }
 
     #[tracing::instrument(skip(self))]
     async fn join(
         mut self,
-        target_section_key: BlsPublicKey,
-        recipients: Vec<Peer>,
+        bootstrap_addr: SocketAddr,
```
joshuef, line 144: shall we use the term `join_addr` over bootstrap here? We have `bootstrap` elsewhere in the code for setting up the initial network comms. So i feel like _one_ of those needs renamed....

davidrusu: I traced through this logic and found out that `bootstrap_addr` was dead code. I've pushed up a commit to remove it entirely.

grumbach (approved):
LGTM

davidrusu:
bors cancel

sn_node/src/node/bootstrap/join.rs
```diff
@@ -294,9 +270,6 @@ impl<'a> Joiner<'a> {
 
                         info!("Setting Node name to {} (age {})", new_name, expected_age);
                         self.node = NodeInfo::new(new_keypair, self.node.addr);
-                    } else if !is_new_sap {
```
maqi, line 297: will this cause the joining node keeps switching to the new name on every retry responses it received, but those responses are for the some join request?

davidrusu: :thinking: I hadn't considered that, thanks for catching that. I'll revert this change from this PR.

PR #1610, opened 2022-09-22
feat: use semaphore to limit processing cmds

b-zee:


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -69,16 +70,22 @@ impl FlowCtrl {
     }
 
     /// Process the next pending cmds
```
joshuef, line 72: Lets add a bit of documentation around the semaphore and limit, and it's intention?

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -139,6 +146,8 @@ impl FlowCtrl {
         let mut last_link_cleanup = Instant::now();
         let mut last_dysfunction_check = Instant::now();
 
+        let semaphore = Arc::new(Semaphore::new(5));
```
joshuef, line 149: lets make this value a `const` and get some docs around it/why it was chosen. (eventually we can make it configurable, but for now I think it's fine)

joshuef:
Closing as this is likely not our solution

PR #1613, opened 2022-09-26
feat: spawn the wait/update for session in comms:is_sent

joshuef:


sn_node/src/comm/mod.rs
```diff
@@ -156,55 +155,62 @@ impl Comm {
...
-                    Err(error) => match error {
-                        Error::PeerLinkDropped => {
-                            // remove the peer link
-                            let perhaps_session = self.sessions.remove(&peer);
-                            if let Some((_peer, session)) = perhaps_session {
-                                session.disconnect().await;
+        let sessions = self.sessions.clone();
+
+        // TODO: we could cache the handles above and check them as part of loop...
+        let _handle = tokio::spawn(async move {
```
maqi, line 161: just wondering, as there is no calling of `_handle.await` to allow a waiting on this spawned task, hence the returns on the error branches, such as `return Err(Error::PeerLinkDropped);` or `Err(Error::FailedSend(peer))` actually has no effect on reporting back the errors during send?

sn_node/src/comm/peer_session.rs
```diff
@@ -43,8 +43,7 @@ impl PeerSession {
     pub(crate) fn new(link: Link) -> PeerSession {
         let (sender, receiver) = mpsc::channel(1000);
 
-        let _ =
-            tokio::task::spawn_local(PeerSessionWorker::new(link, sender.clone()).run(receiver));
+        let _ = tokio::task::spawn(PeerSessionWorker::new(link, sender.clone()).run(receiver));
```
maqi, line 46: just wondering, shall we check all the places using `spawn_local` to confirm it's really being used properly?
I feel that for the most cases, we shall use `spawn` as default ?

sn_node/src/comm/mod.rs
joshuef: Yup, this is the loop we need to close but without blocking in the send channel. Right now we can't send any other message until we have something to report. Instead I think we need a channel to feed back reports to the node perhaps?

joshuef:
Closing as #1614 and #1615 are both ahead of this

PR #1616, opened 2022-09-26
feat: client retry spend on unknown section key

jacderida:
**Important note**: the first three refactoring commits were to support the automated test, which needed setup code moved to a common location. After deciding the test was too much hassle, strictly speaking we don't need this code moved around, but it will probably come in handy in the future, so I think it's OK to leave it in.

- f7d41e512 **refactor: get public commitments from sn_dbc**

  The code for retrieving and validating the public commitments was moved out to the sn_dbc crate.

  It's needed for both the spend request and test setup code which is going to be referenced in both
  `sn_node` and `sn_client`.

  Also fixed a clippy error in `SectionTree::get_signed_by_key`.

- e776e0a45 **refactor: move build_spent_proof_share to sn_interface**

  We move this function from `sn_node` for the same reason we moved `get_public_commitments`.

  The location for it is not perfect, but it may as well sit alongside the section keys provider, as
  the function uses that to generate the share.

- 8e77ca39e **refactor: move test_utils module**

  Previously this module was located specifically under the `section_authority_provider` module, but
  now it moves to its own module and various DBC-related testing utilities are moved from `sn_node`
  to this location. Again, this test setup is needed by both `sn_node` and `sn_client`.

- 0aedf4f35 **feat: client retry spend on unknown section key**

  When the client receives an unknown section key error, it will obtain the proof chain and SAP for
  the unknown section and resubmit the request with this additional information.

  There is no automated client test for this scenario. We went to great lengths to try, but it proved
  not worth the effort. It was too difficult to modify the network knowledge with a fake section and
  still have things function correctly. The scenario is unit tested on the node side, and we done
  enough testing to know that the retry loop does what's intended.

  There are a few misc changes along with this commit:

  * Debugging message to indicate a spend request being processed correctly, which proved useful when
    trying to get the automated test working.
  * Remove the current section key from the unknown section key error. It's not necessary to include
    this.
  * When running the baby fleming network with the Makefile, include log messages from `sn_interface`.


sn_client/src/api/spentbook_apis.rs
```diff
@@ -42,14 +45,63 @@ impl Client {
...
+                            .map_err(|_| Error::SectionsDagKeyNotFound(unknown_section_key))?;
+                        let signed_sap = network
+                            .get_signed_by_key(&unknown_section_key)
+                            .ok_or(Error::SignedSapNotFound(unknown_section_key))?;
+                        cmd = SpentbookCmd::Spend {
+                            key_image,
+                            tx: tx.clone(),
+                            spent_proofs: spent_proofs.clone(),
+                            spent_transactions: spent_transactions.clone(),
+                            network_knowledge: Some((proof_chain, signed_sap.clone())),
```
davidrusu, line 95: Is it necessary to include the `signed_sap` in this network_knowledge update? Is it not enough to just include the proof-chain?

sn_interface/src/network_knowledge/test_utils.rs
```diff
@@ -0,0 +1,283 @@
...
+    let nodes = gen_sorted_nodes(&prefix, elder_count + adult_count, false);
+    let elders = nodes.iter().map(NodeInfo::peer).take(elder_count);
+    let members = nodes.iter().map(|i| NodeState::joined(i.peer(), None));
+    let section_auth =
+        SectionAuthorityProvider::new(elders, prefix, members, sk_set.public_keys(), 0);
+    (section_auth, nodes)
+}
+
+// Create signature for the given payload using the given secret key.
+pub fn prove<T: Serialize>(secret_key: &bls::SecretKey, payload: &T) -> Result<KeyedSig> {
```
davidrusu, line 114: Let's give this a more descriptive name?
```suggestion
pub fn keyed_sig<T: Serialize>(secret_key: &bls::SecretKey, payload: &T) -> Result<KeyedSig> {
```

sn_client/src/api/spentbook_apis.rs
davidrusu: We can leave this change to a future PR. (I do think  it's necessary though since we won't always have access to the SAP that corresponds to a proof-chain).

davidrusu (approved):
Looks good :+1:

sn_interface/src/network_knowledge/test_utils.rs
davidrusu: Just realized this function was just being moved. We can skip the rename as well

sn_client/src/api/spentbook_apis.rs
jacderida: Thanks for the feedback.

Updating the network knowledge seems to require sending the SAP too:
https://github.com/maidsafe/safe_network/blob/ec61fcc7c2241c65e00388dfdbbde5cb9c571158/sn_node/src/node/flow_ctrl/dispatcher.rs#L194

PR #1619, opened 2022-09-27
refactor:

dirvine:
ServiceAuth -> ClientAuth
Service -> Client
NodeBlsShare -> SectionPart
BlsShareAuth -> SectionAuthPart
SystemMsg -> Node2NodeMsg
OutgoingMsg::System -> OutgoingMsg::Node2Node
+ fmt / fix

Looking for comment, especially Service -> Client

It feels like client actions are what this describes, but using service allows nodes to make the same request. I feel like a node acts as a client when doing client things (like DBC spends?) Service kinda sounds ok but I keep confusing that with elder/system messages. 

SystemMsg had a comment saying it was only Node2Node messages, so it seems like a better name as it's easier to reason. 

BlsShareAuth felt like it tied us to BLS and also it's unclear to the reader what bls is etc. So I went with SectionAuth as that is what we are trying to achieve.


sn_interface/src/messaging/authority.rs
```diff
@@ -62,7 +62,7 @@ impl NodeAuth {
 
 /// Authority of a single peer that uses it's BLS Keyshare to sign the message.
 #[derive(Clone, Debug, Eq, PartialEq, serde::Deserialize, serde::Serialize)]
-pub struct BlsShareAuth {
+pub struct SectionAuthPart {
```
davidrusu, line 65: We've been using to the `Share` terminology throughout other modules. I think it's descriptive enough and it's used in the papers as well (i.e. threshold crypto work always refers to "shares").

```suggestion
pub struct SectionAuthShare {
```

sn_interface/src/messaging/system/mod.rs
```diff
@@ -57,7 +57,7 @@ pub enum AntiEntropyKind {
 #[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
 #[allow(clippy::large_enum_variant, clippy::derive_partial_eq_without_eq)]
 /// Message sent over the among nodes
-pub enum SystemMsg {
+pub enum Node2NodeMsg {
```
davidrusu, line 60: I'd say it's clear enough as `NodeMsg`.

```suggestion
pub enum NodeMsg {
```

sn_node/src/node/messaging/mod.rs
```diff
@@ -36,9 +37,9 @@ use std::collections::BTreeSet;
 #[derive(Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum OutgoingMsg {
-    System(SystemMsg),
-    Service(ServiceMsg),
-    DstAggregated((BlsShareAuth, Bytes)),
+    Node2Node(Node2NodeMsg),
+    Client(ClientMsg),
+    Consensus((SectionAuthPart, Bytes)),
```
davidrusu, line 42: I think this will just confuse everyone if we call this elder signature aggregation "consensus" when we have other modules running "consensus protocols".

Can we call it `SectionAuth`?
```suggestion
    SectionAuth((SectionAuthPart, Bytes)),
```

sn_node/src/node/messaging/mod.rs
```diff
@@ -36,9 +37,9 @@ use std::collections::BTreeSet;
 #[derive(Debug, Clone)]
 #[allow(clippy::large_enum_variant)]
 pub(crate) enum OutgoingMsg {
-    System(SystemMsg),
-    Service(ServiceMsg),
-    DstAggregated((BlsShareAuth, Bytes)),
+    Node2Node(Node2NodeMsg),
```
davidrusu, line 40: I'd reduce this to `Node` as well.
```suggestion
    Node(Node2NodeMsg),
```

sn_interface/src/messaging/auth_kind.rs
```diff
@@ -35,17 +35,17 @@ pub enum AuthKind {
     ///
     /// Section share authority is needed for messages related to section administration, such as
     /// DKG and relocation.
-    NodeBlsShare(BlsShareAuth),
+    SectionPart(SectionAuthPart),
```
davidrusu, line 38: I'd keep the `Share` name here too.
```suggestion
    SectionShare(SectionAuthPart),
```

davidrusu (commented):
I like the changes! my suggestions would be:

1. rather than use `Part`, let's keep using `Share`, it's used by the wider community and is clear enough what it means.
2. reduce `Node2Node` to `Node`.

But yeah, I like the `service -> client`, `system -> node` renames.

dirvine:
Thanks @davidrusu I will push those suggestions now.

sn_node/src/node/bootstrap/relocate.rs
```diff
@@ -301,7 +301,7 @@ mod tests {
                 }
                 if !matches!(
                     msg,
-                    OutgoingMsg::System(SystemMsg::JoinAsRelocatedRequest(_))
+                    OutgoingMsg::Node2Node(NodeMsg::JoinAsRelocatedRequest(_))
```
davidrusu, line 304: Looks like you may have missed the tests. Try verifying with `cargo clippy --all-targets`.
```suggestion
                    OutgoingMsg::Node(NodeMsg::JoinAsRelocatedRequest(_))
```

sn_node/src/node/messaging/anti_entropy.rs
```diff
@@ -592,11 +592,11 @@ mod tests {
                     &sender,
                 );
 
-            let msg = assert_matches!(cmd, Ok(Some(Cmd::SendMsg { msg: OutgoingMsg::System(msg), .. })) => {
+            let msg = assert_matches!(cmd, Ok(Some(Cmd::SendMsg { msg: OutgoingMsg::Node2Node(msg), .. })) => {
```
davidrusu, line 595: same here
```suggestion
            let msg = assert_matches!(cmd, Ok(Some(Cmd::SendMsg { msg: OutgoingMsg::Node(msg), .. })) => {
```

davidrusu (approved):
:100:

sn_node/CHANGELOG.md
```diff
@@ -4797,12 +4797,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Bug Fixes
 
- - <csr-id-0e505be2f57ab427cd3ed8c9564fd8b84909f6f3/> restore `ServiceMsg` authority check
+ - <csr-id-0e505be2f57ab427cd3ed8c9564fd8b84909f6f3/> restore `ClientMsg` authority check
```
grumbach, line 4800: Do we want to do the replacement in the Changelog file as well?

grumbach (approved):
LGTM

sn_node/CHANGELOG.md
dirvine: Sorry I missed that. I recon it's probably OK mind you. We have bigger fish to fry I reckon

PR #1620, opened 2022-09-28
fix(sn_node): restart after accumulated retry_response

maqi:

### DIFF
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -19,7 +19,7 @@ use sn_interface::{
         },
         AuthKind, Dst, MsgType, NodeAuth, WireMsg,
     },
-    network_knowledge::{NetworkKnowledge, NodeInfo, SectionTree, MIN_ADULT_AGE},
+    network_knowledge::{NetworkKnowledge, NodeInfo, SectionTree},
     types::{keys::ed25519, log_markers::LogMarker, Peer},
 };
 
@@ -27,6 +27,7 @@ use backoff::{backoff::Backoff, ExponentialBackoff};
 use bls::PublicKey as BlsPublicKey;
 use futures::future;
 use resource_proof::ResourceProof as ChallengeSolver;
+use std::collections::{BTreeMap, BTreeSet};
 use tokio::{
     sync::mpsc,
     time::{Duration, Instant},
@@ -70,6 +71,7 @@ struct Joiner<'a> {
     network_contacts: SectionTree,
     backoff: ExponentialBackoff,
     aggregated: bool,
+    retry_responses_cache: BTreeMap<Peer, u8>,
 }
 
 impl<'a> Joiner<'a> {
@@ -97,6 +99,7 @@ impl<'a> Joiner<'a> {
             network_contacts,
             backoff,
             aggregated: false,
+            retry_responses_cache: Default::default(),
         }
     }
 
@@ -234,7 +237,7 @@ impl<'a> Joiner<'a> {
                     }
 
                     // make sure we received a valid and trusted new SAP
-                    let is_new_sap = match self.network_contacts.update(section_tree_update) {
+                    let _is_new_sap = match self.network_contacts.update(section_tree_update) {
                         Ok(updated) => updated,
                         Err(err) => {
                             debug!("Ignoring JoinResponse::Retry with an invalid SAP: {err:?}");
@@ -242,10 +245,8 @@ impl<'a> Joiner<'a> {
                         }
                     };
 
-                    // if it's not a new SAP, ignore response unless the expected age is different.
-                    if self.node.age() != expected_age
-                        && (self.node.age() > expected_age || self.node.age() <= MIN_ADULT_AGE)
-                        && !self.aggregated
+                    if let Some(expected_age) =
+                        self.insert_retry_response(sender, expected_age, signed_sap.elders_set())
                     {
                         // adjust our joining age to the expected by the network
                         trace!(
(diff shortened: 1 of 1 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +372,43 @@ impl<'a> Joiner<'a> {
...
+    // The retry responses may containing different expected ages, or only having
+    // more than 1/3 of elders copies (which blocks the aggregation).
+    // So a restart with new name shall happen when: received 1/3 of elders copies
+    // And with an age of: the lowest expected age
+    fn insert_retry_response(
+        &mut self,
+        sender: Peer,
+        expected_age: u8,
+        elders: BTreeSet<Peer>,
+        elders_count: u8,
```
davidrusu, line 384: Why do we need to pass this `elders_count` in? Can't we calculate it from `elders.len()`?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +372,43 @@ impl<'a> Joiner<'a> {
...
+    ) -> Option<u8> {
+        if !elders.contains(&sender) {
+            error!(
+                "Sender {:?} of the retry-response is not part of the elders {:?}",
+                sender, elders
+            );
```
davidrusu, line 390: We can save a lot of vertical space if you move variables into the format strings:
```suggestion
            error!("Sender {sender:?} of the retry-response is not part of the elders {elders:?}");
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +372,43 @@ impl<'a> Joiner<'a> {
...
+        let mut responded = 0;
+        let mut min_expected_age = 255;
+        for (elder, expected_age) in self.retry_responses_cache.iter() {
+            if elders.contains(elder) {
+                responded += 1;
+                if min_expected_age > *expected_age {
+                    min_expected_age = *expected_age;
+                }
+            }
+        }
```
davidrusu, line 403: This logic does not seem to not check for elder agreement on _an_ expected age.

What if we change the `retry_response_cache` to be a map `BTreeMap<expected_age, BTreeSet<sender>>`.

Then to see how many elders responded with retry for a particular age, we can just check the size of the set.

```suggestion
        // log that this sender asked us to retry for this age.
        self.retry_responses_cache.entry(expected_age).or_default().insert(sender);
        
        let min_expected_age = self.retry_responses_cache
            .iter()
            .filter(|(_, responders)| responders.len() > elder_count / 3)
            .map(|(age, _)| age)
            .min()?
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +372,43 @@ impl<'a> Joiner<'a> {
...
+
+        if responded > elders_count / 3 && self.node.age() != expected_age && !self.aggregated {
+            Some(min_expected_age)
+        } else {
+            None
+        }
```
davidrusu, line 409: Should the node age check be against the `min_expected_age`?
```suggestion
        if responded > elders_count / 3 && self.node.age() != min_expected_age && !self.aggregated {
            Some(min_expected_age)
        } else {
            None
        }
```

sn_node/src/node/bootstrap/join.rs
maqi: I will change the result age to be `the expected_age get voted by the most valid elders`.
However, using the struct of `BTreeMap<expected_age, BTreeSet<sender>>` might have other issue of `senders` could be from different `SAP`.

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +369,51 @@ impl<'a> Joiner<'a> {
...
+            error!("Sender {sender:?} of the retry-response is not part of the elders {elders:?}");
+            return None;
+        }
+        let _ = self.retry_responses_cache.insert(sender, expected_age);
+        let mut responded = 0;
+        let mut expected_ages: BTreeMap<u8, u8> = Default::default();
+        for (elder, expected_age) in self.retry_responses_cache.iter() {
+            if elders.contains(elder) {
+                responded += 1;
+                *expected_ages.entry(*expected_age).or_insert(1) += 1;
```
davidrusu, line 392: You are double counting the first entry
```suggestion
                *expected_ages.entry(*expected_age).or_insert(0) += 1;
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +369,51 @@ impl<'a> Joiner<'a> {
...
+            }
+        }
```
davidrusu, line 394: You are still counting responses across multiples ages. (`responded` get's incremented for each elder, even if they voted differently).

sn_node/src/node/bootstrap/join.rs
```diff
@@ -375,6 +369,51 @@ impl<'a> Joiner<'a> {
...
+        if responded > elders.len() / 3 && !self.aggregated {
+            let mut voted_expected_age = 255;
+            let mut highest_count = 0;
+            for (expected_age, count) in expected_ages {
+                // In case two ages got the same votes, we prefer the lower age
+                if highest_count < count {
+                    highest_count = count;
+                    voted_expected_age = expected_age;
+                }
+            }
```
davidrusu, line 405: That comment doesn't seem to match what the code is doing? Are you relying on the btreeset ordering the elements by age?
Can we rewrite this more clearly where we show how the ties are resolved explicitly?
```suggestion
            // In case two ages got the same votes, we prefer the lower age
            let (voted_expected_age, count) = expected_ages
                .iter()
                .max_by_key(|(age, count)| (count, -(age as i32)))?;
```

sn_node/src/node/bootstrap/join.rs
maqi: Yes, and that's because when elders think the joining node shall be retry with a new age, they could give different expected age based on their network knowledge, which could be a split view that no age get a majority votes.
So, basically the responded counter is to ensure the retry action is really needed, then choose an expected age (with highest votes) that having the highest chance to be the approved age.

sn_node/src/node/bootstrap/join.rs
davidrusu: Ok, I can understand the reluctance to filter here, but any reason why you don't use the iterators and the `max_by_key()` function? Then you won't need to add a comment to explain the BTreeMap implementation detail

sn_node/src/node/bootstrap/join.rs
davidrusu: Ok, makes sense :+1:

PR #1622, opened 2022-09-29
feat: use separate genesis dbc and section keys

jacderida:
For more information, see the comment on the diff.

The generation of the DBC was also updated to use non-zero values for the `alpha` and `r` fields of `MlsagMaterial`, because zeroes and ones are special values in elliptic curve cryptography.

Also fixed a Clippy warning.


bochaco (approved):

davidrusu (approved):
Looks good, mind just changing the default values in the MlsagMateraial? just want to avoid using 0's or 1's for the `alpha` and `r`

davidrusu:
bors retry

PR #1628, opened 2022-10-05
refactor: NodeAuth -> NodeEvidence

dirvine:
Nodes signs messages they send.

This signature is irrefutable proof of the nodes actions & is used to penalise bad behaviour.

BREAKING CHANGE:


grumbach (approved):
Looks amazing yeah!

PR #1631, opened 2022-10-05
increase restart threshold during join

maqi:


maqi:
work got merged into main branch already with PR 1636

PR #1632, opened 2022-10-05
refactor(messaging)!: removing dst_address fn from ClientMsg as it doesn't always contain that info

bochaco:
Also removing `ClientMsg:ClientError` type which is not needed/used.


grumbach (approved):
Looking good!

PR #1636, opened 2022-10-06
Accumulate retry

maqi:
This PR contains two main work:
1, liftup the joining node restart threshold, as with the feature branch with slower communication, the previous threshold is too sensetive that will restart BEFORE receive the approval. 

2, an attempt to expand range of changed code file that will trigger the fmt/clippy check.

### DIFF
--- a/.github/workflows/sn_pr_tests.yml
+++ b/.github/workflows/sn_pr_tests.yml
@@ -84,19 +84,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - uses: dorny/paths-filter@v2
-        id: sn_changes
-        with:
-          filters: |
-            src:
-              - 'sn_api/src/**'
-              - 'sn_cli/src/**'
-              - 'sn_client/src/**'
-              - 'sn_interface/src/**'
-              - 'sn_node/src/**'
-
       - uses: actions-rs/toolchain@v1
-        if: steps.sn_changes.outputs.src == 'true'
         id: toolchain
         with:
           profile: minimal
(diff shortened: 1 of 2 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -358,22 +358,23 @@ impl<'a> Joiner<'a> {
...
+        if !self.aggregated {
+            let threshold = 2 * elders.len() / 3;
+            let voted_expected_age = expected_ages
                 .iter()
-                // In case two ages got the same votes, we prefer the lower age
-                .max_by_key(|(age, count)| (*count, -(**age as i32)))?;
-            if self.node.age() != *voted_expected_age {
-                return Some(*voted_expected_age);
+                .find(|(_age, count)| **count >= threshold as u8)
+                .map(|(age, _count)| *age)?;
```
davidrusu, line 375: Don't you want to prefer the smaller age if you have multiple ages that have passed the threshold? i.e.
```suggestion
                .filter(|(_age, count)| **count >= threshold as u8)
                // In case two ages got the same votes, we prefer the lower age
                .max_by_key(|(age, count)| (*count, -(**age as i32)))?;
```

.github/workflows/sn_pr_tests.yml
```diff
@@ -90,10 +90,17 @@ jobs:
...
-              - 'sn_client/src/**'
+              - 'sn_cli/tests/**'
+              - 'sn_client/**'
+              - 'sn_cmd_test_utilities/**'
+              - 'sn_dysfunction/**'
               - 'sn_interface/src/**'
               - 'sn_node/src/**'
+              - 'sn_node/benches/**'
+              - 'sn_node/examples/**'
+              - 'testnet/**'
```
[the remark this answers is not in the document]
jacderida: Yeah, I would be for that. I think this was originally used for another purpose than controlling just Clippy. Maybe for running particular tests or something. That doesn't seem to apply any more, so let's just get rid of it.

sn_node/src/node/bootstrap/join.rs
maqi: the threshold is 2/3 of the elders.len(), and the votes shall all from the same elders, which means there shall only be one `age` got more than 2/3 elders' votes.
so no need to have the `max_by_key` anymore

davidrusu: we talked through this, the point is that we always track the most-recent votes from each elder, so it's impossible to get multiple super-majorities here.

davidrusu (approved):

PR #1642, opened 2022-10-09
refactor: Remove resource proof

dirvine:
We use dysfunctional for this check


sn_interface/src/messaging/system/join.rs
```diff
@@ -8,59 +8,32 @@
...
-use bls::PublicKey as BlsPublicKey;
-use ed25519_dalek::Signature;
 use serde::{Deserialize, Serialize};
 use sn_consensus::Decision;
-use std::{collections::VecDeque, net::SocketAddr};
+use std::net::SocketAddr;
 
 /// Response to a request to join a section
 #[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 #[allow(clippy::large_enum_variant)]
```
b-zee, line 17: This line should be removed.

PR #1644, opened 2022-10-10
chore(node): new Cargo feature for disabling dysfunctional nodes checks

bochaco:
This is also using the new feature disabling dysfunctional nodes checks for CI testing.

### DIFF
--- a/sn_node/src/node/flow_ctrl/periodic_checks.rs
+++ b/sn_node/src/node/flow_ctrl/periodic_checks.rs
@@ -6,8 +6,6 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::FlowCtrl;
-
 use crate::node::{flow_ctrl::cmds::Cmd, messaging::Peers, MyNode, Result};
 
 use ed25519_dalek::Signer;
@@ -23,7 +21,7 @@ use sn_interface::{
     types::{ChunkAddress, PublicKey, Signature},
 };
 
-use std::{collections::BTreeSet, sync::Arc, time::Duration};
+use std::{sync::Arc, time::Duration};
 use tokio::{sync::RwLock, time::Instant};
 
 const PROBE_INTERVAL: Duration = Duration::from_secs(30);
@@ -38,7 +36,8 @@ const DYSFUNCTION_CHECK_INTERVAL: Duration = Duration::from_secs(5);
 const ADULT_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(2);
 const ELDER_HEALTH_CHECK_INTERVAL: Duration = Duration::from_secs(3);
 
-pub(super) struct PeriodicChecksTimestamps {
+pub(super) struct PeriodicChecks {
+    node: Arc<RwLock<MyNode>>,
     last_probe: Instant,
     last_section_probe: Instant,
     last_adult_health_check: Instant,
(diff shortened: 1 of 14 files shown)


sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -64,130 +64,115 @@ impl PeriodicChecksTimestamps {
...
 
-        self.enqueue_cmds_for_elder_periodic_checks().await;
+        cmds.extend(self.enqueue_cmds_for_elder_periodic_checks().await);
+
+        cmds
     }
 
     /// Periodic tasks run for elders and adults alike
-    async fn enqueue_cmds_for_standard_periodic_checks(&mut self) {
+    async fn enqueue_cmds_for_standard_periodic_checks(&mut self) -> Vec<Cmd> {
```
davidrusu, line 86: rename since we no longer enqueue
```suggestion
    async fn standard_periodic_checks(&mut self) -> Vec<Cmd> {
```

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -64,130 +64,115 @@ impl PeriodicChecksTimestamps {
...
-        for cmd in cmds {
-            // dont use sender here incase channel gets full
-            self.fire_and_forget(cmd, None).await;
-        }
+        cmds
     }
 
     /// Periodic tasks run for elders only
-    async fn enqueue_cmds_for_elder_periodic_checks(&mut self) {
+    async fn enqueue_cmds_for_elder_periodic_checks(&mut self) -> Vec<Cmd> {
```
davidrusu, line 127:
```suggestion
    async fn elder_periodic_checks(&mut self) -> Vec<Cmd> {
```

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -64,130 +64,115 @@ impl PeriodicChecksTimestamps {
...
-        for cmd in cmds {
-            // dont use sender here incase channel gets full
-            self.fire_and_forget(cmd, None).await;
-        }
+        cmds
     }
 
     /// Periodic tasks run for adults only
-    async fn enqueue_cmds_for_adult_periodic_checks(&mut self) {
+    async fn enqueue_cmds_for_adult_periodic_checks(&mut self) -> Vec<Cmd> {
```
davidrusu, line 114:
```suggestion
    async fn adult_periodic_checks(&mut self) -> Vec<Cmd> {
```

davidrusu (commented):
This seems like a "disable periodic checks" change, which I don't think will do what you want.

The periodic checks are also used get us resilience to dropped messages, the gossip functions that are marked with `#[allow(dead_code)]` will no longer be re-sending messages when we see that we have not made progress after some timeout.

An alternative, how about we just disable the `vote dysfunctional nodes offline` logic? That is, dysfunction continues to work as it does right now, periodic checks are still sent out, and dysfunctional nodes are tracked except that we ignore any recommendations to remove dysfunctional nodes.

bochaco:
The main objective of this was to remove all type of msgs that are not the ones directly/undirectly related to a query/cmd sent by clients, for two main reason:
1- in our CI we are not yet testing churn, so there shouldn't be any dysfunctional checks triggered or periodic checks needed, every query/cmd from client should function correctly without any of those features. So, we want to disable it to make sure we are not hiding bugs, a dysfunctional check related msg could be fixing a data inconsistency bug and we don't want that to happen but the bug to be exposed.
2- it's easier to debug issues when we can check that msgs required to process a client cmd/query are correct, all dysfunctional/periodic checks related msgs introduce a lot of noise useless for the type of debugging and investigation we are right now doing.

PR #1645, opened 2022-10-10
refactor: adapt confusing auth related code

grumbach:
Removes the confusion @dirvine pointed out around:

```rust
pub fn verify_src_section_key_is_known(&self, known_keys: &BTreeSet<bls::PublicKey>) -> bool {
         let section_pk = match &self {
             // NB TODO this shouldnt be true! Remove all this
             Self::Node(_) => return true,
```


dirvine (approved):
LGTM Purrfect :-)

PR #1648, opened 2022-10-11
Sign Registers and Chunks with Section Key

jacderida:
- fa8c617e9 **feat: sign replicated chunks with section key**

  BREAKING CHANGE: the `ReplicatedData::Chunk` enum variant now uses a new `SignedChunk` type and
  several APIs were updated to use this type rather than `Chunk`.

  A new `SignedChunk` type is introduced. It wraps a `Chunk` and provides an `authority` field, which
  is a `SectionSig`. When a chunk is replicated, it is signed using the current section key and the
  signed chunk is then stored. The local node storage was updated to store this new type. Previously,
  only the bytes of the chunk were being stored and not a serialised representation of the type, but
  now the serialised representation is being stored so that the signature is also persisted.

- 36feda252 **refactor: rename authority field to owner**

  This was something we discussed in a team meeting.

  Since a new `SignedRegister` type will be introduced which will have an `authority` field with a
  section signature, it was more appropriate for the current `authority` field to be renamed to `owner`.

  The `SignedChunk` type also uses `authority`, so that keeps it consistent with that.

- e7590ab43 **feat: sign registers with section key**

  The register creation command already had a `section_sig` field on it, but this was being populated
  with a dummy value. It has now been changed to an `Option` field. It is assigned `None` when the
  register creation command is created on the client, then when it's received by the node, the field
  is populated with the section key.

  BREAKING CHANGE: in similar fashion to the `SignedChunk` type, a new `SignedRegister` type is
  introduced. Register storage now uses this new type and queries also return it.

### DIFF
--- a/sn_node/src/node/messaging/client_msgs.rs
+++ b/sn_node/src/node/messaging/client_msgs.rs
@@ -31,7 +31,8 @@ use sn_interface::{
     types::{
         log_markers::LogMarker,
         register::{Permissions, Policy, Register, User},
-        Keypair, Peer, PublicKey, RegisterCmd, ReplicatedData, SPENTBOOK_TYPE_TAG,
+        Keypair, Peer, PublicKey, RegisterCmd, ReplicatedData, SectionSig, SignedChunk,
+        SPENTBOOK_TYPE_TAG,
     },
 };
 
@@ -249,7 +250,20 @@ impl MyNode {
                 let reg_cmd = self.gen_register_cmd(&key_image, &spent_proof_share)?;
                 ReplicatedData::SpentbookWrite(reg_cmd)
             }
-            ClientMsg::Cmd(DataCmd::StoreChunk(chunk)) => ReplicatedData::Chunk(chunk),
+            ClientMsg::Cmd(DataCmd::StoreChunk(chunk)) => {
+                let section_key = self.network_knowledge.section_key();
+                let (_, sig_share) = self
+                    .section_keys_provider
+                    .sign_with(chunk.value(), &section_key)?;
+                let section_sig = SectionSig {
+                    public_key: section_key,
+                    signature: sig_share.0,
+                };
+                ReplicatedData::Chunk(SignedChunk {
+                    chunk,
+                    authority: section_sig,
+                })
+            }
             ClientMsg::Query(query) => {
                 return self
                     .read_data_from_adults(
--- a/sn_node/src/node/messaging/client_msgs.rs
+++ b/sn_node/src/node/messaging/client_msgs.rs
@@ -206,7 +206,18 @@ impl MyNode {
         // extract the data from the request
         let data = match msg {
             // These reads/writes are for adult nodes...
-            ClientMsg::Cmd(DataCmd::Register(cmd)) => ReplicatedData::RegisterWrite(cmd),
+            ClientMsg::Cmd(DataCmd::Register(cmd)) => match cmd {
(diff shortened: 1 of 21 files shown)


sn_node/src/node/messaging/client_msgs.rs
```diff
@@ -430,4 +448,13 @@ impl MyNode {
...
         Ok(RegisterCmd::Edit(signed_edit))
     }
+
+    fn section_sig(&self, data: &[u8]) -> Result<SectionSig> {
+        let section_key = self.network_knowledge.section_key();
+        let (_, sig_share) = self.section_keys_provider.sign_with(data, &section_key)?;
+        Ok(SectionSig {
+            public_key: section_key,
+            signature: sig_share.0,
+        })
```
grumbach, line 458: Why do we create a new signature here? 
And shouldn't a `SectionSig` be the result of super majority of section `sig_share` aggregated together instead of just one share?

sn_node/src/storage/registers.rs
```diff
@@ -427,24 +441,26 @@ fn create_reg_w_policy(
...
+        section_sig: section_sig()?,
     })
 }
 
-fn section_sig() -> SectionSig {
-    let sk = bls::SecretKey::random();
+fn section_sig() -> Result<Option<SectionSig>> {
+    let sk = bls::SecretKey::from_hex(
+        "01275e09359849ec8bada62eaf7319bd1558663e25e3accaae9f0fd2e5748964",
+    )?;
```
grumbach, line 451: Why is this not in the test module? I'm feeling it's dangerous to have it lay around, if someone calls it accidentally.

sn_node/src/node/messaging/client_msgs.rs
jacderida: Is the signature not based on the data of the particular register/chunk? Did I misunderstand that?

```
/// Signature created when a quorum of the section elders has agreed on something.
#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize)]
pub struct SectionSig {
    /// The BLS public key.
    pub public_key: bls::PublicKey,
    /// The BLS signature corresponding to the public key.
    pub signature: bls::Signature,
}
```

So should the type of that `signature` field be a `SignatureShare` then?

sn_node/src/storage/registers.rs
jacderida: Yeah, I'll make this test only. Thanks.

sn_node/src/node/messaging/client_msgs.rs
grumbach: Maybe I'm missing something but the way I see the flow is:
- client asks elders to store chunk
- elders send their signature share
- client aggregate and create the section signature
- client sends that signature along with the data to ANY network elder
- elder stores it as long as the data is signed with a trusted section key
- when there this chunk is copied over to other nodes, elders send the data along with the original signature

joshuef: Yeh, that's the flow I think @grumbach. We'll need a flow that mimics the payment flow (or perhaps to dive right in there). 

In order to get this in, I think we'll need to sign the data xorname at Elders? (and later that can be switched to be the sig over the data payment, which itself will contain the data xorname?).

jacderida:
I'm going to close this and just start on another branch. It's way out of date now and it was completely wrong anyway.

PR #1651, opened 2022-10-11
Bootstrap section tree before starting join process

davidrusu:

### DIFF
--- a/sn_interface/src/network_knowledge/sections_dag.rs
+++ b/sn_interface/src/network_knowledge/sections_dag.rs
@@ -563,8 +563,6 @@ pub(crate) mod tests {
 
         assert_lists(dag.keys(), [pk_gen, info_a1.key, info_a2.key, info_b.key]);
 
-        // cannot get partial dag till genesis
-        assert!(dag.partial_dag(&pk_gen, &pk_gen).is_err());
         assert_lists(
             dag.partial_dag(&pk_gen, &info_a2.key)?.keys(),
             [pk_gen, info_a1.key, info_a2.key],
--- a/sn_interface/src/network_knowledge/sections_dag.rs
+++ b/sn_interface/src/network_knowledge/sections_dag.rs
@@ -205,6 +205,17 @@ impl SectionsDAG {
     pub fn partial_dag(&self, from: &bls::PublicKey, to: &bls::PublicKey) -> Result<Self> {
         // start from the "to" key (bottom of the tree) and traverse to the root
         let mut crdt_ops: Vec<Node<SectionInfo>> = Vec::new();
+
+        // Return the singleton DAG when from == to == genesis
+        // This is a special case because the genesis entry does not have a `Node` so `get_node(to)` will fail
+        if to == &self.genesis_key {
+            if from != &self.genesis_key {
+                return Err(Error::InvalidBranch);
+            } else {
+                return Ok(Self::new(self.genesis_key));
+            }
+        }
+
         // if "to" key is genesis, returns error
         let mut node = self.get_node(to)?;
         loop {
@@ -501,6 +512,18 @@ pub(crate) mod tests {
     use rand::{rngs::StdRng, thread_rng, Rng, RngCore, SeedableRng};
     use std::collections::{BTreeMap, BTreeSet};
     use xor_name::Prefix;
+
+    #[test]
+    fn test_partial_dag_for_genesis_key() {
+        let (_, pk) = gen_keypair();
+        let dag = SectionsDAG::new(pk);
+
+        let partial_dag = dag
+            .partial_dag(&pk, &pk)
+            .expect("Should have succeeded in creating a partial dag");
+
+        assert_eq!(Vec::from_iter(partial_dag.keys()), vec![pk]);
+    }
 
     #[test]
     fn insert_last() -> Result<()> {
--- a/sn_interface/src/network_knowledge/sections_dag.rs
+++ b/sn_interface/src/network_knowledge/sections_dag.rs
@@ -523,6 +523,13 @@ pub(crate) mod tests {
             .expect("Should have succeeded in creating a partial dag");
 
         assert_eq!(Vec::from_iter(partial_dag.keys()), vec![pk]);
+
+        let (_, pk_other) = gen_keypair();
+
+        assert!(matches!(
+            dag.partial_dag(&pk_other, &pk),
+            Err(Error::InvalidBranch)
+        ));
     }
 
     #[test]
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -542,7 +542,7 @@ mod tests {
         let state = Joiner::new(node.clone(), send_tx, &mut recv_rx, tree);
 
         // Create the bootstrap task, but don't run it yet.
-        let bootstrap = async move { state.try_join(join_timeout).await.map_err(Error::from) };
+        let bootstrap = async { state.try_join(join_timeout).await.map_err(Error::from) };
 
         let (next_sap, next_sk_set, next_elders, _) =
             TestSapBuilder::new(Prefix::default()).build();
@@ -557,37 +557,35 @@ mod tests {
         // Create the task that executes the body of the test, but don't run it either.
         let others = async {
             // Receive JoinRequest
-            let (wire_msg, recipients) = send_rx
-                .recv()
-                .await
-                .ok_or_else(|| eyre!("JoinRequest was not received"))?;
-
+            let (wire_msg, recipients) = send_rx.recv().await.expect("Expected join message");
             itertools::assert_equal(recipients, genesis_sap.elders());
+            let node_msg =
+                assert_matches!(wire_msg.into_msg(), Ok(MsgType::Node { msg, .. }) => msg);
+            assert_matches!(node_msg, NodeMsg::AntiEntropyProbe(_));
 
-            let node_msg = assert_matches!(wire_msg.into_msg(), Ok(MsgType::Node { msg, .. }) =>
-                msg);
-
+            let (wire_msg, recipients) = send_rx.recv().await.expect("Expected join message");
+            itertools::assert_equal(recipients, genesis_sap.elders());
+            let node_msg =
+                assert_matches!(wire_msg.into_msg(), Ok(MsgType::Node { msg, .. }) => msg);
             assert_matches!(node_msg, NodeMsg::JoinRequest(JoinRequest { .. }));
 
             // Send JoinResponse::Retry with new SAP
             let other_elders: Vec<&MyNodeInfo> =
                 next_elders.iter().take(2 * elder_count() / 3).collect_vec();
             for elder in other_elders.iter() {
-                send_response(
+                send_join_response(
                     &recv_tx,
                     JoinResponse::Retry {
                         section_tree_update: section_tree_update.clone(),
                     },
                     elder,
                     next_sap.section_key(),
-                )?;
+                );
             }
 
             // Receive the second JoinRequest with correct section info
-            let (wire_msg, recipients) = send_rx
-                .recv()
-                .await
-                .ok_or_else(|| eyre!("JoinRequest was not received"))?;
+            let (wire_msg, recipients) =
+                send_rx.recv().await.expect("JoinRequest was not received");
             let (node_msg, dst) = assert_matches!(wire_msg.into_msg(), Ok(MsgType::Node { msg, dst,.. }) =>
                 (msg, dst));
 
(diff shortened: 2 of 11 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -443,19 +469,42 @@ impl<'a> Joiner<'a> {
...
     #[tracing::instrument(skip(self))]
     async fn receive_join_response(&mut self) -> Result<(JoinResponse, Peer)> {
-        while let Some(MsgFromPeer { sender, wire_msg }) = self.incoming_msgs.recv().await {
-            // We are interested only in `JoinResponse` type of messages
+        loop {
+            let (msg, sender) = self.receive_node_msg().await?;
+            match msg {
+                NodeMsg::JoinResponse(resp) => return Ok((*resp, sender)),
+                _ => {
+                    trace!("Bootstrap message discarded: sender: {sender:?} msg: {msg:?}")
```
joshuef, line 477: I wonder if this log is clearer as 
```suggestion
                    trace!("Non-Bootstrap message received and discarded: sender: {sender:?} msg: {msg:?}")
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -443,19 +469,42 @@ impl<'a> Joiner<'a> {
...
+            match msg {
+                NodeMsg::AntiEntropy {
+                    section_tree_update,
                     ..
-                }) => return Ok((*resp, sender)),
-                Ok(MsgType::Client { msg_id, .. } | MsgType::Node { msg_id, .. }) => {
-                    trace!("Bootstrap message discarded: sender: {sender:?} msg_id: {msg_id:?}");
+                } => return Ok(section_tree_update),
+                _ => {
+                    trace!("Bootstrap message discarded: sender: {sender:?} msg: {msg:?}")
```
joshuef, line 493: same thing here. (and perhaps something to distinguish between the two discard cases?)

joshuef (commented):
non critical, just something that's bugged me whenever I've poked about this msg previously

sn_node/src/node/bootstrap/join.rs
```diff
@@ -443,19 +469,42 @@ impl<'a> Joiner<'a> {
...
-                Err(err) => {
-                    error!("Failed to deserialize message payload: {:?}", err);
+            }
+        }
+    }
+
+    #[tracing::instrument(skip(self))]
+    async fn receive_node_msg(&mut self) -> Result<(NodeMsg, Peer)> {
+        while let Some(MsgFromPeer { sender, wire_msg }) = self.incoming_msgs.recv().await {
+            // We are interested only in `JoinResponse` type of messages
```
maqi, line 502: only interested in `Node` type of messages?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -357,6 +360,29 @@ impl<'a> Joiner<'a> {
...
+                true,
+            )
+            .await?;
+
+            let section_tree_update =
+                tokio::time::timeout(response_timeout, self.receive_section_tree_update())
+                    .await
+                    .map_err(|_| Error::JoinTimeout)??;
+
+            if !self.section_tree.update(section_tree_update)? {
```
maqi, line 380: better put a comment here that: keeps fetching section_tree as long as there is an update being carried out

Meanwhile, this is one section_tree from one node, if the update returns with `updated` , if means we will send a probe to `All` elders again? which is abuse of messaging?

Also, if the update returns with an error, the loop will be terminated too early? i.e. we shall wait for other elders' section_tree as well?

davidrusu: I've reworked the logic here a little bit. We now wait until we receive a threshold of updates before we check for new information.

The only error that would terminate the loop early is if our connections are closed or if we timeout waiting for a section tree update.

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -205,6 +205,17 @@ impl SectionsDAG {
...
+        // Return the singleton DAG when from == to == genesis
+        // This is a special case because the genesis entry does not have a `Node` so `get_node(to)` will fail
+        if to == &self.genesis_key {
+            if from != &self.genesis_key {
+                return Err(Error::InvalidBranch);
+            } else {
+                return Ok(Self::new(self.genesis_key));
+            }
+        }
+
```
davidrusu, line 218: cc @RolandSherwin small change to `SectionsDag::partial_dag` semantics here.

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -196,10 +193,11 @@ impl MyNode {
...
                             );
 
-                            let recipients: Vec<_> =
-                                section_tree_update.signed_sap.elders().cloned().collect();
-
                             let section_tree = node.network_knowledge.section_tree().clone();
+                            let section_tree_update = node
+                                .network_knowledge
+                                .section_tree()
+                                .generate_section_tree_update(&Prefix::default())?; // TODO: remove this dummy update
```
maqi, line 200: I'd prefer this TODO to be resolved within this PR.
and using a `dummy` update could incur unexpected issue.
Maybe a quick one is just to just change the constructor to:
```rust
NetworkKnowledge::new(SectionTree, Option<SectionTreeUpdate>)
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -274,40 +229,22 @@ impl<'a> Joiner<'a> {
                         continue;
```
maqi, line 229: This may not related to this PR, however, as I understand,
the above `if !section_auth.prefix().matches(&self.node.name())` shall use the `section_auth` of the received msg, 
not the `section_auth of JoinResponse::Redirect(section_auth)` .
As the node could be redirected to a different section?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -274,40 +229,22 @@ impl<'a> Joiner<'a> {
...
                     info!(
                         "Newer JoinResponse::Redirect for us {:?}, SAP {:?} from {:?}",
                         self.node.name(),
                         section_auth,
                         sender
                     );
 
-                    section_key = new_section_key;
-                    self.prefix = section_auth.prefix();
+                    self.bootstrap_section_tree(section_auth, response_timeout)
```
maqi, line 239: just to verify:
this means for each individual redirect/retry response, we will startup a `bootstrap_section_tree` round, which means sending messages to all elders?
That will cause un-necessary messages and may even a message storming?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -274,40 +229,22 @@ impl<'a> Joiner<'a> {
...
+                        .await?;
 
-                    let msg = JoinRequest { section_key };
+                    let target_sap = self.join_target_sap()?;
```
maqi, line 242: when a new node get redirected, it actually need to re-generate it's name to match the target section?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -321,6 +258,48 @@ impl<'a> Joiner<'a> {
...
+            )
+            .await?;
+
+            // We wait till we receive a threshold of updates.
+
+            let mut any_new_information = false;
+
+            for _ in 0..target_sap.public_key_set().threshold() {
+                let update =
+                    tokio::time::timeout(response_timeout, self.receive_section_tree_update())
```
maqi, line 282: the passed in `response_timeout` is actually the overall `join_time_out` 
but here the work to be carried out is just for `query  one elder` , and we only start a new one AFTER the previous one completed or timedout. 
which means whenever we encountered the first timeout here, the overall timeout is reached.
In normal case, this only leaves us ONE chance to just query the first elder?

maybe we shall use `join_time_out / elders.len()` as per action timeout, or make the actions to be undertaken parrallellized?

further more, the `receive_section_tree_update` function itself is a `looping receiver` , 
whichmeans the `for` clause here is bit confusing/wasted or even incorrect?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -321,6 +258,48 @@ impl<'a> Joiner<'a> {
...
+                    break;
+                }
+            }
+
+            if any_new_information {
+                // Update the target sap since we've received new information and try again.
+                target_sap = self.join_target_sap()?;
+            } else {
+                // We are up to date with these nodes so we can end the bootstrap
+                return Ok(());
```
maqi, line 298: the default value of `any_new_information ` is `false`, which means if all timedout, we will still return as `OK`.
Currently, the overall timeout is the only way to bail us out. However this could be tricky/dangerous.

sn_node/src/node/bootstrap/join.rs
davidrusu: Not quite, the retry cache prevents us from re-bootstrapping until we have t + 1 retries.

The redirect logic did not have that cache in front of it so it was doing a join request for every redirect response. The flow remains the same with this refactor, the difference is that we add a bootstrap phase in-front of the join request.

I'd prefer to keep the logic as it was before i.e. no cache for redirect as that's an orthogonal change to this PR. We can make the change in a followup PR.

sn_node/src/node/bootstrap/join.rs
davidrusu: Perhaps it should, the retry logic re-generates the name, so maybe redirect should as well.

But here as well, that change seems orthogonal, we can fix this in another PR since the existing behaviour does not re-generate names.

sn_node/src/node/bootstrap/join.rs
davidrusu: Not quite, the timeout passed in is `join_time_out / 10` (see `try_join(..)` definition.

And the timeout here is to receive *any* section tree update. We've sent an AEProbe broadcast above to all elders, we are now waiting for any of them to respond.

The responses are all coming in in parallel, they get queued in the `mpsc::channel`, and `self.receive_section_tree_update()` simply waits on that channel until it sees a section tree update. They will be processed as quickly as they are received. (these are not serial requests to each elder).

> further more, the receive_section_tree_update function itself is a looping receiver , whichmeans the for clause here is bit confusing/wasted or even incorrect?

The for loop here is saying, "wait for a threshold of section tree updates to come in". The `received_section_tree_update()` waits for a single section tree update to come in.

Let me know if anything here is unclear.

sn_node/src/node/bootstrap/join.rs
davidrusu: Then what would be the purpose of the section-auth in the `JoinResponse::Redirect(section_auth)`?

My understanding here is that we are not receiving this response from the redirect target section, we are receiving it from the elders we had just sent the request to.

sn_node/src/node/bootstrap/join.rs
davidrusu: The any timeout in the for-loop will bail out (note the ??) in:

```rust
                    tokio::time::timeout(response_timeout, self.receive_section_tree_update())
                        .await
                        .map_err(|_| Error::JoinTimeout)??;
```

sn_node/src/node/messaging/node_msgs.rs
davidrusu: I tried quite a bit today to make this work. The problem is that we need a signed_sap for the network_knowledge and it's currently being retrieved from `section_tree_update.signed_sap`.

Any refactors I did spiralled into much bigger changes.

How strong is your preference to get this resolved before merging?

maqi: Ok for leave it for a follow-up PR, as long as the comment can make it clear using a `dummy` here is a really `temp` and shall have no critical impact.
and thx for clarify it.

sn_node/src/node/bootstrap/join.rs
maqi: >Then what would be the purpose of the section-auth in the JoinResponse::Redirect(section_auth)

At https://github.com/maidsafe/safe_network/blob/main/sn_node/src/node/bootstrap/join.rs#L300 we use the elders of the received section_auth within the Redicrect as the new elders target, and at https://github.com/maidsafe/safe_network/blob/main/sn_node/src/node/messaging/join.rs#L57 it's obvious that the `sap of the closest to the target section` we wrap within the Redirect message.

This check might get wrong due to a copy/paste from Retry branch ?
another possibility could be this check is to resolve the issue when the node got redirected to the self section.
However, a resolvement like this will be problematic if the node get redirected to other section.

davidrusu: My reading of that check was to ensure that this joining node's name fits within that target section we are being redirected to. (Recall that the join logic doesn't regenerate a name on redirect, names are only regenerated on retries)

Another option would be to always regenerate names, then I think we don't need to keep this prefix check since we'll regenerate the name for the target section's prefix.

sn_node/src/node/messaging/node_msgs.rs
davidrusu: I'll make this a ticket so we don't forget about it.

davidrusu: #1798

maqi (approved):

PR #1653, opened 2022-10-12
Move DkgStart auth to Dkg code instead of hidden deep in msg flows

grumbach:
`DkgStart` is now signed, checked and aggregated in Dkg code instead of relying on hidden code deep within messaging auth. 
- It is now clear when we read the code that SM of Elders need to agree on a DKG session for it to start
- Signing/Checking/Aggregating `DkgStart` is in one place instead of spread apart
- Removes now unused `OutgoingMsg::Elder`


sn_node/src/node/messaging/mod.rs
```diff
@@ -38,6 +38,8 @@ use std::collections::BTreeSet;
 pub(crate) enum OutgoingMsg {
     Node(NodeMsg),
     Client(ClientMsg),
+    #[allow(dead_code)]
+    // NB TODO will remove this soon
     Elder((SectionSigShare, Bytes)),
```
grumbach, line 43: With this PR we'll finally be able to remove this awkward variant and all the confusion that it brings in the codebase!

sn_node/src/node/mod.rs
```diff
@@ -174,6 +174,7 @@ mod core {
         pub(crate) message_aggregator: SignatureAggregator,
         pub(crate) proposal_aggregator: SignatureAggregator,
         // DKG/Split/Churn modules
+        pub(crate) dkg_start_aggregator: HashMap<Digest256, BTreeSet<SectionSigShare>>,
```
grumbach, line 177: I purposely did not use `SignatureAggregator` here as this module needs some reworking as it has timeouts

dirvine (approved):
LGTM Much cleaner and safer too.

PR #1654, opened 2022-10-12
Refactored Aggregator

grumbach:
- Simplifies `SignatureAggregator`'s innards
- Removes timers
- Distinguish the case when the aggregation is still waiting for more share from `Error`s by returning `None` instead of an `Error`
- Make DKG use this one now that timers are gone

### DIFF
--- a/sn_interface/src/messaging/signature_aggregator.rs
+++ b/sn_interface/src/messaging/signature_aggregator.rs
@@ -68,7 +68,7 @@ impl SignatureAggregator {
         hasher.finalize(&mut hash);
 
         // save the sig share
-        let current_shares = self.map.entry(hash).or_insert(BTreeSet::new());
+        let current_shares = self.map.entry(hash).or_default();
         let _ = current_shares.insert(sig_share.clone());
 
         // try aggregate
--- a/sn_interface/src/messaging/signature_aggregator.rs
+++ b/sn_interface/src/messaging/signature_aggregator.rs
@@ -7,17 +7,10 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::messaging::system::{SectionSig, SectionSigShare};
-use std::{
-    collections::BTreeMap,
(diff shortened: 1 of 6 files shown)


sn_interface/src/messaging/signature_aggregator.rs
```diff
@@ -71,90 +67,30 @@ impl SignatureAggregator {
...
-            if state.modified.elapsed() >= expiration {
-                to_remove.push(*digest);
-            }
-        }
-
-        self.map.retain(|digest, _| !to_remove.contains(digest))
-    }
-}
+        // save the sig share
+        let current_shares = self.map.entry(hash).or_insert(BTreeSet::new());
```
davidrusu, line 71: Use `.or_default()` here, `.or_insert(BTreeSet::new())` will always allocate a new `BTreeSet` even when we already have an entry for the hash.
```suggestion
        let current_shares = self.map.entry(hash).or_default();
```

sn_interface/src/messaging/signature_aggregator.rs
```diff
@@ -71,90 +67,30 @@ impl SignatureAggregator {
...
-                .combine_signatures(&self.shares)
-                .map_err(Error::Combine)?;
-
-            self.shares.clear();
+                .combine_signatures(
+                    current_shares
+                        .iter()
+                        .map(|s| (s.index, s.signature_share.clone())),
+                )
+                .map_err(AggregatorError::FailedToCombineSigShares)?;
```
davidrusu, line 83: We don't need to map the error here explicitly since you have `#[from] bls::error::Error` in `FailedToCombineSigShares(#[from] bls::error::Error)`.

`?` will do the conversion automatically.
```suggestion
                .combine_signatures(
                    current_shares
                        .iter()
                        .map(|s| (s.index, s.signature_share.clone())),
                )?;
```

davidrusu (approved):
Looks good

PR #1656, opened 2022-10-12
Remove section share auth

grumbach:
Removes `AuthKind::SectionShare` and all the awkwardness that came with it.

### DIFF
--- a/sn_node/src/comm/listener.rs
+++ b/sn_node/src/comm/listener.rs
@@ -67,8 +67,6 @@ impl MsgListener {
                     };
 
                     let src_name = match wire_msg.auth() {
-                        // NB TODO is default acceptable??
-                        AuthKind::SectionShare(_) => Default::default(),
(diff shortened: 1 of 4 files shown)


sn_node/src/comm/listener.rs
```diff
@@ -67,8 +67,6 @@ impl MsgListener {
                     };
```
grumbach, line 68: This is now solved!

sn_node/src/comm/listener.rs
```diff
@@ -67,8 +67,6 @@ impl MsgListener {
...
 
                     let src_name = match wire_msg.auth() {
-                        // NB TODO is default acceptable??
-                        AuthKind::SectionShare(_) => Default::default(),
                         AuthKind::Client(auth) => auth.public_key.into(),
                         AuthKind::Node(auth) => {
```
grumbach, line 71: Now I'm happy again 🕺

davidrusu (approved):
:fire:

PR #1657, opened 2022-10-13
Bidirectional streams for Client messages

b-zee:
- fix: safe.svg naming
- chore: dont try and establish connections for ServiceMsgs
- feat(client): force retries to use fresh connection
- ci: add client test logs
- chore(client): use same MsgId for all Cmds going out
- chore: refactor client ACK receipt
- chore(node): cleanup dropped peers
- fix: retry cmd if any sends failed
- chore(client): write query responses as they come in
- ci: fail fast
- ci: dont retry api/cli tests
- feat(node): do no requeue existing validate msg cmds
- dont force new cmd link
- ci: no retries in ci profile
- chore(node): dont manually clean up client peers
- chore(api): removing retry looping in tests
- how does this fail?
- chore(client): check for longest spentproof response
- feat(node): make sessions return if they're empty, and if so, clean them up then
- chore(client): fix unintentionally fast query retry time
- ci: disable api/cli until client is stable
- ci: have separate pre build test step for ubuntu
- ci: disable unit tests for now to focus on client
- chore(client): rejig retry loop, use max retries and keep querying if data not found until timeout
- test: add retry loop patterns to some nrs tests
- chore(client): return error responses event if we dont have 7.
- chore(client): increase query loop time
- chore(node): increase data query timeout
- chore(client): deem spentproof shares < 5 data not found
- chore(node): only perform periodic checks when cmd queue is empty
- refactor(messaging)!: removing op id from query response
- feat: use semaphore to limit processing cmds
- chore: remove intermediate function
- fix: apply clippy suggestion for iterator
- feat: use bi stream from client; process in Node
- feat: send client msg back on bidi stream


joshuef:
closing as merged into node comm refactor

PR #1661, opened 2022-10-13
(tests): assert_lists actually asserts; switch SectionPeers to using BTreeMap's instead of DashMaps

davidrusu:
- `assert_lists` checks if two lists have the same elements independent of order. It was returning Result's but it should really be asserting.

- SectionPeer reads and writes are single threaded so no need to use dashmap

### DIFF
--- a/sn_interface/src/network_knowledge/sections_dag.rs
+++ b/sn_interface/src/network_knowledge/sections_dag.rs
@@ -53,7 +53,7 @@ impl<'de> Deserialize<'de> for SectionsDAG {
 
 impl Debug for SectionsDAG {
     fn fmt(&self, formatter: &mut Formatter) -> fmt::Result {
-        let keys: Vec<_> = self.keys().cloned().collect();
+        let keys: Vec<_> = self.keys().collect();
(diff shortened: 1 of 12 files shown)


sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -304,8 +304,8 @@ impl SectionsDAG {
     }
 
     /// Iterator over all the keys in the `SectionsDAG`
-    pub fn keys(&self) -> impl Iterator<Item = &bls::PublicKey> {
-        iter::once(&self.genesis_key).chain(self.dag.all_nodes().map(|node| &node.value.key))
+    pub fn keys(&self) -> impl Iterator<Item = bls::PublicKey> + '_ {
+        iter::once(self.genesis_key).chain(self.dag.all_nodes().map(|node| node.value.key))
```
davidrusu, line 308: Every caller was converting these to owned public keys, public key's are very light so lets just do the copy here

jacderida:
Much cleaner without the `Result` on `assert_lists`! :+1:

jacderida (approved):

RolandSherwin (approved):

PR #1664, opened 2022-10-16
Comment out tests on Bochaco PR

b-zee:
Single commit on top of Bochaco PR, same as https://github.com/bochaco/safe_network/pull/5


b-zee:
I think this PR became irrelevant due to various other changes introduced since that moment.

PR #1666, opened 2022-10-17
chore(sn_client): cleanup un-necessary attempts counter

maqi:


maqi:
close as the base branch is obselete

PR #1668, opened 2022-10-17
chore(sn_client): cleanup un-necessary attempts counter

maqi:


sn_client/src/api/queries.rs
```diff
@@ -89,7 +88,7 @@ impl Client {
             let signature = self.keypair.sign(&serialised_query);
             debug!(
                 "Attempting {:?} (attempt #{}) will force new: {force_new_link}",
```
joshuef, line 90: this log needs updated as it's about `adult_index` now

joshuef (approved):

PR #1669, opened 2022-10-18
fix: make membership ae >= from gen

joshuef:
We were not providing info about the requested generation, just the one afterwards, but obviously this is not enough info for our slow node to proceed


grumbach (approved):
Looks like a good fix, this way we can unblock those stuck on the current gen

PR #1676, opened 2022-10-19
refactor(client): verify each Chunk individually in the `upload_and_verify` API

bochaco:
- This shouldn't only reduce the amount of memory used when uploading large files, but it shall also be more precise as to when a Chunk is ready for verification after uploading it, and which chunk specifically failed the verification.
- Sending a DataCmd now is not retried but only make sure it returns within the configured cmd timeout window.
- Using `expect()` in some client file tests to make sure Rust return useful info when errors occur, there is an issue with Rust test when Err is returned not showing enough failure info.

Also some changes are included to improve CI stability until some performance/design issues are solved:
- Run sn_client tests in single-threaded mode.
- Nodes to wait 30secs before retrying to joining the network when rejected.
- Testnet to launch nodes with an interval of 30secs.

### DIFF
--- a/sn_client/src/api/file_apis.rs
+++ b/sn_client/src/api/file_apis.rs
@@ -22,7 +22,7 @@ use bytes::Bytes;
 use futures::future::join_all;
 use itertools::Itertools;
 use self_encryption::{self, ChunkInfo, DataMap, EncryptedChunk};
-use tokio::task;
+use tokio::{task, time::sleep};
 use tracing::trace;
 use xor_name::XorName;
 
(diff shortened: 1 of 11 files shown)


sn_client/src/api/file_apis.rs
```diff
@@ -194,15 +167,36 @@ impl Client {
...
     #[instrument(skip_all, level = "trace")]
-    async fn upload_large(&self, large: LargeFile) -> Result<XorName> {
+    async fn upload_large(&self, large: LargeFile, verify: bool) -> Result<XorName> {
         let (head_address, all_chunks) = Self::encrypt_large(large)?;
 
         let tasks = all_chunks.into_iter().map(|chunk| {
-            let writer = self.clone();
-            task::spawn(async move { writer.send_cmd(DataCmd::StoreChunk(chunk)).await })
+            let client_clone = self.clone();
+            task::spawn(async move {
```
maqi, line 192: just to confirm, so `send_cmd` will be timed out after `self.cmd_timeout` , then `verify_chunk_is_stored` will be timed out after `self.query_timeout`.
Which means the total time out for the upload will be `self.cmd_timeout + self.query_timeout` . is it ok?

sn_client/src/api/file_apis.rs
```diff
@@ -222,16 +216,41 @@ impl Client {
...
+        // reached adults...and so, we retry once but within the timeout limit set
+        let _chunk = tokio::time::timeout(self.query_timeout, async {
+            let mut response = self.get_chunk(&address).await;
+            if response.is_err() {
+                error!(
+                    "Error when attempting to verify chunk was stored at {:?}: {:?}",
+                    address, response
+                );
+
+                sleep(self.max_backoff_interval).await;
```
maqi, line 244: `self.max_backoff_interval` is defaulted to `DEFAULT_MAX_QUERY_CMD_BACKOFF_INTERVAL` which is 10s,
meanwhile `self.query_timeout` is 15s, will the sleep being too long?

sn_client/src/api/file_apis.rs
bochaco: if you are verifying it was stored yes

sn_client/src/api/file_apis.rs
bochaco: `DEFAULT_MAX_QUERY_CMD_BACKOFF_INTERVAL` is 3secs and query_timeout default is 45secs from my understanding

joshuef (approved):

PR #1680, opened 2022-10-20
refactor(cmds): replace ack+error with response

oetyng:
Replaces `CmdAck` + `CmdError` with `CmdResponse`.
Still to do: 
- Move responses creation to where it is finally handled (e.g. Adults for data).

BREAKING CHANGE: `ClientMsg` enum variants changed.


sn_node/src/node/messaging/client_msgs.rs
```diff
@@ -320,7 +376,24 @@ impl MyNode {
             traceroute,
         ));
 
-        Ok(cmds)
+        cmds
+    }
+
+    fn extract_data(&self, cmd: SpentbookCmd) -> Result<ReplicatedData> {
```
joshuef, line 382: `extract_data` doesn't feel clear enough to me. `get/extract_replicated_data` perhaps. 

Also some docs here would be good.

joshuef:
Closing as this has meen merged atop a rebased NodeCommRefactor branch

PR #1685, opened 2022-10-21
Remove node auth

grumbach:
Remove signatures from node messages
- remove `msg_authority` from `MsgType::Node`
- remove the `NodeMsgAuthority` type altogether

### DIFF
--- a/sn_interface/src/messaging/serialisation/wire_msg.rs
+++ b/sn_interface/src/messaging/serialisation/wire_msg.rs
@@ -9,7 +9,7 @@
 use super::wire_msg_header::WireMsgHeader;
 use crate::messaging::{
     data::ClientMsg, system::NodeMsg, AuthKind, AuthorityProof, ClientAuth, Dst, Error, MsgId,
-    MsgType, NodeMsgAuthority, Result,
+    MsgType, Result,
 };
 use bytes::Bytes;
 use custom_debug::Debug;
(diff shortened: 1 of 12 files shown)


sn_interface/src/messaging/serialisation/wire_msg.rs
```diff
@@ -251,17 +251,14 @@ impl WireMsg {
                     msg,
                 })
             }
-            AuthKind::Node(node_signed) => {
+            // NB TODO remove this enum
+            AuthKind::Node(_) => {
```
dirvine, line 255: I thnk this should be the next PR when this one is merged.

dirvine (approved):
This code now reads significantly easier.  Looks good to me

sn_node/src/node/messaging/anti_entropy.rs
```diff
@@ -483,51 +482,6 @@ mod tests {
             .await
     }
 
-    #[tokio::test]
```
joshuef, line 486: @grumbach is this test not still valid? It's all about the SAP proof chain,as opposed to signed msgs

sn_interface/src/messaging/serialisation/wire_msg.rs
grumbach: Yes that was the plan!

sn_node/src/node/messaging/anti_entropy.rs
grumbach: It's testing a function that doesn't exist anymore, testing the key in the msg auth. 
Similar checks are also in the update function for section tree, so it was redundant.

PR #1691, opened 2022-10-23
refactor(data): move outside of node

oetyng:
BREAKING CHANGE: Name change of node msg.


sn_node/src/data/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+        /// The data this peer currently holds.
+        currently_at_peer: Vec<DataAddress>,
+        /// Other peers we know of.
+        other_peers: Vec<XorName>,
+    },
+}
+
+/// Events produced in the data module.
+#[derive(Debug, Clone)]
+pub enum Event {
```
davidrusu, line 198: Are we still using these events for some business logic?

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -49,6 +51,13 @@ impl Dispatcher {
         trace!("doing actual processing cmd: {cmd:?}");
 
         match cmd {
+            Cmd::Data(cmd) => {
+                if let Some(event) = self.data.handle(cmd).await? {
+                    Ok(self.process_data_event(event).await.into_iter().collect())
+                } else {
+                    Ok(vec![])
+                }
+            }
```
davidrusu, line 60: Why don't we have `data.handle(..)` returning commands as well? Any reasons why we need to use DataEvent's here?

sn_node/src/node/mod.rs
```diff
@@ -527,20 +510,38 @@ mod core {
...
+                    .network_knowledge()
+                    .adults()
+                    .iter()
+                    .map(Peer::name)
+                    .collect();
+                let added: BTreeSet<_> = current.difference(&old.adults).copied().collect();
+                let removed: BTreeSet<_> = old.adults.difference(&current).copied().collect();
+                let adults_changed = !added.is_empty() || !removed.is_empty();
+
+                if adults_changed {
```
davidrusu, line 534: Is this not the same as:
```suggestion
                if current != old.adults {
```

davidrusu (commented):
It's looking good!

My main question is why we need the `DataEvent`. Can we not have `Data::handle` return commands?

sn_node/src/node/mod.rs
oetyng: Yes, I believe so. Remnant from when `added`, `removed` and `remaining` where used variables 👍

sn_node/src/data/mod.rs
oetyng: I believe these are correct to keep, as to actually have the decoupling. Otherwise we introduce the full range of cmds into Data module, while those contain all sorts of things related to the node. It shouldn't be possible to call the node and ask it to do various things, from within data module, as that breaks inversion of control.

sn_node/src/node/flow_ctrl/dispatcher.rs
oetyng: It shouldn't be possible to call the node and ask it to do various things, from within data module, as that breaks inversion of control.

sn_interface/src/network_knowledge/node_info.rs
```diff
@@ -37,8 +37,12 @@ impl MyNodeInfo {
         Peer::new(self.name(), self.addr)
     }
 
+    pub fn id(&self) -> PublicKey {
```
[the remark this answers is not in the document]
oetyng: Also I see that the public key is used unnecessarily for those flows, since old.

davidrusu (approved):
:ship:

sn_node/benches/data_storage.rs
```diff
@@ -103,22 +104,17 @@ fn bench_data_storage_writes(c: &mut Criterion) -> Result<()> {
...
-            .map(|_| create_random_register_replicated_data())
-            .collect();
+        let data_set: Vec<_> = (0..*size).map(|_| create_random_register_cmd()).collect();
         group.bench_with_input(
             BenchmarkId::new("register_writes", size),
             &(size, &data_set),
             |b, (size, data_set)| {
-                let storage = get_new_data_store()
-                    .context("Could not create a temp data store")
+                let storage = get_new_register_store()
```
joshuef, line 112: Can we maintain one interface for storing of data. Feels cleaner to me

joshuef (changes requested):
I think we need to break this up more. 

There's some renaming here that could go in it's own PR that'll be much easier to get in. So it'd be great to get those pulled out as a first step.

-----

Also I'm not sold on DataEvents... I think we may need to more solidly look at the API we want of a DataStorage module.

To my mind beyond the API itself it should be a black box (so that it is replaceable with some other impl) if we're going that way. Having the `/node` code need to handle `DataEvents` feels more complex than necessary.

---

w/r/t the data API. I also think we may well be wanting some sync calls. Or at least the option of sync calls from an API standpoint (thinking w/r/t returning ACKs only when data has been stored)

oetyng:
> w/r/t the data API. I also think we may well be wanting some sync calls. Or at least the option of sync calls from an API standpoint (thinking w/r/t returning ACKs only when data has been stored)

Yeah, I'll be revisiting this later like we discussed, so we can nail the sync aspect of ACKS.

> Also I'm not sold on DataEvents... I think we may need to more solidly look at the API we want of a DataStorage module.

Sure, we'll have to figure out what makes most sense to actually reach the decoupling.

> To my mind beyond the API itself it should be a black box (so that it is replaceable with some other impl) if we're going that way. Having the `/node` code need to handle `DataEvents` feels more complex than necessary.

Absolutely agree about the black box and being able to replace it with some other impl.
Also for usage of the actual `DataStorage` in other contexts than our network (including but not limited to; examples, tests etc.).

Events, sure that is a question of both pro's and cons.

In one aspect, our whole system is basically built on events, it's just a matter of syntax and structure.
Also when it comes to testability and debugging, cmds->events has lots of benefits. (These are actually the main ones I've been hoping that we would be able to get.)

In another aspect, there is a more feature rich and specific use of events that can - and cannot - be suitable for different systems.

The most suitable types of systems using the full fledged pattern is highly concurrent systems with many different parts concurrently operating on results from other parts. And our system doesn't actually look quite like that right now (and may never do).

***

The two common approaches to decoupling that I think of now are messaging and interfaces (i.e. traits in rust).
Since we do not use traits much, and already have a fundamentally message based system, I thought the message way was appropriate.

But the devils are in the details, so we'll have to look into the tradeoffs closer.

One way or another, I still think we really should decouple `DataStorage` from `MyNode`.

oetyng:
Will revisit this later.

PR #1700, opened 2022-10-25
chore: impl custom `Debug` trait to compact the logs

RolandSherwin:
~~- Readable `AntiEntropyKind` debug messages; previously it printed out the entire binary~~
- `NodeState` is compacted to use half the characters
- `SectionAuthorityProvider` now provides more information while being compact

### DIFF
--- a/sn_interface/src/messaging/system/mod.rs
+++ b/sn_interface/src/messaging/system/mod.rs
@@ -33,16 +33,22 @@ use xor_name::XorName;
 /// List of peers of a section
 pub type SectionPeers = BTreeSet<SectionSigned<NodeState>>;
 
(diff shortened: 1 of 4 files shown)


sn_interface/src/messaging/system/mod.rs
```diff
@@ -47,6 +47,18 @@ pub enum AntiEntropyKind {
     Update { members: SectionPeers },
 }
 
+impl Debug for AntiEntropyKind {
```
joshuef, line 50: you should just be able to use the `custom_debug_derive` impl to skip this via macro, eg: https://github.com/maidsafe/safe_network/blob/main/sn_node/src/comm/peer_session.rs#L198

joshuef (approved):

PR #1701, opened 2022-10-25
Disentangle modules

oetyng:
On top of decouple-data-storage.


oetyng:
Will revisit this later.

PR #1702, opened 2022-10-25
Reenable cmd ids and child ids

oetyng:
Used for logging purposes.

***

The id of a cmd is a vec, where first entry represents the sequential nr of incoming cmds (does not increment by subsequent cmds - i.e. child cmds - spawned from such a cmd). 

I.e. `[0]` is the first cmd, `[235]` is the 235th cmd.

Second entry is the sequential nr of the child cmd spawned from the cmd at first entry, i.e. `[0, 2]` is the third child cmd of the first cmd, `[235, 0]` is the first child cmd of the 235th cmd.
And the first child cmd resulting from child `[235, 0]` would have the id `[235, 0, 0]`.

And so on. 

The trace back to the initial cmd is preserved, by entering the sequence nr of each child at every level of new children.

### DIFF
--- a/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
+++ b/sn_node/src/node/flow_ctrl/cmd_ctrl.rs
@@ -8,24 +8,20 @@
 
 use crate::node::flow_ctrl::{cmds::Cmd, dispatcher::Dispatcher};
 
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc,
+};
 use tokio::sync::RwLock;
 use xor_name::XorName;
 
-/// A module for enhanced flow control.
-///
-/// Orders the incoming cmds (work) according to their priority,
-/// and executes them in a controlled way, taking into
-/// account the rate limits of our system load monitoring.
-///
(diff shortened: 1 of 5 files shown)


sn_node/src/node/flow_ctrl/cmd_ctrl.rs
```diff
@@ -45,29 +48,36 @@ impl CmdCtrl {
 
     /// Processes the next priority cmd
     pub(crate) async fn process_cmd_job(
-        dispatcher: Arc<Dispatcher>,
+        &self,
         cmd: Cmd,
-        id: Option<usize>,
+        mut id: Vec<usize>,
```
joshuef, line 53: It's not obvious to me why the id is a vec now?

oetyng: I had an explanation in the code comments. But I added additional clarification to the PR description now.

joshuef (approved):

sn_node/src/node/flow_ctrl/cmd_ctrl.rs
joshuef: awesome thanks for clarifying

PR #1707, opened 2022-10-26
Node adult bi

joshuef:

### DIFF
--- a/sn_node/src/comm/link.rs
+++ b/sn_node/src/comm/link.rs
@@ -187,38 +187,38 @@ impl Link {
             trace!("Grabbing a connection from link.. {:?}", self.peer());
             // temp store for dead conn ids. We cannot remove them in a loop reading conns
             // otherwise DashMap might deadlock, so we need to do it afterwards.
-            let mut dead_conns = vec![];
+            // let mut dead_conns = vec![];
             let mut live_conn = None;
             for entry in self.connections.iter() {
                 let conn = entry.value();
                 debug!("iterating conns");
 
-                // TODO: replace this with simple connection check when available.
-                let is_connected = conn.open_bi().await.is_ok();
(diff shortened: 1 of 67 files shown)


sn_node/src/comm/link.rs
```diff
@@ -127,27 +131,25 @@ impl Link {
         recv_stream.next().await.map_err(SendToOneError::Recv)
     }
 
-    pub(crate) async fn get_or_connect(
-        &mut self,
-        should_establish_new_connection: bool,
-    ) -> Result<Connection, SendToOneError> {
+    pub(crate) async fn get_or_connect(&mut self) -> Result<Connection, SendToOneError> {
```
oetyng, line 134:
```
pub(crate) async fn get_or_connect(&mut self) -> Result<Connection, SendToOneError> {
    if let Some(conn) = self.connections.iter().next() {
            trace!("Grabbing a connection from link..");
            return Ok(conn.value().clone());
        } else {
            debug!("attempting to create a connection");
            self.create_connection().await        
        }
    }
```

sn_node/src/comm/mod.rs
```diff
@@ -166,6 +163,25 @@ impl Comm {
         Ok(())
     }
 
+    // TODO: can we not serialize w/ client query response from adult and just pass through?
```
oetyng, line 166: > // TODO: can we not serialize w/ client query response from adult and just pass through?

Should be possible, but then we'll lose the type safe msg boundaries (Adults only deal in `NodeMsgs`, not `ClientMsgs`, since they are not supposed to have direct contact with them).

Seems there is no processing at the Elder of the response beyond trying to deserialize it into a proper msg.

I was thinking that the client had to report back to elders if a bad msg was sent out by adults, but I see there is no sig from adults on it, so it can't be proven..

PR #1710, opened 2022-10-26
chore(node): move conn acquisition off thread too.

joshuef:
This was currently potentially a blocker if there'd been connection issues. And might prevent other messages being sent


dirvine (approved):
Nice one

PR #1716, opened 2022-10-27
Cleanup around DKG/Handover, remove Proposal duplication

grumbach:
- Renamed confusing/collisional `Proposal::SectionInfo` to clearer `Proposal::RequestHandover`
- Removed `agreement.rs` and put content in files that make more sense:
    - Moved `SectionSigned` to `section_sig.rs`
    - Moved `DkgSessionId` to `dkg.rs`
    - Moved `Proposal` to `proposal.rs`
- Updated docs around Dkg and Proposals
- Got rid of duplicate `Proposal` type in `sn_node` causing confusion/duplication and unnecessary conversions

### DIFF
--- a/sn_interface/src/messaging/system/proposal.rs
+++ b/sn_interface/src/messaging/system/proposal.rs
@@ -15,7 +15,7 @@ use crate::{
 use serde::{Deserialize, Serialize};
 
 /// A Proposal about the state of the section
-/// This can be a result of seeing a node come online, go offline, changes to section info etc.
+/// This can be a result of seeing a node go offline, changes to section info etc.
 /// Anything where we need section authority before action can be taken
 /// Proposals sent by elders or elder candidates, to elders or elder candidates
 #[allow(clippy::large_enum_variant)]
--- a/sn_interface/src/messaging/system/proposal.rs
+++ b/sn_interface/src/messaging/system/proposal.rs
@@ -6,25 +6,36 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use crate::node::Result;
-
-use sn_interface::{
-    messaging::system::{Proposal as ProposalMsg, SectionSigShare, SectionSigned},
+use crate::{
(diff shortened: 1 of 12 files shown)


sn_interface/src/messaging/system/proposal.rs
```diff
@@ -6,25 +6,36 @@
...
-    messaging::system::{Proposal as ProposalMsg, SectionSigShare, SectionSigned},
+use crate::{
+    messaging::system::{SectionSigShare, SectionSigned},
+    messaging::{Error, Result},
     network_knowledge::{NodeState, SectionAuthorityProvider},
 };
 
+use serde::{Deserialize, Serialize};
+
+/// A Proposal about the state of the network
```
maqi, line 17: May not related to this PR, however the `Proposal` has an entry to `vote node off` but doesn't have an entry to `vote node on (or say recruit a node)`.
I understand that part moved into `decision` , but if so, maybe this Proposal is not about `the state of the network`, but about `the sate of elders on the network` ?

grumbach: Yeah you're right, possibly even the `state of the section` since we can vote off nodes that are not elders?

sn_interface/src/messaging/system/proposal.rs
```diff
@@ -6,25 +6,36 @@
...
+use crate::{
+    messaging::system::{SectionSigShare, SectionSigned},
+    messaging::{Error, Result},
     network_knowledge::{NodeState, SectionAuthorityProvider},
 };
 
+use serde::{Deserialize, Serialize};
+
+/// A Proposal about the state of the section
+/// This can be a result of seeing a node come online, go offline, changes to section info etc.
```
maqi, line 18: `This can be a result of seeing a node come online`
seems we no longer cast `VoteNodeOn` votes via proposal ? especially for adults?

grumbach: Oh right!

maqi (approved):

maqi:
replaced by PR https://github.com/maidsafe/safe_network/pull/1780

PR #1719, opened 2022-10-27
ci: reorganise jobs in pr workflow

jacderida:
We now have infrastructure to spin up an EC2 instance if the job has been marked with `runs_on:
self-hosted`. These larger instances can help reduce the time taken for the overall workflow run.

In light of this, the jobs for the PR workflow have been redefined as follows:

    * Various checks: a series of miscellaneous checks like `cargo fmt` and `cargo clippy`. The job for
      linting the commits was combined into this job since there didn't seem to be a good reason for it
      to use its own machine.
    * Unused dependency check: as before. It would have been good to combine this job in the one above,
      but it needs the nightly toolchain.
    * Build node and testnet binaries: use an EC2 instance to quickly build the node and testnet
      binaries, then upload the artifacts to S3. The purpose is to reduce build time on other jobs. The
      node is built using musl because there were issues on the GHA machines with glibc when trying to
      run the node. All jobs below will wait on this job to complete to access the artifacts it
      produces.
    * Unit tests and E2E tests: use an EC2 instance to run the unit tests, then run the E2E test suite
      against a network running on the same instance. The unit tests used to be in their own job, but
      compilation time made it very slow. Since they don't take that long to run, we can run them on the
      large EC2 instance while we have it.
    * E2E tests: as before, except we use the `sn_node` and `testnet` binaries from the build job to
      reduce some time.
    * API tests: as before, except we use the `sn_node` and `testnet` binaries from the build job to
      reduce some time.
    * CLI tests: as before, except we use the `sn_node` and `testnet` binaries from the build job to
      reduce some time.

The `testnet` binary was also changed to check for the existence of the node binary at
`~/.safe/node/sn_node` to avoid performing a `cargo build` if you already have a built node.

### DIFF
--- a/.github/workflows/run_pr_checks.yml
+++ b/.github/workflows/run_pr_checks.yml
@@ -12,12 +12,26 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  CARGO_INCREMENTAL: 0  # bookkeeping for incremental builds has overhead, not useful in CI.
+  CARGO_INCREMENTAL: 0 # bookkeeping for incremental builds has overhead, not useful in CI.
   SAFE_AUTH_PASSPHRASE: "x"
   SAFE_AUTH_PASSWORD: "y"
   NODE_COUNT: 15
+  BUILD_ARTIFACTS_BUCKET: s3://maidsafe-build-artifacts/safe_network
+  BUILD_ARTIFACTS_URL: https://maidsafe-build-artifacts.s3.eu-west-2.amazonaws.com/safe_network
 
 jobs:
+  # TODO: Reenable when blst has been updated and this isn't just red the whole time.
+
+  # cargo-deny:
+  #   if: "!startsWith(github.event.pull_request.title, 'Automated version bump')"
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - uses: actions/checkout@v2
+
+  #   - shell: bash
+  #     run: wget https://raw.githubusercontent.com/maidsafe/QA/master/misc-scripts/deny.toml
+
+  #   - uses: EmbarkStudios/cargo-deny-action@v1
   cargo-udeps:
     if: "!startsWith(github.event.pull_request.title, 'Automated version bump')"
     name: Unused dependency check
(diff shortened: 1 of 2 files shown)


.github/workflows/run_pr_checks.yml
```diff
@@ -85,69 +79,36 @@ jobs:
...
-      - name: Run sn_cli tests
-        timeout-minutes: 25
-        run: cd sn_cli && cargo test --release --bin safe
+      - name: build sn_node and testnet
+        run: |
+          cargo build --release --bin sn_node
+          cargo build --release --bin testnet
+      - name: upload artifacts to s3
+        run: |
+          tar -zcvf target-$GITHUB_RUN_ID.tar.gz -C target/ .
```
joshuef, line 103: So we need all these artifacts? I would have just thought we needed `sn_node` and `testnet` bins?

.github/workflows/run_pr_checks.yml
```diff
@@ -253,177 +214,115 @@ jobs:
...
+
+      - uses: Swatinem/rust-cache@v1
+        continue-on-error: true
+        with:
+          cache-on-failure: true
+          sharedKey: ${{github.run_id}}
+
+      - name: Build sn bins
+        run: |
+          cd sn_node && cargo build --release --bins
```
joshuef, line 236: Can we not pull this from the prior run?

.github/workflows/run_pr_checks.yml
```diff
@@ -253,177 +214,115 @@ jobs:
...
+
+      - name: Build testnet
```
joshuef, line 238: and this...?

joshuef:
I'm not sure EC2 makes sense for CLI (which I think could alllso pull artifacts from a prior build run no?).

At the moment, we're getting a ~18 min speed up for all the tests.... about 40% faster. But from aws billing, we already are paying $50 for that, against... however many test runs we have. So I'm not sure how worthwhile that is (if we had it against all the PRs eg). 

-----

I think @jacderida that if we get only the bins up/down. And use them all over the other jobs... but have those running on GHA infra we're more viable. 

We could then have the split/churn tests go on the AWS infra for the bors `merge` job to further reduce when we need to run those?


----

What do you reckon?

.github/workflows/run_pr_checks.yml
jacderida: Yeah, that's right, and actually that was what I had originally. This was an experiment to see if having the whole target directory could help speed up the slow compilation times for the CLI job. It didn't help though. I didn't change it back because it's lightning fast to copy stuff from an EC2 machine to S3, so I didn't really notice it. I'll change it back now though.

.github/workflows/run_pr_checks.yml
jacderida: The reason for this was so that this job could start right away rather than wait for the `build` job. It's not really a big deal to build on the EC2 instance, so it meant this job could just go ahead without waiting. I'll change it to be dependent on the build job, but I don't think it'll make a huge difference.

jacderida:
> I'm not sure EC2 makes sense for CLI (which I think could alllso pull artifacts from a prior build run no?).
> 
> At the moment, we're getting a ~18 min speed up for all the tests.... about 40% faster. But from aws billing, we already are paying $50 for that, against... however many test runs we have. So I'm not sure how worthwhile that is (if we had it against all the PRs eg).
> 
> I think @jacderida that if we get only the bins up/down. And use them all over the other jobs... but have those running on GHA infra we're more viable.
> 
> We could then have the split/churn tests go on the AWS infra for the bors `merge` job to further reduce when we need to run those?
> 
> What do you reckon?

OK no problem. The CLI job is really slow even with building the `sn_node` and `testnet` binaries on the other machine because it takes so long to compile the `safe` binary and the testing code on the GHA infrastructure. I'll change it back.

joshuef (approved):

PR #1721, opened 2022-10-28
feat(join): remove age stepping in genesis sections

davidrusu:
There used to be a worry about too many nodes being relocated at the same time but that is no longer a concern there since we only relocate one node at a time.

This PR remove the age stepping logic so that the genesis sections do not have a special join process.

### DIFF
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -25,7 +25,7 @@ use backoff::{backoff::Backoff, ExponentialBackoff};
 use bls::PublicKey as BlsPublicKey;
 use futures::future;
 
-use std::collections::{BTreeMap, BTreeSet};
+use std::collections::BTreeSet;
 use tokio::{sync::mpsc, time::Duration};
 use tracing::Instrument;
 use xor_name::Prefix;
@@ -65,8 +65,7 @@ struct Joiner<'a> {
     prefix: Prefix,
     section_tree: SectionTree,
     backoff: ExponentialBackoff,
-    aggregated: bool,
-    retry_responses_cache: BTreeMap<Peer, u8>,
+    retry_responses_cache: BTreeSet<Peer>,
 }
 
 impl<'a> Joiner<'a> {
@@ -93,7 +92,6 @@ impl<'a> Joiner<'a> {
             prefix: Prefix::default(),
             section_tree,
             backoff,
-            aggregated: false,
             retry_responses_cache: Default::default(),
         }
     }
@@ -194,15 +192,13 @@ impl<'a> Joiner<'a> {
                 }
                 JoinResponse::Retry {
                     section_tree_update,
-                    expected_age,
                 } => {
                     let signed_sap = section_tree_update.signed_sap.clone();
 
                     trace!(
-                        "Joining node {:?} - {:?}/{:?} received a Retry from {}, expected_age: {expected_age}, SAP: {}, proof_chain: {:?}",
+                        "Joining node {:?} - {:?} received a Retry from {}, SAP: {}, proof_chain: {:?}",
                         self.prefix,
                         self.node.name(),
-                        self.node.age(),
                         sender.name(),
                         signed_sap.value,
                         section_tree_update.proof_chain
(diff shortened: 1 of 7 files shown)


sn_node/src/node/messaging/join.rs
```diff
@@ -78,26 +74,28 @@ impl MyNode {
...
             trace!("{}", LogMarker::SendJoinRetryAgeIssue);
-            trace!("JoinRequest from {peer} (with age {}) doesn't have the expected age: {expected_age}", peer.age());
+            trace!(
+                "JoinRequest from {peer} (with age {}) has invalid age",
+                peer.age()
+            );
         }
 
-        if !section_key_matches || is_age_invalid {
+        if !section_key_matches || !is_age_valid {
```
joshuef, line 94: Just to confirm the flow:

Adults know their age is off now, due to receiving the SAP and knowing how many elders there are, right?

If they get a sap that's not what they've been applying with, they'll adjust their age and reapply. 

(So the check here basically ignores invalid ages? perhaps we can shortcircuit that earlier)

sn_node/src/node/bootstrap/join.rs
```diff
@@ -226,31 +222,13 @@ impl<'a> Joiner<'a> {
...
-                            new_prefix = new_prefix.pushed(push_prefix_0);
-                            cur_age /= 2;
-                        }
-                        trace!("Name shall have the prefix of {:?}", new_prefix);
-
+                    if self.insert_retry_response(sender, signed_sap.elders_set()) {
+                        trace!("Re-generating name for retry");
                         let new_keypair =
-                            ed25519::gen_keypair(&new_prefix.range_inclusive(), expected_age);
+                            ed25519::gen_keypair(&prefix.range_inclusive(), self.node.age());
```
maqi, line 228: previously, the new node's keypair (xorname) will be generated bearring a prefix following it's stepped age.
This is to ensure they have an even split on the second prefix bit, so that we can have a section aplit around the 28th-30th nodes.
Without it, we experienced before that even with 50 nodes, there is no section split happens just because the spreading of nodes' names is not even. And CI machine does have issue with too many nodes running.

Also, the even split allows the 7 nodes within the genesis section to be evenly split into the two sub-sections.
This makes the handover flow more smooth, as there is `old elders` from the genesis section to pass down the previous network knowledge to the `new elders`.
We had issue previously that a new sub-section with all 7 new elders lost some old network knowledge.
This may get resolved with the recent refactor work, but need to be tested carefully.

sn_node/src/node/messaging/join.rs
davidrusu: Well, sort of.

This PR makes it so that all joining nodes must start at `MIN_ADULT_AGE`. So if a node attempts to join with a different age, then this logic will reject them.

But the joining logic now always tries to join with `MIN_ADULT_AGE` as well so if there is a mismatch, then the joining node is faulty or there is a bug.

> If they get a sap that's not what they've been applying with, they'll adjust their age and reapply.

The change here is to stop adjusting adult ages, we'll no longer step the ages of the first nodes joining the section.

sn_node/src/node/bootstrap/join.rs
davidrusu: > previously, the new node's keypair (xorname) will be generated bearring a prefix following it's stepped age.
This is to ensure they have an even split on the second prefix bit, so that we can have a section aplit around the 28th-30th nodes.

I think we've addressed this problem with two changes:
1. we allow nodes two join even if the section has passed the recommended section size as long as they contribute towards a split: https://github.com/maidsafe/safe_network/blob/main/sn_interface/src/network_knowledge/mod.rs#L124-L147
2. nodes retry with a random name
It may take a few retries but nodes will eventually find the right prefix to join the section without needing to co-ordinate ages.

> We had issue previously that a new sub-section with all 7 new elders lost some old network knowledge.
This may get resolved with the recent refactor work, but need to be tested carefully.

That sounds like a bug that the age stepping was hiding. I don't see why this scenario can't happen in future (non-genesis) sections as well.

sn_node/src/node/messaging/join.rs
joshuef: Okay right, thanks for the clarifications 💪

joshuef (approved):

PR #1724, opened 2022-10-31
test(client): feature-gated send_query API to verify data is stored in all replicas for each query

bochaco:
- Introducing a new `check-replicas` feature to sn_client (as well as sn_api and CLI), which switches the `Client::send_query` API behavior to send the query to all Adults replicas and it expects a valid response from all of them instead of from just one.
- Running sn_client, sn_api, and CLI tests in CI with new `check-replicas` feature enabled to verify data was stored in all Adults replicas.
- With `check-replicas` feature enabled, query responses from all Adults replicas are also compared to verify they are all the same.
- Also fixing issue in sn_node comm: create new connection to peer if a SendStream cannot be finished when sending a msg.


sn_client/src/errors.rs
```diff
@@ -221,4 +221,8 @@ pub enum Error {
     /// Occurs if a section key is not found when searching the sections DAG.
     #[error("Section key {0:?} was not found in the sections DAG")]
     SectionsDagKeyNotFound(PublicKey),
+    /// Data replicas check error
+    #[cfg(feature = "check-replicas")]
+    #[error("Data replicas check failed: {0}")]
+    DataReplicasCheck(String),
```
joshuef, line 227: Can we have two distinct errors here to avoid needing to use `String` which leaves this a bit open to error abuse?

bochaco: sounds good, note I'm not returning a single error though, so I'll try to see how to show the list of errors/responses as I'm doing now `[ Adult-#0: <Error> ],  Adult-#2: <Error> ], ...`

joshuef (approved):

PR #1727, opened 2022-11-01
chore(ci): kill all nodes before packaging logs to upload

bochaco:


.github/workflows/sn_pr_tests.yml
```diff
@@ -411,6 +420,15 @@ jobs:
   #         echo "$(pgrep sn_node | wc -l) nodes still running"
   #         ls $HOME/.safe/node/local-test-network
 
+  #      - name: Kill all nodes
```
joshuef, line 423: lets remove this if we're doing it above then :+1

joshuef (changes requested):

.github/workflows/sn_pr_tests.yml
bochaco: do you mean in the previous step? that's only listing them...or what do you mean?

joshuef: i just meant the commented code for `kill all nodes`.

bochaco: But the earlier one is for the other job, i.e. `e2e` job, this one is for whenever we re-enable `e2e-split` job, sorry, but I don't think I understand?

joshuef: Ahhh, no. I don't think _i_ understood. All good then

joshuef (approved):

PR #1729, opened 2022-11-02
refactor: test utilities

RolandSherwin:
- Removes a test utility wrapper for `bls::SecretKeySet` . It was previously wrapped because there was no way to get `SecretKey` from the set. The current version of `bls::SecretKeySet` provides a method to do so.
- Organizes test utility functions into their own structs, e.g., `TestKeys` for all the key related utils, etc,. 
- Adds an utility to get `SecretKeySet` from a list of `SecretKeyShares`
- Adds an utility to get `NodeInfo` with varying age
- Removes unused dependency from `flow_ctrl/tests/`


b-zee (approved):
Looks okay, can't find anything out of the ordinary.

PR #1730, opened 2022-11-03
Bidi streams no client refactor

joshuef:


b-zee (commented):
The commit `chore: dont debug log AE bounced_msg ` doesn't really do what it says?

sn_node/src/node/data/records.rs
```diff
@@ -85,66 +156,165 @@ impl MyNode {
...
+                correlation_id: msg_id,
+            };
+
+            let (kind, payload) = self.serialize_sign_client_msg(client_msg)?;
+
+            if let Some(stream) = client_response_stream {
+                self.send_msg_on_stream(payload, kind, stream, Some(target), msg_id, traceroute)
+                    .await?
+            } else {
+                warn!("No send stream to respond to client on! (This is likely due to the msg originating at the elder as a healthcheck");
```
b-zee, line 231: I do want to note that such a warning might have to be a hard error in some cases. I guess this depends on the context, but assuming context is pretty dangerous I feel. I think this is something we definitely need to addres later, or make the routines even more direct so that we have separate flows for 'healthchecks' and client requests?

sn_node/src/node/messaging/client_msgs.rs
```diff
@@ -157,76 +146,37 @@ impl MyNode {
...
-        let query_peers = self.pending_data_queries.remove(&(op_id, sender_name));
+        )?;
 
-        // First check for waiting peers. If no one is waiting, we drop the response
-        let waiting_peers = if let Some(peers) = query_peers {
-            if peers.is_empty() {
-                warn!("No waiting peers to send {op_id:?} to....");
-                // nothing to do
-                return vec![];
+        if let Some(send_stream) = send_stream {
```
b-zee, line 160: There is no else here, what if the send stream is missing?

sn_node/src/node/data/records.rs
joshuef: Yeh totally agree. We need more concrete separation of those msgs which _need_ responses, and others...

sn_node/src/node/messaging/client_msgs.rs
joshuef: added in an error msg there 👍

sn_client/src/connections/messaging.rs
```diff
@@ -589,90 +592,72 @@ impl Session {
...
-                            }
-                        }
-                    };
+        let mut tasks = vec![];
+        let nodes_len = nodes.len();
 
-                    if let Err(err) = &result {
-                        session.peer_links.remove_link_from_peer_links(&peer).await;
-                        warn!("Issue when sending {msg_id:?} to {peer:?}: {err:?}");
+        for peer in nodes.clone() {
```
bochaco, line 604: It seems we are removing the logic to retry once when a connection lost is reported by the `Link::send_bi`, is that intentional?

sn_node/src/comm/peer_session.rs
```diff
@@ -142,11 +162,14 @@ impl PeerSessionWorker {
                                     job.msg_id
                                 );
                             }
+
+                            job.reporter.send(SendStatus::Sent);
```
bochaco, line 166: We are reporting as `SendStatus::Sent` even when it failed to send the msg, is this ok or it can have bad consequences for whoever is listening to this status report?

sn_client/src/connections/messaging.rs
joshuef: So I remove conns in the `get_or_connect` in the link itself, in order to attempt to shortcircuit those retries. So I don' think they should be necessary? I may be wrong though

sn_node/src/comm/peer_session.rs
joshuef: ah, not intentional. I must have thought we were exiting on err above

sn_client/src/connections/messaging.rs
bochaco: I think, in general, even if we try to check the connection is ok before sending a msg, there could always be a race condition where the conn is lost right after we checked but right before we try to send the msg....it feels to me that we cannot really tell if the conn is ok till we try to send a msg on it, ....if this is the case then we'll have to always act upon a conn failure when sending the msg as we wouldn't be able to avoid such scenario...?.....

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1090,63 +1090,6 @@ async fn handle_demote_during_split() -> Result<()> {
         .await
 }
 
-#[tokio::test]
```
bochaco, line 1093: Perhaps mark these two tests as `ignore` with a TODO reason/message so we refactor it later on?

sn_node/src/node/messaging/mod.rs
```diff
@@ -108,33 +110,21 @@ impl MyNode {
                 dst,
                 auth,
             } => {
+                debug!("valid client msg");
+                if send_stream.is_none() {
+                    error!("No client response stream!");
```
bochaco, line 115: Shouldn't we just bail out here as all msgs from client shall keep a bi-stream open? if so, presumably it will simplify the functions we call for handling client msg since they will all have a stream arg rather than `Option<..>` arg ?

sn_node/src/node/data/records.rs
bochaco: not sure, but perhaps we can just change the `client_response_stream` arg to not be an `Option` but just a stream?

sn_node/src/node/cfg/config_handler.rs
```diff
@@ -127,8 +127,8 @@ pub struct Config {
     /// declaring us offline. If none is supplied we'll default to the documented constant.
     ///
     /// The interval is in milliseconds. A value of 0 disables this feature.
-    #[clap(long)]
-    pub keep_alive_interval_msec: Option<u32>,
+    #[clap(long, default_value = "0")]
```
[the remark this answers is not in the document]
joshuef: Ah, i was going off of the docs above. This is really just a casualty of me trying some stuff so i'll revert this change

sn_node/src/node/data/records.rs
```diff
@@ -6,62 +6,133 @@
...
+                            bytes_to_adult.clone(),
+                        )
+                        .await
+                    // TODO: parse response and fail if err msg
+                }
+                .boxed(),
+            );
+        }
+
+        let (response, _what) = futures::future::select_ok(send_tasks).await?;
```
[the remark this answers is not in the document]
joshuef: Hmm, so docs read like it polls until completion on all of them: 

> this will only return the first successful completion, or the last failure

----
But aye, this makes sense:

> ....and if it failed we didn't send it to any of the Adults so we probably need to report it back to the client....?

----

So it seems like 1) report to client should be easy enough if it all fails. But how do we handle partial success? I'd kind of think that becomes an elder issue for resending/dysfunction/retrying the store, with the same adult or another? But for the client one success is enough? Perhaps no though 🤔

sn_node/src/node/messaging/mod.rs
joshuef: So I did start down that road, but hit a blocker w/r/t the healthchecks as we currently have them. @b-zee caught this as well.. So I'm not sure if best to a) disable healthchecks atm or b) refactor this problem away from healthchecks...

sn_node/src/node/data/records.rs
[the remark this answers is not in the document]
bochaco: > it polls until completion on all of them

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a15064240e36574641cb68861d8cc842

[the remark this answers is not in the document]
joshuef: Okay replacing this with `join_all` and just going to respond after that if all ok.  Anything else should highlight a bug atm

sn_client/src/connections/listeners.rs
```diff
@@ -200,31 +205,17 @@ impl Session {
...
-                    Expecting an AE-Retry next"
-                );
-                self.send_msg(vec![elder], wire_msg, msg_id, false, resp_tx)
-                    .await?;
-            } else {
-                error!("No elder determined for resending AE message");
-            }
+            debug!("Resending original message to {src_peer:?} to new section elders");
+
+            self.send_msg(elders, wire_msg, msg_id, false, resp_tx)
```
bochaco, line 217: Shouldn't we be re-sending it only to the Elder which replied with AE msg? each of the msgs we sent to N Elders would potentially return a AE response, so for each of them we shall retry but only to the corresponding Elder....?

bochaco: ... actually, it seems we may need to treat AE-Retry different from AE-Redirect, I believe my comment/Q above would be applicable only to AE-Retry response cases. For AE-Redirect we may be ok with sending it to all but when counting/aggregating responses we may not be considering it properly...?....e.g. same query response from same Elder would be counted as coming from different Elders which would affect how we compare it with super-majority number....

sn_client/src/connections/messaging.rs
```diff
@@ -589,90 +587,72 @@ impl Session {
...
+                                msg_id,
+                                peer,
+                                recv_stream,
+                                resp_tx_clone,
+                            );
+
+                            Ok(())
+                        }
+                        #[cfg(features = "chaos")]
+                        Err(SendToOneError::ChaosNoConnection) => break Err(Error::ChoasSendFail),
```
bochaco, line 629: The `break` is not needed anymore just the `Err`

sn_interface/src/types/connections/link.rs
```diff
@@ -270,6 +144,8 @@ pub enum SendToOneError {
     #[cfg(feature = "chaos")]
     /// ChaosNoConn
     ChaosNoConnection,
+    /// Sending failed repeatedly
+    SendRepeatedlyFailed,
```
bochaco, line 148: We don't seem to use this error type

sn_client/src/errors.rs
```diff
@@ -193,6 +193,9 @@ pub enum Error {
         /// Number of Chunks retrieved
         retrieved: usize,
     },
+    /// All attempts to initiate bi di streams via Link Connections failed
+    #[error("All attempts to initiate bi di streams via Link Connections failed: {0:?}")]
```
bochaco, line 197: The message seems to imply we've retried at least once but we are not retrying but failing (for a peer) upon first failure?

sn_node/src/node/data/records.rs
```diff
@@ -6,62 +6,196 @@
...
+                    client_response_stream.clone(),
+                    traceroute.clone(),
+                )
+                .await?;
+            } else {
+                // This should not be possible with above checks
+                error!("No valid ack response to send from all responses for {msg_id:?}")
+            }
+        } else {
+            error!("Storage was not completely successful for {msg_id:?}");
```
bochaco, line 144: Shouldn't we be sending errors back to the client (apart from any dysfunctional actions we'll be doing) so it can retry as necessary if the data was not stored, e.g. the error could have been due to lack of data mutation permissions....?

sn_client/src/connections/listeners.rs
joshuef: Hmmm, aye. We _could_. I'm not sure if "next closest" for nodes in a section might yield the same answer for two different initial target elders though? In which case we'd not see the desired number of responses in the end 🤔 

> when counting/aggregating responses we may not be considering it properly...?

We also check that it's not from an elder that's already sent us something. So I think we're okay there?

> e.g. same Query response from same new Elder (coming more than once) would be counted as coming from different Elders

We aggregate based on source socketaddr just now. So if that changes that would be problematic yeh.  We could perhaps check that the ack sender is part of our knowledge _right now_ when we do final aggregation?

bochaco: > based on source socketaddr just now
I don't think we do for queries in `Session::check_query_responses`, we only count them. But even if we adapted that to consider src addr as we do for Cmd responses/Acks, we will still be re-sending the bounced msg several times to the same new Elder I believe....

joshuef: Aye you're right. I've updated this to just use an arbitrary xorname for sorting. As long as that's consistent across the client reponse handling, we'll be mapping index->index in the elders provided. That should work whether it's the same prefix or another one (I think). 

See: https://github.com/maidsafe/safe_network/pull/1730/commits/5b6b42f9db869622de9400fd733fbcbd0c52f19f

bochaco (approved):

PR #1732, opened 2022-11-04
refactor(tests): remove some `?` noise in tests

davidrusu:
This PR makes a few test helpers panic instead of returning Result.

All the helpers changed would never return Err if the tests are written correctly, so a panic here is justified I think.

i.e. `prefix(&str)` takes a string of 0's and 1's, and returns a `Prefix`. passing anything other than 0's and 1's, i.e. `prefix("abc")` is a bad test ==> panic early.

This reduces some symbol noise in the tests, some tests are more affected than others but I think it generally reduces load when reading through.

### DIFF
--- a/sn_interface/src/network_knowledge/test_utils.rs
+++ b/sn_interface/src/network_knowledge/test_utils.rs
@@ -15,9 +15,8 @@ use std::{cell::Cell, collections::BTreeMap, fmt, net::SocketAddr};
 use xor_name::Prefix;
 
 // Parse `Prefix` from string
-pub fn prefix(s: &str) -> Result<Prefix> {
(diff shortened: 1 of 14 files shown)


sn_interface/src/network_knowledge/test_utils.rs
```diff
@@ -177,17 +183,15 @@ pub fn section_decision<P: Proposition>(
     let decision = nodes[0]
```
RolandSherwin, line 183: There seems to be a clippy warning here, the `let` binding can be removed..

RolandSherwin (approved):

b-zee (approved):

PR #1733, opened 2022-11-07
fix(sn_node): only proposal when reached handover consensus first time

maqi:


maqi:
Closed as to raise as new PR agaisnt the main branch

PR #1738, opened 2022-11-08
Feat remove client response streams

joshuef:

### DIFF
--- a/sn_client/src/connections/messaging.rs
+++ b/sn_client/src/connections/messaging.rs
@@ -6,23 +6,24 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use super::{MsgResponse, QueryResult, Session};
-
+use super::{QueryResult, Session};
 use crate::{Error, Result};
-
+use async_recursion::async_recursion;
+use itertools::Itertools;
+use qp2p::UsrMsgBytes;
+use sn_interface::{
+    messaging::{
+        data::{ClientMsg, DataQuery, DataQueryVariant, QueryResponse},
+        system::{AntiEntropyKind, NodeMsg},
+        ClientAuth, Dst, MsgId, MsgKind, MsgType, WireMsg,
+    },
+    network_knowledge::{supermajority, SectionTreeUpdate},
+    types::{log_markers::LogMarker, ChunkAddress, Peer},
+};
 #[cfg(feature = "traceroute")]
 use sn_interface::{
     messaging::{Entity, Traceroute},
     types::PublicKey,
-};
-
-use sn_interface::{
-    messaging::{
-        data::{DataQuery, DataQueryVariant, QueryResponse},
-        ClientAuth, Dst, MsgId, MsgKind, WireMsg,
-    },
-    network_knowledge::supermajority,
-    types::{ChunkAddress, Peer},
 };
 
 use backoff::{backoff::Backoff, ExponentialBackoff};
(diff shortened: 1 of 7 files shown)


sn_client/src/connections/messaging.rs
```diff
@@ -40,14 +41,17 @@ pub(crate) const NUM_OF_ELDERS_SUBSET_FOR_QUERIES: usize = 3;
...
+
 impl Session {
     #[instrument(
         skip(self, auth, payload, client_pk),
         level = "debug",
         name = "session send cmd"
     )]
     pub(crate) async fn send_cmd(
-        &self,
+        &mut self,
```
bochaco, line 54: It sounds to me that you can remove the mutability requirement from `Session::update_network_knowledge` which will allow you to remove all the `Session` mutability requirements from all functions including this one, which is the one causing the large amount of changes in this PR?

joshuef: you were super correct. massively simplified things. pr updated

sn_client/src/connections/messaging.rs
```diff
@@ -105,52 +104,64 @@ impl Session {
         &self,
         msg_id: MsgId,
         elders: Vec<Peer>,
-        mut resp_rx: mpsc::Receiver<MsgResponse>,
+        responses: Vec<(Peer, MsgType)>, // mut resp_rx: mpsc::Receiver<MsgResponse>,
```
bochaco, line 107: we can remove the comment

sn_client/src/connections/messaging.rs
```diff
@@ -105,52 +104,64 @@ impl Session {
...
-                    if received_acks.len() >= expected_acks {
-                        trace!("Good! We've at or above {expected_acks} expected_acks");
-                        return Ok(());
+        for (peer, msg_response) in responses {
+            let src = peer.addr();
+            match msg_response {
+                MsgType::Client {
+                    msg_id,
+                    auth: _,
+                    dst: _,
```
bochaco, line 120: you can use just `..` to ignore all these unused fields

bochaco (commented):
Apart from the cosmetic comments below, I see one difference with current implementation which is that we are not short-circuiting when, for example, we received already enough errors for a msg, we wait for all responses regardless of their result to then finally decide the outcome. We can work on it separatelly though, playing around with streams perhaps we can achieve that.

Edit: another aspect which would be nice to avoid if possible, not sure how to resolve yet, it's the introduction of an `async_recursion` to handle the AE msgs

sn_client/src/connections/messaging.rs
```diff
@@ -82,20 +86,15 @@ impl Session {
         // Don't immediately fail if sending to one elder fails. This could prevent further sends
         // and further responses coming in...
         // Failing directly here could cause us to miss a send success
```
bochaco, line 88: Aren't the above comments outdated with this change?

sn_client/src/connections/messaging.rs
```diff
@@ -582,65 +607,112 @@ impl Session {
...
-                if let Err(err) = &result {
-                    session.peer_links.remove_link_from_peer_links(peer).await;
-                    warn!("Issue when sending {msg_id:?} to {peer:?}: {err:?}");
+                    // TODO: ???? once we drop the stream, do we know the connection is closed ???
+                    trace!("{} to {}", LogMarker::StreamClosed, peer.addr());
+
+                    res
+                }
+                #[cfg(features = "chaos")]
+                Err(SendToOneError::ChaosNoConnection) => break Err(Error::ChoasSendFail),
```
bochaco, line 668: No need for the `break`

joshuef:
looks like #1746 may take over this PRs aims

PR #1742, opened 2022-11-09
Ci improve flow

joshuef:

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1,0 +1,667 @@
+# Run all tests on all platforms.
+name: Merge Flow. Run All Tests
+
+on:
+  push:
+    branches: [staging, trying]
+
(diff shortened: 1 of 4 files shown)


.github/workflows/merge.yml
```diff
@@ -25,7 +24,7 @@ jobs:
     # github jobs that need to have passed for bors to give the all clear
     needs:
       - cargo-udeps
-      - cargo-deny
+      # - cargo-deny
```
jacderida, line 27: Just one question. Should we delete this instead of commenting out? Or do we plan on introducing it later?

joshuef: I'd say we want it for later. Right now it's a dep of a dep preventing this being useful

jacderida (approved):

PR #1743, opened 2022-11-09
Chore break up replication locks

joshuef:

### DIFF
--- a/sn_node/src/node/messaging/mod.rs
+++ b/sn_node/src/node/messaging/mod.rs
@@ -28,7 +28,7 @@ use sn_interface::{
 };
 
 use std::{collections::BTreeSet, sync::Arc};
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, RwLock};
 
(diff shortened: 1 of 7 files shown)


sn_node/src/node/messaging/mod.rs
```diff
@@ -78,10 +81,11 @@ impl MyNode {
 
         match msg_type {
             MsgType::Node { msg_id, dst, msg } => {
+                let read_locked_node = node.read().await;
                 // Check for entropy before we proceed further
                 // Anythign returned here means there's an issue and we should
                 // short-circuit below
-                let ae_cmds = self
+                let ae_cmds = read_locked_node
```
b-zee, line 88:
```suggestion
                let ae_cmds = node.read().await
```

This doesn't make a difference currently, but in the future it might prevent keeping the lock just that little bit longer than needed.

b-zee (approved):
Looks a bit more messy in general, but looks like the right thing is done.

PR #1745, opened 2022-11-09
refactor: remove traceroute feature

RolandSherwin:


sn_client/src/connections/messaging.rs
```diff
@@ -234,9 +217,6 @@ impl Session {
         #[allow(unused_mut)]
         let mut wire_msg = WireMsg::new_msg(msg_id, payload, kind, dst);
```
b-zee, line 219:
```suggestion
        let wire_msg = WireMsg::new_msg(msg_id, payload, kind, dst);
```

b-zee (changes requested):
Good stuff, unfortunate that we couldn't use this very much as we thought we might...

sn_node/src/node/data/records.rs
```diff
@@ -387,14 +340,6 @@ impl MyNode {
         #[allow(unused_mut)]
```
b-zee, line 340: Same applies here.

RolandSherwin: The other feature gated code block which calls `set_payload_debug()` requires `wire_msg` to be mutable..

b-zee: Aah, sorry, that was meant for another piece in the same file, don't know what happened. Solved!

b-zee (approved):

davidrusu (approved):

PR #1747, opened 2022-11-10
tests: DKG

RolandSherwin:
- https://github.com/maidsafe/safe_network/commit/fed9eed1b4459a5c7f6e6ba97b31f678cae2d5db Simulate a DKG round from start till termination and verify the newly generated `SecretKeyShare` for each node. The test bypasses the comm module and passes the `NodeMsgs` directly to the corresponding peers.
- https://github.com/maidsafe/safe_network/commit/6868415191ae8bed30b42da31f90b9775db47b09 After DKG termination, each node proposes a new `SectionInfo` to denote the change in the `SectionKey`. But if a lagging node (DKG in-progress) has received the `SectionAuthorityProvider` from a terminated node through AE, then the lagging node should not propose `SectionInfo`.
- https://github.com/maidsafe/safe_network/commit/e81b205339275a8f18122ecf97bdeabbe55403bc DKG requires all the votes from the participants to progress
- https://github.com/maidsafe/safe_network/commit/94805d0c511cd5542004721c332e82c1c3343686 Simulate nodes dropping some `NodeMsgs` directed at them. This will cause the DKG to stall since they missed some votes. Send DKG gossip from a random node to the participants whenever the DKG is stalled. This should allow the participant to catchup and terminated.

### DIFF
--- a/sn_node/src/node/messaging/dkg.rs
+++ b/sn_node/src/node/messaging/dkg.rs
@@ -819,7 +819,7 @@ mod tests {
         collections::{BTreeMap, BTreeSet},
         sync::Arc,
     };
-    use tokio::sync::RwLock;
+    use tokio::sync::{mpsc, RwLock};
     use xor_name::{Prefix, XorName};
 
     /// Simulate an entire round of dkg till termination; The dkg round creates a new keyshare set
@@ -876,7 +876,7 @@ mod tests {
                                     .mock_dkg_outcome_proposal(section_auth, outcome)
                                     .await?;
                                 assert_matches!(msg, NodeMsg::Propose { proposal, .. } => {
-                                    assert_matches!(proposal, Proposal::SectionInfo(_))
+                                    assert_matches!(proposal, Proposal::RequestHandover(_))
                                 });
                             }
                             _ => panic!("got a different cmd {:?}", cmd),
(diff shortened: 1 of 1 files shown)


sn_node/src/node/messaging/dkg.rs
```diff
@@ -786,3 +786,715 @@ impl MyNode {
...
+
+    /// Simulate an entire round of dkg till termination; The dkg round creates a new keyshare set
+    /// without any elder change (i.e., the dkg is between the same set of elders). The test
+    /// collects the `NodeMsg`s and passes them to the recipient nodes directly instead of using the
+    /// comm module.
+    #[tokio::test]
+    async fn simulate_dkg_round() -> Result<()> {
+        init_logger();
+        // Construct a local task set that can run `!Send` futures.
+        let loc = tokio::task::LocalSet::new();
```
b-zee, line 833: Everything is executed within a task set, but I don't see any `spawn_local` being used. This applies to the other tests as well.

RolandSherwin: Yeah! You're right! But I do remember seeing a `spawn_local` somewhere that made me use the localset, I assume it must've been removed during the latest changes.  Anyways, fixed it! Thank you!

b-zee (approved):

PR #1748, opened 2022-11-10
fix(sn_node): only proposal when reached handover consensus first time

maqi:

### DIFF
--- a/sn_node/src/node/messaging/agreement.rs
+++ b/sn_node/src/node/messaging/agreement.rs
@@ -23,22 +23,24 @@ impl MyNode {
         &mut self,
         proposal: Proposal,
         sig: SectionSig,
-    ) -> Result<Option<Cmd>> {
+    ) -> Result<Vec<Cmd>> {
         debug!("{:?} {:?}", LogMarker::ProposalAgreed, proposal);
+        let mut cmds = Vec::new();
         match proposal {
             Proposal::VoteNodeOffline(node_state) => {
-                Ok(self.handle_offline_agreement(node_state, sig))
+                cmds.extend(self.handle_offline_agreement(node_state, sig))
             }
-            Proposal::SectionInfo(sap) => self.handle_section_info_agreement(sap, sig),
+            Proposal::SectionInfo(sap) => {
+                cmds.extend(self.handle_section_info_agreement(sap, sig)?)
+            }
             Proposal::NewElders(_) => {
                 error!("Elders agreement should be handled in a separate blocking fashion");
-                Ok(None)
             }
             Proposal::JoinsAllowed(joins_allowed) => {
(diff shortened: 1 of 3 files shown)


sn_node/src/node/messaging/agreement.rs
```diff
@@ -23,22 +23,26 @@ impl MyNode {
...
-    ) -> Result<Option<Cmd>> {
+    ) -> Result<Vec<Cmd>> {
         debug!("{:?} {:?}", LogMarker::ProposalAgreed, proposal);
+        let mut cmds = Vec::new();
         match proposal {
             Proposal::VoteNodeOffline(node_state) => {
-                Ok(self.handle_offline_agreement(node_state, sig))
+                if let Some(cmd) = self.handle_offline_agreement(node_state, sig) {
+                    cmds.push(cmd);
+                }
```
davidrusu, line 33: I think you can write this as an extend as well:
```suggestion
                cmds.extend(self.handle_offline_agreement(node_state, sig))
```

sn_node/src/node/messaging/handover.rs
```diff
@@ -28,20 +28,29 @@ impl MyNode {
     pub(crate) fn propose_handover_consensus(
         &mut self,
         sap_candidates: SapCandidate,
-    ) -> Result<Option<Cmd>> {
+    ) -> Result<Vec<Cmd>> {
+        let mut cmds = Vec::new();
         match &self.handover_voting {
```
davidrusu, line 33: A bit unrelated, but the clones here seem a bit wasteful. Do you know if this would work?

```suggestion
        match &mut self.handover_voting {
```

This way we can mutate the handover state directly instead cloning and writing back the updated state.

sn_node/src/node/messaging/handover.rs
```diff
@@ -290,14 +300,18 @@ impl MyNode {
...
-                        "{}: {:?}",
-                        LogMarker::HandoverConsensusTermination,
-                        candidates_sap
-                    );
+                    // The Termination & Decision Broadcasting shall only undertaken
+                    // on the first time the consensus reached.
+                    if !had_consensus_value {
+                        debug!(
+                            "{}: {:?}",
+                            LogMarker::HandoverConsensusTermination,
```
davidrusu, line 308: Do you want to add this LogMarker to the early termination case above as well?

sn_node/src/node/messaging/handover.rs
maqi: we cann't do it as within the match clause, the `self.broadcast_handover_vote_msg` and `self.broadcast_handover_decision` also requires mutable access to `self` which will get `locked` if we do the `as_mut` call.

davidrusu (approved):

PR #1749, opened 2022-11-10
chore(node): Pass around MyNodeState to avoid holding locks

joshuef:
For longer running message handling, we now pass around the inital MyNodeState. This avoids a tonnnn of read locks and therefore hopefully prevents holding up write and reads needlessly.

### DIFF
--- a/.github/workflows/run_pr_checks.yml
+++ b/.github/workflows/run_pr_checks.yml
@@ -191,7 +191,7 @@ jobs:
 
       - name: Run client tests
         env:
-          RUST_LOG: "sn_client=trace,qp2p=debug"
+          RUST_LOG: "sn_client,sn_interface"
         run: cd sn_client && cargo test --release --features check-replicas -- --test-threads=1
         timeout-minutes: 50
 
@@ -454,7 +454,7 @@ jobs:
         run: ./target/release/testnet --interval 30000
         id: section-startup
         env:
-          RUST_LOG: "sn_node,sn_consensus,sn_dysfunction=trace,sn_interface=trace"
+          RUST_LOG: "sn_node,sn_consensus,sn_dysfunction,sn_interface"
 
       - name: Wait for all nodes to join
         shell: bash
@@ -467,7 +467,7 @@ jobs:
 
       - name: Run API tests
         env:
-          RUST_LOG: "sn_client=trace"
+          RUST_LOG: "sn_client,sn_interface"
         run: cd sn_api && cargo test --release --features check-replicas -- --test-threads=1
         timeout-minutes: 30
 
@@ -558,7 +558,7 @@ jobs:
         run: ./target/release/testnet --interval 30000
         id: section-startup
         env:
-          RUST_LOG: "sn_node,sn_consensus,sn_dysfunction=trace,sn_interface=trace"
+          RUST_LOG: "sn_node,sn_consensus,sn_dysfunction,sn_interface"
 
       - name: Wait for all nodes to join
         shell: bash
--- a/sn_node/src/node/api.rs
+++ b/sn_node/src/node/api.rs
@@ -32,6 +32,8 @@ use sn_dbc::{
 use std::{path::PathBuf, sync::Arc};
 use xor_name::XorName;
 
+use super::core::MyNodeSnapshot;
+
 impl MyNode {
     pub(crate) async fn first_node(
         comm: Comm,
@@ -103,8 +105,12 @@ impl MyNode {
     }
 
     /// Returns the SAP of the section matching the name.
-    pub(crate) fn matching_section(&self, name: &XorName) -> Result<SectionAuthorityProvider> {
-        self.network_knowledge
+    pub(crate) fn matching_section(
+        snapshot: &MyNodeSnapshot,
+        name: &XorName,
+    ) -> Result<SectionAuthorityProvider> {
+        snapshot
+            .network_knowledge
             .section_auth_by_name(name)
             .map_err(Error::from)
     }
--- a/sn_node/src/node/api.rs
+++ b/sn_node/src/node/api.rs
@@ -17,8 +17,7 @@ use crate::{
 
 use sn_interface::{
     network_knowledge::{
-        MyNodeInfo, NetworkKnowledge, SectionAuthorityProvider, SectionKeyShare, SectionsDAG,
-        GENESIS_DBC_SK,
+        MyNodeInfo, NetworkKnowledge, SectionKeyShare, SectionsDAG, GENESIS_DBC_SK,
     },
     types::log_markers::LogMarker,
 };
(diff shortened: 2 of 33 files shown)


joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

sn_node/src/node/mod.rs
```diff
@@ -148,7 +149,47 @@ mod core {
...
+        pub(crate) event_sender: EventSender,
+        pub(crate) joins_allowed: bool,
+        pub(crate) membership: Option<Membership>,
+    }
+
     impl MyNode {
+        /// Generate a read only snapshot of a subset of the node's state
+        /// Useful for longer running processes to avoid having to acquire
+        /// read locks eg
+        pub(crate) fn get_snapshot(&self) -> MyNodeSnapshot {
```
b-zee, line 174: Rust convention is to not use `get_` prefixes.

sn_node/src/node/api.rs
```diff
@@ -103,8 +105,12 @@ impl MyNode {
     }
 
     /// Returns the SAP of the section matching the name.
-    pub(crate) fn matching_section(&self, name: &XorName) -> Result<SectionAuthorityProvider> {
-        self.network_knowledge
+    pub(crate) fn matching_section(
```
b-zee, line 108: Perhaps better to move this to the `impl MyNodeSnapshot`?

b-zee (commented):
Generally I have mixed feelings as it doesn't help readability of code. It seems like a sledge hammer approach, because I assume it solves problems, but it's not exactly sure where the exact bottleneck was.
Overall I think it's an improvement for performance, but I can't verify.

joshuef:
Generally what I was seeing was some longer than desireable waits on `locks`, that could be `read` or `write` locks. 

So yeh, this is a sledgehammer to reduce that for sure. In general I think it clears up some misplaced mutability, and removes potential for deadlocks if we're holding them for longer than desired. (And we had a bunch over `await` calls).

I think  off the back of this we could look to pull out a bunch of functionality from the node into some more testable functions eg. (everything that is now `MyNode::<X>`. Which could take distinct arguments as opposed to the whole state. (That was much simpler for now though).

Here I feel it's also simpler to reason about the code and where we do need to mutate some state. It's not as often as you think and removing the `self` code I think helps clarify that...

joshuef:
bors try

sn_node/src/node/mod.rs
```diff
@@ -254,30 +309,30 @@ mod core {
         ////////////////////////////////////////////////////////////////////////////
 
         /// Generates a random AE probe for _anywhere_ on the network.
-        pub(crate) fn generate_probe_msg(&self) -> Result<Cmd> {
+        pub(crate) fn generate_probe_msg(snapshot: &MyNodeSnapshot) -> Result<Cmd> {
```
bochaco, line 312: It sounds like all APIs like this one which work on a snapshot shall now be part of the `MyNodeSnapshot` impl instead?

sn_node/src/node/mod.rs
```diff
@@ -723,12 +777,12 @@ mod core {
...
-    #[derive(Clone)]
-    pub(crate) struct StateSnapshot {
-        pub(crate) is_elder: bool,
-        pub(crate) section_key: bls::PublicKey,
-        pub(crate) prefix: Prefix,
-        pub(crate) elders: BTreeSet<XorName>,
-        pub(crate) members: BTreeSet<XorName>,
-    }
+    // #[derive(Clone)]
+    // pub(crate) struct StateSnapshot {
```
bochaco, line 781: I guess we can remove this struct

.github/workflows/run_pr_checks.yml
```diff
@@ -191,7 +191,7 @@ jobs:
 
       - name: Run client tests
         env:
-          RUST_LOG: "sn_client=trace,qp2p=debug"
+          RUST_LOG: "sn_client,sn_interface"
```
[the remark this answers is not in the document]
joshuef: its not needed, this'll just default to _all_ logs from that lib.

sn_node/src/node/mod.rs
joshuef: Aye they could be. I wasn't sure where to draw the line atm. I think there's a lot more refactoring we could do atop this... but maybe better for follow up PRs?

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -100,20 +104,35 @@ impl Dispatcher {
...
-                // drop the write lock
-                drop(node);
-
-                let node = self.node.read().await;
-
-                debug!("Network knowledge was updated: {updated}");
-                node.handle_valid_client_msg(msg_id, msg, auth, origin, send_stream)
+                let mut snapshot = self.node.read().await.snapshot();
+                // block off the write lock to ensure it's dropped
+                {
```
bochaco, line 109: Isn't the scope provided by `if there_was_an_update {` below enough for dropping the lock ASAP ?

sn_node/src/node/mod.rs
bochaco: I see, yeah, makes sense

sn_node/src/node/flow_ctrl/dispatcher.rs
joshuef: the read lock? 

That should be released at the end of the line I believe: `let mut snapshot = self.node.read().await.snapshot();` , (The snapshot itself is pure cloning, so mutability here doesnt actually affect `MyNode` or need a lock).

Or do you mean something else?

joshuef: ahhhhh with you. yeh, you're totally right! That's a hangover from an earlier changeset I think. I'll clear it up.

sn_node/src/node/mod.rs
```diff
@@ -148,7 +148,59 @@ mod core {
         pub(crate) dysfunction_tracking: DysfunctionDetection,
     }
 
+    #[derive(Clone)]
+    pub(crate) struct MyNodeSnapshot {
+        pub(crate) root_storage_dir: PathBuf,
+        pub(crate) is_elder: bool,
+        pub(crate) is_not_elder: bool,
```
bochaco, line 155: I think we shall remove either `is_elder` or `is_not_elder`

bochaco (commented):
As a general comment, perhaps we shall use a diff name than `MyNodeSnapshot`/`snapshot`, I'm thinking perhaps something like `context` as it's the pattern we are actually implementing here...?

joshuef:
bors try

bochaco (approved):

PR #1750, opened 2022-11-11
refactor(tests): remove `tokio::LocalSet`

RolandSherwin:
- https://github.com/maidsafe/safe_network/pull/1750/commits/502a9769cdd786b2b5b2c3c559854355d2b64444  The `Comm` module does not use `spawn_local` anymore, hence we do not need to use a local task set for the tests


joshuef (approved):
Good tidy up

davidrusu (approved):
Nice

PR #1755, opened 2022-11-15
chore(test_utils): use `TestSapBuilder`

RolandSherwin:
- https://github.com/maidsafe/safe_network/pull/1755/commits/cd0b27a3aed7bb10a7bb1ab933bea208fb17b432 feat(test_utils): introduce `TestSapBuilder` to generate custom SAPs
Replaces the standalone utils into a single one to generate random
SAPs for testing.
- https://github.com/maidsafe/safe_network/pull/1755/commits/08054dbdb4d34cb7fc045189316d322f02d1c629 chore(test_utils): replace `TestSAP` with `TestSapBuilder`

### DIFF
--- a/sn_interface/src/network_knowledge/section_authority_provider.rs
+++ b/sn_interface/src/network_knowledge/section_authority_provider.rs
@@ -225,105 +225,186 @@ impl SectionAuthorityProvider {
 #[cfg(any(test, feature = "test-utils"))]
 pub mod test_utils {
     use crate::{
-        messaging::system::SectionSigned,
-        network_knowledge::{MyNodeInfo, NodeState, SectionAuthorityProvider},
-        test_utils::{gen_sorted_nodes, TestKeys},
+        elder_count,
+        network_knowledge::{supermajority, MyNodeInfo, NodeState, SectionAuthorityProvider},
+        test_utils::gen_sorted_nodes,
     };
-    use rand::RngCore;
+    use rand::{thread_rng, RngCore};
     use xor_name::Prefix;
 
-    /// `SectionAuthorityProvider` related utils for testing
-    pub struct TestSAP {}
+    /// Generate sk_set with the provided threshold, else use sup(elder_count)-1. It's because,
+    /// generally we need sup(elder_count) shares to be valid. Thus threshold should be 1
+    /// less than that
+    pub fn gen_sk_set(
+        mut rng: impl RngCore,
+        elder_count: usize,
+        sk_threshold_size: Option<usize>,
+    ) -> bls::SecretKeySet {
+        bls::SecretKeySet::random(
+            sk_threshold_size.unwrap_or_else(|| supermajority(elder_count).saturating_sub(1)),
+            &mut rng,
+        )
+    }
 
-    impl TestSAP {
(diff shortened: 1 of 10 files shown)


sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -325,5 +350,18 @@ pub mod test_utils {
 
             (signed_sap, nodes, secret_key_set.secret_key())
         }
+
+        pub fn _gen_sk_set<R: RngCore>(
```
davidrusu, line 354: How about this instead?
```suggestion
        pub fn gen_sk_set_with_threshold<R: RngCore>(
```

sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -307,6 +306,32 @@ pub mod test_utils {
             (section_auth, nodes)
         }
 
+        /// Generate a `SectionAuthorityProvider` for testing with the provided set of members.
+        pub fn random_sap_with_members<R: RngCore>(
+            rng: &mut R,
```
davidrusu, line 311: `RngCore` is implemented for `&mut RngCore`. The `rand` people did this so that we don't have to request mutable references to the rng's:

So you can rewrite this as:
```suggestion
            mut rng: impl RngCore,
```

Any function that takes a &mut RngCore can be rewritten to this

sn_interface/src/network_knowledge/section_authority_provider.rs
davidrusu, line 354: The `_` here is throwing me off, It seems like you want to use this helper in tests so it's not intended to be private. 

How about just `gen_sk_set`
```suggestion
        pub fn gen_sk_set<R: RngCore>(
```

sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -325,5 +350,18 @@ pub mod test_utils {
 
             (signed_sap, nodes, secret_key_set.secret_key())
         }
+
+        pub fn _gen_sk_set<R: RngCore>(
+            rng: &mut R,
```
davidrusu, line 355:
```suggestion
            mut rng: impl RngCore,
```

and in the body, you can take a mutable reference where needed (i.e. &mut rng).

Later on we want to switch `blsttc` to use this `rng: impl RngCore` style API as well.

sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -325,5 +350,18 @@ pub mod test_utils {
...
+            elder_count: usize,
+            sk_threshold_size: Option<usize>,
+        ) -> bls::SecretKeySet {
+            // by default we need sup(elder_count) shares to be valid. Thus threshold should be 1
+            // less than that
+            bls::SecretKeySet::random(
+                sk_threshold_size.unwrap_or(supermajority(elder_count) - 1),
```
davidrusu, line 362: I think this may underflow when `elder_count` is low. Use a `saturating_sub` here instead. 
```suggestion
                sk_threshold_size.unwrap_or(supermajority(elder_count).saturating_sub(1)),
```

More generally, I always prefer to do saturating subs when working with unsigned ints.

davidrusu (approved):

PR #1766, opened 2022-11-18
refactor(responses): return correct cmd response

oetyng:
- Returns the correct response type corresponding to the cmd.
- Adds the data address in question to the ack.
- Renames `ClientMsgResponse` to `ClientDataResponse`.
- Adds `NodeDataResponse` and aligns it to the pattern of `ClientDataResponse`.
- Moves data write acks to `NodeDataReponse`.
- Makes `NodeEvent` only be Adult to Elder notifications.
- Removes the `NodeQueryResponse` dead code path from `NodeMsg`.
- Removes some unnecessary derive macros from `QueryResponse`, `CmdResponse` and `sn_interface::messaging::data::Error`,

### DIFF
--- a/sn_client/src/connections/listeners.rs
+++ b/sn_client/src/connections/listeners.rs
@@ -62,10 +62,10 @@ impl Session {
             if attempt > MAX_AE_RETRIES_TO_ATTEMPT {
                 break MsgResponse::Failure(
                     addr,
-                    Box::new(Error::AntiEntropyMaxRetries {
+                    Error::AntiEntropyMaxRetries {
                         msg_id: correlation_id,
                         retries: attempt - 1,
-                    }),
+                    },
                 );
             }
 
@@ -90,17 +90,17 @@ impl Session {
                         attempt += 1;
                         continue;
                     }
-                    Err(err) => break MsgResponse::Failure(addr, Box::new(err)),
+                    Err(err) => break MsgResponse::Failure(addr, err),
                 },
                 Ok(msg @ MsgType::Client { .. }) => {
                     warn!("Unexpected ClientMsg type received for {correlation_id:?}: {msg:?}");
                     break MsgResponse::Failure(
                         addr,
-                        Box::new(Error::UnexpectedMsgType {
+                        Error::UnexpectedMsgType {
                             correlation_id,
                             peer,
                             msg,
-                        }),
+                        },
                     );
                 }
                 Ok(msg @ MsgType::NodeDataResponse { .. }) => {
@@ -109,14 +109,14 @@ impl Session {
                     );
                     break MsgResponse::Failure(
                         addr,
-                        Box::new(Error::UnexpectedMsgType {
+                        Error::UnexpectedMsgType {
                             correlation_id,
                             peer,
                             msg,
-                        }),
+                        },
                     );
                 }
-                Err(err) => break MsgResponse::Failure(addr, Box::new(err)),
+                Err(err) => break MsgResponse::Failure(addr, err),
             }
         };
 
--- a/sn_client/src/connections/listeners.rs
+++ b/sn_client/src/connections/listeners.rs
@@ -13,7 +13,7 @@ use crate::{Error, Result};
 use qp2p::{RecvStream, UsrMsgBytes};
 use sn_interface::{
     messaging::{
-        data::{ClientMsg, ClientMsgResponse},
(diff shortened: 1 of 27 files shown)


sn_client/src/connections/listeners.rs
```diff
@@ -62,18 +62,18 @@ impl Session {
             if attempt > MAX_AE_RETRIES_TO_ATTEMPT {
                 break MsgResponse::Failure(
                     addr,
-                    Error::AntiEntropyMaxRetries {
+                    Box::new(Error::AntiEntropyMaxRetries {
```
joshuef, line 65: How come we need a `Box` now?

sn_interface/src/messaging/data/mod.rs
```diff
@@ -205,35 +208,60 @@ impl QueryResponse {
...
     //
     // ===== Spentbook Data =====
     //
     /// Response to SpentbookCmd::Spend.
-    SpendKey(Result<()>),
+    SpendKey(Result<DataAddress>),
 }
 
 impl CmdResponse {
+    #[allow(clippy::result_large_err)]
```
joshuef, line 233: What's making this large error here? 

(i'm concious we just reworked some things to avoid this error, so if we can avoid it, that'd be suuuper)

sn_interface/src/types/errors.rs
```diff
@@ -66,6 +66,8 @@ pub enum Error {
     /// BLS key error
     #[error(transparent)]
     BlsError(#[from] BlsError),
+    #[error("The operation is invalid in its context: {0}.")]
+    InvalidOperation(String),
```
joshuef, line 70: Can we add concrete errors for what we need here? 

It's good to avoid these catchall `String` error types as they always seem to get abused a bit later.

joshuef (changes requested):
Can we add concrete types for the errors to avoid the catchall `String`?

sn_client/src/connections/listeners.rs
oetyng: It was a clippy warning.
Maybe showed up before i updated rust. Will check.

sn_interface/src/messaging/data/mod.rs
oetyng: I'm really not sure, it was weird actually for the warning to show up.

sn_interface/src/types/errors.rs
oetyng: Yeah I agree. But we already have it in other errors though.

The concrete error in this case is a bit silly.

When getting the correct error from a ReplicatedData we do this:

```
ReplicatedData::RegisterLog(_) => {
                return Err(Error::InvalidOperation(
                    "RegisterLog is not resulting from a cmd.".to_string(),
                ))
            }
            ReplicatedData::SpentbookLog(_) => {
                return Err(Error::InvalidOperation(
                    "SpentbookLog is not resulting from a cmd.".to_string(),
                ))
            }
```

When producing an error CmdResponse we do this
```
Self::SpentbookLog(_) => Err(Error::InvalidOperation(
                "There is no CmdResponse for the SpentbookLog variant of ReplicatedData."
                    .to_string(),
            )),
            Self::RegisterLog(_) => Err(Error::InvalidOperation(
                "There is no CmdResponse for the RegisterLog variant of ReplicatedData."
                    .to_string(),
            )),
```

(Converting those to specific errors  would be like `CannotGetACmdErrorFromRegisterLog` which seems... meh.

oetyng: These are really unreachable paths that should eventually be designed out, if we can do that in a sane way.

sn_client/src/connections/listeners.rs
oetyng: It's because the Error itself has variants that contain the same large stuff like `QueryResponse`, `NodeMsg` and so on.

oetyng: I just checked now by removing the box. The warning came before but was not there now. So no need for the `Box`!

Will fix 👍

oetyng:
> Can we add concrete types for the errors to avoid the catchall `String`?

Hey, check this https://github.com/maidsafe/safe_network/pull/1766#discussion_r1027283000
What do you think?

joshuef (approved):

joshuef:
okay, we actually have the `large_enum` clippy skipper throughout the code. I didn't realise that, so let's ignore that for this PR.

PR #1767, opened 2022-11-19
Ae lock reduction2

joshuef:
previously we locked on handling _all_ anti entropy messages. now we pass in the `Arc<RwLock<MyNode>>` and only take the write lock if there _would_ be an update.

This should reduce the amount of write locks needed significantly as we have many more anti entropy messages that go out and may do _nothing_ (as only the first received for any message sent out will trigger an update... and we send messages to many nodes at once)

### DIFF
--- a/sn_node/src/node/messaging/anti_entropy.rs
+++ b/sn_node/src/node/messaging/anti_entropy.rs
@@ -157,6 +157,58 @@ impl MyNode {
         }
     }
 
+    /// returns (we_have_a_share_of_this_key, we_should_become_an_elder)
+    fn does_section_tree_update_make_us_an_elder_with_key(
+        context: &NodeContext,
+        section_tree_update: &SectionTreeUpdate,
+    ) -> (bool, bool) {
+        let our_name = context.name;
+        let sap = section_tree_update.signed_sap.clone();
+
+        let we_have_a_share_of_this_key = context
+            .section_keys_provider
+            .key_share(&sap.section_key())
+            .is_ok();
+
+        // check we should be _becoming_ an elder
+        let we_should_become_an_elder = sap.contains_elder(&our_name);
+
+        trace!("we_have_a_share_of_this_key: {we_have_a_share_of_this_key}, we_should_become_an_elder: {we_should_become_an_elder}");
+
+        (we_have_a_share_of_this_key, we_should_become_an_elder)
+    }
+
+    /// Test a context to see if we would update Network Knowledge
+    /// returns
+    ///   Ok(true) if the update had new valid information
+    ///   Ok(false) if the update was valid but did not contain new information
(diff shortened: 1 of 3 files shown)


joshuef:
bors try

joshuef:
bors try

joshuef:
bors try

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -156,9 +156,11 @@ impl MyNode {
                 kind,
             } => {
                 trace!("Handling msg: AE from {sender}: {msg_id:?}");
+                let context = node.read().await.context();
                 // as we've data storage reqs inside here for reorganisation, we have async calls to
                 // the fs
-                MyNode::handle_anti_entropy_msg(node, section_tree_update, kind, sender).await
+                MyNode::handle_anti_entropy_msg(node, &context, section_tree_update, kind, sender)
```
oetyng, line 162: Shouldn't need to pass in a borrow of the context, since it is not used afterwards. Can be consumed.

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -471,14 +473,8 @@ impl MyNode {
...
                 // store data and respond w/ack on the response stream
-                MyNode::store_data_as_adult_and_respond(
-                    &mut context,
-                    data,
-                    send_stream,
-                    sender,
-                    msg_id,
-                )
-                .await
+                MyNode::store_data_as_adult_and_respond(&context, data, send_stream, sender, msg_id)
```
oetyng, line 476: Same here: shouldn't need to pass in a borrow of the context, since it is not used afterwards. Can be consumed.

oetyng (approved):
Nice one!

oetyng (changes requested):
Ach, missed to pick the Request changes option in the first review.

sn_node/src/node/messaging/anti_entropy.rs
```diff
@@ -157,26 +157,74 @@ impl MyNode {
...
+        (we_have_a_share_of_this_key, we_should_become_an_elder)
+    }
+
+    /// Test a context to see if we would update Network Knowledge
+    /// returns
+    ///   Ok(true) if the update had new valid information
+    ///   Ok(false) if the update was valid but did not contain new information
+    ///   Err(_) if the update was invalid
+    ///
+    /// Operates on the context mutably
```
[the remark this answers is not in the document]
joshuef: Ah right yeh. I had id operating on the context before, but figured that could lead to unintended consequences for us down the line. Better to do a clean clone and never actually update the context during these checks.

joshuef:
bors try

oetyng (approved):

PR #1768, opened 2022-11-19
fix(node): ae_msg_from_the_future_is_handled unit test was failing du…

bochaco:
…e to wrong assumption of test utility

- The new SAP created by the test was sometimes removing the node used to perform the final test of the AE msg due to assume the list of Elder nodes returned by `TestSapBuilder::build` was sorted in the same order as the `Peer`s obtained from the SAP generated by same utility function.


oetyng (approved):
Nice find.

PR #1769, opened 2022-11-20
chore(node): Remove locks in data replication flows.

joshuef:
Moves the replication flow to be outwith of node, using channels to gather data and send it in another thread entirely. This should remove another cause of serious node.write().lock usage

### DIFF
--- a/sn_node/src/node/error.rs
+++ b/sn_node/src/node/error.rs
@@ -39,7 +39,7 @@ pub enum Error {
     PeerSessionChannel,
     /// SendChannel error for the data replication flow. This is a critical error and the node no longer functions.
     #[error("Data replication channel could not be sent to. This means the receiver has been dropped, the node can no longer replicate data and must shut down.")]
-    DataReplicationChanel,
+    DataReplicationChannel,
     /// This peer has no connections, and none will be created
     #[error("Peer link has no connections ")]
     NoConnectionsForPeer,
--- a/sn_node/src/node/error.rs
+++ b/sn_node/src/node/error.rs
@@ -37,6 +37,9 @@ pub enum Error {
     /// This Peer SendJob could not be sent. We should remove this peer
     #[error("Peer channel errored")]
     PeerSessionChannel,
+    /// SendChannel error for the data replication flow. This is a critical error and the node no longer functions.
+    #[error("Data replication channel could not be sent to. This means the receiver has been dropped, the node can no longer replicate data and must shut down.")]
+    DataReplicationChanel,
     /// This peer has no connections, and none will be created
     #[error("Peer link has no connections ")]
     NoConnectionsForPeer,
--- a/sn_node/src/node/flow_ctrl/dispatcher.rs
+++ b/sn_node/src/node/flow_ctrl/dispatcher.rs
@@ -11,23 +11,35 @@ use crate::node::{messaging::Peers, Cmd, Error, MyNode, Result};
 use sn_interface::{
     messaging::{system::NodeMsg, Dst, MsgId, MsgKind, WireMsg},
     network_knowledge::{NetworkKnowledge, SectionTreeUpdate},
-    types::Peer,
+    types::{DataAddress, Peer},
 };
 
 use qp2p::UsrMsgBytes;
 
 use bytes::Bytes;
-use std::{collections::BTreeSet, sync::Arc};
-use tokio::sync::RwLock;
+use std::sync::Arc;
+use tokio::sync::{
+    mpsc::{channel, Receiver, Sender},
+    RwLock,
+};
 use xor_name::XorName;
 // Cmd Dispatcher.
 pub(crate) struct Dispatcher {
     node: Arc<RwLock<MyNode>>,
+    data_replication_sender: Sender<(Vec<DataAddress>, Peer)>,
 }
 
 impl Dispatcher {
-    pub(crate) fn new(node: Arc<RwLock<MyNode>>) -> Self {
-        Self { node }
+    /// Creates dispatcher and returns a receiver for enqueing DataAddresses for replication to specific peers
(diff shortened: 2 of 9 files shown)


sn_node/src/node/error.rs
```diff
@@ -37,6 +37,9 @@ pub enum Error {
     /// This Peer SendJob could not be sent. We should remove this peer
     #[error("Peer channel errored")]
     PeerSessionChannel,
+    /// SendChannel error for the data replication flow. This is a critical error and the node no longer functions.
+    #[error("Data replication channel could not be sent to. This means the receiver has been dropped, the node can no longer replicate data and must shut down.")]
+    DataReplicationChanel,
```
oetyng, line 42: Typo-nit: `DataReplicationChanel` => `DataReplicationChannel`

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -11,23 +11,35 @@ use crate::node::{messaging::Peers, Cmd, Error, MyNode, Result};
...
 use bytes::Bytes;
-use std::{collections::BTreeSet, sync::Arc};
-use tokio::sync::RwLock;
+use std::sync::Arc;
+use tokio::sync::{
+    mpsc::{channel, Receiver, Sender},
+    RwLock,
+};
 use xor_name::XorName;
 // Cmd Dispatcher.
```
oetyng, line 26: The tiniest of nits perhaps, but a space between usings and the following code should be better? It's the convention I follow at least.

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -179,23 +191,28 @@ impl Dispatcher {
...
-                            .pending_data_to_replicate_to_peers
-                            .insert(data, peers_set);
-                    };
-                }
+                self.data_replication_sender
+                    .send((data_batch, recipient))
+                    .await
+                    .map_err(|_| Error::DataReplicationChanel)?;
+
+                // // we should queue this
```
oetyng, line 199: Remove commented out code.

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -15,14 +15,13 @@ mod periodic_checks;
...
 use periodic_checks::PeriodicChecksTimestamps;
-
-use crate::comm::MsgFromPeer;
-use crate::node::{flow_ctrl::cmds::Cmd, MyNode, Result};
-
+use sn_interface::messaging::system::{NodeCmd, NodeMsg};
 use sn_interface::types::log_markers::LogMarker;
-
+use sn_interface::types::{DataAddress, Peer};
 use std::sync::Arc;
```
oetyng, line 25: Also super-nit: But the convention I follow with the structure of usings, is to put space between the different levels of proximity to our code base.
I use this order, and when several of any, I put space between.
```
super
crate
sn_
[other crates by maidsafe]
third party
```

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -72,6 +76,56 @@ impl FlowCtrl {
             }
         });
 
+        // start a new thread to kick off data replication
+        let _ = tokio::task::spawn(async move {
+            // on init grab a data storage clone for replication processes
+            let node_data_storage = node_for_data_replication.read().await.data_storage.clone();
```
oetyng, line 82: Can we clone the data storage before ln 79, to avoid doing a read lock within this async task?

So, on line 62 we can do

```
let node_identifier = node.read().await.info().name();
let node_data_storage = node.read().await.data_storage.clone();
```

Also we avoid a node clone by that.

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -72,6 +76,56 @@ impl FlowCtrl {
...
+            // let pending_data_to_replicate_to_peers = BTreeMap::new();
+            // is there a simple way to dedupe common data going to many peers?
+            // is any overhead reduction worth the increased complexity?
+            while let Some((data_addresses, peer)) = data_replication_receiver.recv().await {
```
oetyng, line 86: I feel this while loop can be made simpler.. But not a biggie for this PR. So, just a note.

The deduping if going to many peers depends a bit on what pattern we'll have with new peers. How what will the churn process look like and how likely are multiple peers to need bunch of same data like this?

(Connects a bit to the "hit the road running" node join process I've been thinking about).

oetyng (approved):
Nice. Looks good.

I like that we remove the periodic check and the randomness of the replication instant.
More work will be coming in this area I feel.

oetyng (changes requested):
Ach, missed to pick the `Request changes` option in the first review.

sn_node/src/node/flow_ctrl/mod.rs
joshuef: Ah, i did try that, but then we're doing an await in non-async code. A `block_on` or making the fn async seemed overkill. 

The lock is taken just once at startup, so maybe that's fine?

sn_node/src/node/flow_ctrl/mod.rs
joshuef: Aye, initially I had that w/o the `batch`, but we could get gbs in one msg... Which network wise, should be fine. But it'd be holding a looot in mem, which is probs less fine.

oetyng (approved):

sn_node/src/node/flow_ctrl/mod.rs
oetyng: Yeah, it's what goes in mem that will be an issue.

sn_node/src/node/flow_ctrl/mod.rs
oetyng: Aha, ah I missed that it was non-async code there. Well, I see. Never mind then.

But just to check, holding locks across awaits was something mentioned that could be problematic, right?

PR #1773, opened 2022-11-21
chore(node): reduce write locks for dysfunction checks

joshuef:
Moves the time sensitive cleanup into its own function, which we call over a longer period.

The get_dysfunctional_nodes call can now happen with a node read


joshuef:
@RolandSherwin went and did the real job in #1777 so closing this in favour of that

Issue #1772, opened 2022-11-21
Remove DysfunctionDetection from MyNode.


joshuef:
DysfunctionDetection isn't really a feature of `MyNode` state that needs to be tied to/tracked with membership. 

We could move this up to track in the dispatcher, eg. Or out of both and into its own thread entirely.

RolandSherwin:
I will take up this issue..

PR #1777, opened 2022-11-21
move `DysfunctionDetection` to a separate task

RolandSherwin:
Fixes https://github.com/maidsafe/safe_network/issues/1772

- chore(sn_dysfunction): remove unusued `Result` from methods. 
- chore(sn_dysfunction): remove unusued `Error` module. 
- chore(sn_dysfunction): move `DysfunctionDetection` to a new task
Decouple the Dysfunction module from `MyNode` by moving it to a separate `tokio::task`. Use `mpsc` channels to send `DysCmds` between the main task and the "dysfunction" task. These are then used to execute various "dysfunction" methods. The methods which remove any tracked issues comes under `DysCmds::UntrackIssues` variant. Also use another `mpsc` channel to get back the set of dysfunctional nodes back from the "dysfunction" task.
- chore(sn_dysfunction): rename `IssueType` variants 
The `IssueType` is now used to track as well as untrack dysfunctional nodes. Hence modify the names to be more generic.


joshuef:
Nice one @RolandSherwin ! Good work here 💪 

So you've got all the separation in here, but I think we can push it a bit further (and I think without too much difficulty, I hope). Since we have the channels and the dysf operation, can we pull it out of `MyNode` altogether?

I think the `DysfunctionTracking` could be kicked off at the `FlowCtrl` layer, perhaps? With the `Sender` passed into the node? (At least as a first pass... I know we have these `log_comm_issue` fns that take place where we can't always return Cmds... So it means for now the `Sender` needs to be available for those APIs in `MyNode`.)

----

Down the line, I'd like to verify that our pattern for AeDysfunction etc makes sense... I'm not 100% sure at the moment. But that's not one for this PR

joshuef (changes requested):
Lets pull the whole Dysf out of Mynode, as it doesn't need to be strictly tied to the node state.

sn_node/src/node/flow_ctrl/dysfunction.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub(crate) fn start_dysfunction_detection(
+        node: Arc<RwLock<MyNode>>,
+        mut dys_cmds_from_node: Receiver<DysCmds>,
+    ) -> Receiver<BTreeSet<XorName>> {
+        let (dys_nodes_sender, dys_nodes_receiver) = mpsc::channel(20);
+
+        let _ = tokio::task::spawn(async move {
+            debug!("[NODE READ]: flowctrl start dysfunction detection");
+            let mut dysfunction = DysfunctionDetection::new(
+                node.read()
```
joshuef, line 40: can we pass in context here instead?

sn_node/src/node/flow_ctrl/dysfunction.rs
```diff
@@ -0,0 +1,103 @@
...
+        });
+
+        dys_nodes_receiver
+    }
+
+    /// returns names that are relatively dysfunctional
+    pub(crate) async fn get_dysfunctional_node_names(&mut self) -> BTreeSet<XorName> {
+        // send a DysCmd asking for the dysfunctional nodes
+        debug!("[NODE READ]: flowctrl send DysCmds to read dysfunctional nodes");
+        if let Err(error) = self
```
joshuef, line 83: we shouldn't need to reach into the `node` for this. `Sender` is `clone`, so we can keep an instance of that on `FlowCtrl`

joshuef (approved):

PR #1786, opened 2022-11-24
chore(client): don't assume Register response with more entries as its most up to date replica

bochaco:
If a Register's replica is found to contain more entries than another replica retrieved from another Adult, it doesn't necessarily mean it's a more up to date version of it, it could be the case that the last operation the user has made was to remove an entry from such Register.


bochaco:
bors try

joshuef (approved):

PR #1793, opened 2022-11-26
Small tweaks

joshuef:


joshuef:
bors try

joshuef:
bors try-

bochaco (approved):
Perhaps it's not important as we are not making releases right now, but this should be marked as `BREAKING CHANGE` since the msgs are serialised using a different `MsgPack` format (compact).

joshuef:
Updated for that @bochaco

PR #1797, opened 2022-11-28
refactor(test): use `TestNetworkBuilder` to build the testing env

RolandSherwin:
- **chore(test): remove `network_utils` module**

- **refactor(test): simple usage of `TestNetwork`**
  - Modifies the test cases under `flow_ctrl/tests` to make use of `TestNetwork`
  - A network is built with a couple of sections. Then a node is retrieved from one of the sections to be used in the tests.
  - The `peer` from `TestNodeBuilder::build()` refers to `node.info().peer()` where `node` is from the prefix that was passed into `TestNodeBuilder::new()`

- **refactor(test): handle_agreement_on_online_of_elder_candidate**
  - The test is modified to make use of `TestNetwork` utility.
  - The previous env created a custom `NetworkKnowledge` with varying ages for the elders. A node was then created with this custom `NetworkKnowledge`
  - The `elder_age_pattern` can be directly passed into `TestNetworkBuilder::sap()` to achieve the same result.

- **refactor(test): use TestNetwork to simulate elder change**
  - Modifies the test cases under `flow_ctrl/tests` to make use of `TestNetwork`
  - The previous envs had a section go through an elder change resulting in two SAPs.
  - An elder node was then created such that it had the `sk_share` for the new SAP, but not the complete SAP; indicating that its handover is still in progress.
  - The same scenario can be simulated by calling `sap_with_members()` which creates a random SAP but with the provided set of members.

- **refactor(test): spentbook_spend_with_updated_network_knowledge_should_update_the_node**
  - The test is modified to make use of `TestNetwork` utility.
  - Creates a Network with 2 sections.
  - The `dispatcher` is obtained from the genesis prefix

- **refactor(test): modify anti_entropy test cases**
  - Modifies the test cases under `messaging/ae` to make use of `TestNetwork`
  - Simulate a network with two sections, where "our section" has gone through
    3 churn events.
  - The `Env::other_signed_sap` refers to the SAP from prefix1
  - The `Env::proof_chain` refers to the section chain of prefix1.

- **fix(network_builder): check member type before modifying node**
  - The `SocketAddr` of the node is replaced with a new one. Hence the check for the `TestMemberType` should happen prior to that.


sn_node/src/node/messaging/anti_entropy.rs
```diff
@@ -753,26 +774,35 @@ mod tests {
...
     #[tokio::test]
     async fn ae_wrong_dst_key_of_our_section_returns_retry() -> Result<()> {
-        let env = Env::new().await?;
-        let our_prefix = env.node.network_knowledge().prefix();
-
-        let msg = env.create_msg(&our_prefix, env.node.network_knowledge().section_key())?;
-        let sender = env.node.info().peer();
-
-        let bogus_env = Env::new().await?;
+        // create an env with 3 churns in prefix0. And a single chrun in prefix1
```
oetyng, line 785: Nit: chrun => churn

sn_node/src/node/flow_ctrl/tests/network_builder.rs
```diff
@@ -137,22 +137,23 @@ impl<R: RngCore> TestNetworkBuilder<R> {
 
         let mut nodes = Vec::new();
         for node in node_infos {
+            // check MemberType
+            let mem_type = if sap.elders_set().contains(&node.peer()) {
```
oetyng, line 141: Super-nit here.. but maybe `memb_type` is preferrable over `mem_type`?

sn_node/src/node/flow_ctrl/tests/network_builder.rs
```diff
@@ -137,22 +137,23 @@ impl<R: RngCore> TestNetworkBuilder<R> {
...
+            };
+
             let (tx, rx) = mpsc::channel(TEST_EVENT_CHANNEL_SIZE);
             let socket_addr: SocketAddr = (Ipv4Addr::LOCALHOST, 0).into();
             let comm = futures::executor::block_on(Comm::new(socket_addr, Default::default(), tx))
                 .expect("failed to create comm");
             let mut node = node.clone();
             node.addr = comm.socket_addr();
 
-            // check MemberType
```
oetyng, line 147: > The SocketAddr of the node is replaced with a new one. Hence the check for the TestMemberType should happen prior to that.

This is probably not very important, but still I wonder, what difference does it make when checking the sap of elders before or after that `node.addr` is updated?

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -652,23 +603,13 @@ async fn relocation(relocated_peer_role: RelocatedPeerRole) -> Result<()> {
...
-        genesis_sk_set,
-        mpsc::channel(10).0,
-    )
-    .await?;
+    let prefix = Prefix::default();
+    let mut env = TestNetworkBuilder::new(thread_rng())
+        .sap(prefix, 1, 0, None, None)
+        .build();
+
+    let node = env.get_nodes(prefix, 1, 0, None).remove(0).0;
```
oetyng, line 611: I'm still a bit unclear here, does `get_nodes` instantiate `MyNode` and add it to a list in env, and then taking/cloning/borrowing it from that, or re-instantiating every time `get_node` is called?

I know it can be looked up, but I wonder from the perspective of just understanding it while reading the tests.
Is it possible to make it clearer by the fn name what is happening and what we can expect side effect way?

Not a biggie I'd say, but in lack of real things to comment on.. :)

oetyng: (Possible duplicate of Q from previous PR, I actually missed to look up your response to that one. I will go check it now.)

oetyng (changes requested):
LGTM :)

The agility of this new setup can be seen both in the details, and in the removed LOC.
Nice!

(I'll request changes, even though there is hardly anything important to change.)

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -9,70 +9,76 @@
...
 
-    let (dispatcher, _, _, sk_set) =
-        network_utils::TestNodeBuilder::new(Prefix::default(), elder_count())
-            .build()
-            .await?;
+    let prefix = Prefix::default();
+    let env = TestNetworkBuilder::new(thread_rng())
+        .sap(prefix, elder_count(), 0, None, None)
+        .build();
+    let dispatcher = env.get_dispatchers(prefix, 1, 0, None).remove(0).0;
```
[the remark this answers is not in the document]
RolandSherwin: Yeah, it's not intuitive.. I will modify the method to just return `Vec<MyNode/Dispatcher>`.

sn_node/src/node/flow_ctrl/tests/network_builder.rs
RolandSherwin: Ah, its just when we call `sap.elder_set().contains()`, we compare using `Peer` which contains `Peer::socket_addr` and we don't wanna change that before the comparison. Could've gone with a map, but this seemed simpler..

sn_node/src/node/flow_ctrl/tests/mod.rs
RolandSherwin: Maybe, yeah.. `get_nodes` can mean a lot of things, maybe `build_*` will be more apt? But my reasoning was that the `get_*` methods were "retrieving stuffs" from the testing env.

Something like `build_sk_share` sounds a bit off too, as it seems like we are building a new share each time, while instead we are just retrieving the share from the `sk_set`

Maybe I shall update the docs to something like `Get the MyNode instance => Build a new MyNode instance`

sn_node/src/node/flow_ctrl/tests/network_builder.rs
oetyng: Ok, I see!

There's no risk that we try fetch the node from the sap anywhere after this, and get a `Peer` with an oudated `SocketAddr`?

RolandSherwin: Yup! That's risky and can lead to logical issues.. 
I think the best way to solve it is by constructing all the SAPs via the `NetworkBuilder` instead of creating it outside. And I think we need to pass in the `TestSapBuilder` directly into the `NetworkBuilder` to achieve that.. 

This should remove all the `sap_*` methods and instead have a single way to create SAPs.. I still have that in my todo list! Will work on it and do a follow-up PR!

sn_node/src/node/flow_ctrl/tests/mod.rs
oetyng: Yeah, no worries. This can be refined!

oetyng (approved):

PR #1807, opened 2022-11-29
refactor: remove event sender

### LINKED ISSUE #1778: Remove `event_sender` in node
the event_sender struct is not used for anything beyond generating some logs, but adds a fair bit of code complexity.

We should remove this event sender as essentially unused.

Buttt... We still want to generate cmd logs. (Can this just be shifted to directly log inside the `process_cmd` func?

RolandSherwin:
Fixes #1778 
- refactor(event_sender): remove `EventSender` and `Event`
  - Also removes the `routing_minimal` and `routing_stress` examples. They were broken prior to this, as most of the `Event` logic was not present in sn_node. Creating an issue to re-enable them.
- feat(node): try to rejoin network if removed
    - Restart the node bootstrap process if we've been removed from a section. Use `Receiver<RejoinNetwork>` to listen for the command.


joshuef (approved):

joshuef:
nice cleanup 💪

PR #1809, opened 2022-11-29
refactor(msgs): remove one layer of indirection

oetyng:
- `ValidateMsg` can be replaced with `HandleMsg`.

### DIFF
--- a/sn_node/src/node/flow_ctrl/tests/mod.rs
+++ b/sn_node/src/node/flow_ctrl/tests/mod.rs
@@ -107,7 +107,7 @@ async fn membership_churn_starts_on_join_request_from_relocated_node() -> Result
     )?;
 
     let _ = run_and_collect_cmds(
-        Cmd::ValidateMsg {
+        Cmd::HandleMsg {
             origin: relocated_node.peer(),
             wire_msg,
             send_stream: None,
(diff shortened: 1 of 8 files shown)


sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1498,12 +1498,12 @@ async fn spentbook_spend_with_updated_network_knowledge_should_update_the_node()
     )
     .await?;
 
-    // The commands returned here should include the new command to update the network
-    // knowledge and also the other two commands to replicate the spent proof shares and
-    // the ack command, but we've already validated the other two as part of another test.
-    assert_eq!(cmds.len(), 3);
-    let update_cmd = cmds[0].clone();
-    assert_matches!(update_cmd, Cmd::UpdateNetworkAndHandleValidClientMsg { .. });
+    // // The commands returned here should include the new command to update the network
```
joshuef, line 1501: Think we're missing to update this test, no?

oetyng: Indeed. But it is already ignored. Not sure it's worth it since it will be reimplemented.

joshuef: ah the containing thing here is ignored? Aye right cool.

should we remove the commented stuff or it's there until re-impl time as a reminder?

joshuef (approved):

PR #1813, opened 2022-11-29
ci(benchmarks): run the job on an ec2 instance

oetyng:


oetyng:
Later thing to worry about.

PR #1814, opened 2022-11-29
refactor(client): use internal send_query API to send initial probe msg to the network

bochaco:
It also fixes the private `send_query_without_retry` API which is currently ignoring the no-retry flag.


bochaco:
bors try

bochaco:
bors try

oetyng (approved):
Nice! Clean and simple.

PR #1816, opened 2022-11-30
Remove node Link connection storage

joshuef:

### DIFF
--- a/sn_client/src/api/mod.rs
+++ b/sn_client/src/api/mod.rs
@@ -84,7 +84,7 @@ impl Client {
             self.session.endpoint.public_addr()
         );
 
-        let mut attempts = 1;
+        let mut attempts = 0;
         loop {
             // Send the dummy message to probe the network for it's infrastructure details.
             match self.send_query_without_retry(query.clone()).await {
@@ -114,6 +114,10 @@ impl Client {
                         "Initial probe msg to network failed. Trying again (attempt #{}): {:?}",
                         attempts, result
                     );
+
+                    // do a tiny sleep to not hammer the network and allow any AE flows there
+                    // time to complete
+                    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await
                 }
             }
         }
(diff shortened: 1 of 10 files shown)


joshuef:
bors try

joshuef:
bors try

sn_node/src/node/error.rs
```diff
@@ -30,6 +30,10 @@ pub type Result<T, E = Error> = std::result::Result<T, E>;
 #[derive(Debug, Error)]
 #[allow(missing_docs)]
 pub enum Error {
+    /// Client message response would have redirected to ourselves _again_.
+    /// So we bail and attempt to update our knowledge instead
+    #[error("Node is missing network knowledge of sibling section. We should trigger AntiEntropy probe to update.")]
+    MissingSiblingNetworkKnowledge,
```
bochaco, line 36: I think having context info like the node's own/sibling section prefix will be useful in order to know which is the sibling section info missing in that node.

sn_client/src/api/mod.rs
```diff
@@ -84,7 +84,7 @@ impl Client {
             self.session.endpoint.public_addr()
         );
 
-        let mut attempts = 1;
+        let mut attempts = 0;
```
bochaco, line 87: This is wrong as per how we use it within the loop, and how we compare it against `NETWORK_PROBE_MAX_ATTEMPTS`

sn_client/src/api/mod.rs
```diff
@@ -114,6 +114,10 @@ impl Client {
                         "Initial probe msg to network failed. Trying again (attempt #{}): {:?}",
                         attempts, result
                     );
+
+                    // do a tiny sleep to not hammer the network and allow any AE flows there
+                    // time to complete
+                    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await
```
bochaco, line 120: I'm not sure we need this, the AE would be all completed within the `send_query_without_retry` lifecycle before returning, although we may need a timeout around the call to `send_query_without_retry` as I think it's not bounded and could be stuck for long?

bochaco (changes requested):

joshuef:
bors try

joshuef:
I'm going to close this, as I have the feeling it introduces some more issues in/around too many conections being spawned and slowing us down. (I've seen big pauses where we have eg 300 incoming connections in a oner... this appears to block things, though I'm not sure why atm).

https://github.com/maidsafe/safe_network/pull/1823 is a candidate to sort this in another way

PR #1820, opened 2022-11-30
feat(client): retry once if connection was lost when trying to send on a bi-stream

bochaco:
Also, adding more context info to sn_client `Error::FailedToInitateBiDiStream` and `Error::ChunkUploadValidationTimeout` error variants.


bochaco:
bors try

joshuef (approved):

PR #1822, opened 2022-11-30
feat(knowledge): update sibling on split

oetyng:
This PR lets both children get the sibling sap at split.

When sending a msg that was intended for the sibling, and expecting an ae update with the sibling sap, it would instead return the sap of the first section itself (because it didn't know its sibling).

With the changes, both children are aware of each other from start, and we now get the sibling sap returned in the ae update.

### DIFF
--- a/sn_interface/src/messaging/system/proposal.rs
+++ b/sn_interface/src/messaging/system/proposal.rs
@@ -7,9 +7,9 @@
 // permissions and limitations relating to use of the SAFE Network Software.
 
 use crate::{
-    messaging::system::{SectionSigShare, SectionSigned},
+    messaging::system::SectionSigShare,
     messaging::{Error, Result},
-    network_knowledge::{NodeState, SectionAuthorityProvider},
+    network_knowledge::{NodeState, SapCandidate, SectionAuthorityProvider},
 };
 
 use itertools::Either;
@@ -29,13 +29,7 @@ pub enum Proposal {
     RequestHandover(SectionAuthorityProvider),
     /// After Handover consensus, the elders inform the candidates that they are promoted (elder->candidates)
     /// Contains the candidate's SAP signed by the current elder's section key
-    NewElders(SectionSigned<SectionAuthorityProvider>),
-    /// After Handover consensus, the elders inform the candidates that they are promoted (elder->candidates)
-    /// Contains the candidate's SAP signed by the current elder's section key
-    NewSections {
-        sap1: SectionSigned<SectionAuthorityProvider>,
-        sap2: SectionSigned<SectionAuthorityProvider>,
-    },
+    HandoverCompleted(SapCandidate),
     /// Proposal to change whether new nodes are allowed to join our section (elder->elder)
     JoinsAllowed(bool),
 }
@@ -69,9 +63,11 @@ impl Proposal {
         let bytes = match self {
             Self::VoteNodeOffline(node_state) => bincode::serialize(node_state),
             Self::RequestHandover(sap) => bincode::serialize(sap),
-            Self::NewElders(signed_sap) => bincode::serialize(&signed_sap.sig.public_key), // the pub key of the new elders
-            Self::NewSections { sap1, sap2 } => {
+            Self::HandoverCompleted(SapCandidate::ElderHandover(signed_sap)) => {
                 // the pub key of the new elders
+                bincode::serialize(&signed_sap.sig.public_key)
+            }
+            Self::HandoverCompleted(SapCandidate::SectionSplit(sap1, sap2)) => {
                 let sap1_sig = bincode::serialize(&sap1.sig.public_key).map_err(|err| {
                     Error::Serialisation(format!(
                         "Couldn't serialise the Proposal '{:?}': {:?}",
@@ -104,6 +100,7 @@ mod tests {
     use crate::test_utils::{TestKeys, TestSapBuilder};
 
     use eyre::Result;
+    use itertools::Either;
     use serde::Serialize;
     use std::fmt::Debug;
     use xor_name::Prefix;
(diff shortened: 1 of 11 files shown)


sn_node/src/node/messaging/handover.rs
```diff
@@ -147,9 +154,13 @@ impl MyNode {
...
-        let sap_bytes = Proposal::RequestHandover(sap.value.clone()).as_signable_bytes()?;
-        if !sap.sig.verify(&sap_bytes) {
-            return Err(Error::InvalidSignature);
+        match Proposal::RequestHandover(sap.value.clone()).as_signable_bytes()? {
+            Either::Right(_) => return Err(Error::InvalidSignature),
+            Either::Left(sap_bytes) => {
+                if !sap.sig.verify(&sap_bytes) {
+                    return Err(Error::InvalidSignature);
+                }
+            }
```
davidrusu, line 163: This code should agree first and foremost with the SectionsDag logic right? So we shouldn't be wrapping the sap in a `Proposal::RequestHandover(..)` in order to get the signable bytes since that's not what SectionsDag does.

SectionsDag just bincode encodes the sap.

How about we factor out a function in the sections_dag module to encode the SAP? and then whenever we need to encode a SAP for signing, we encode it through the `sections_dag::encode_sap_for_dag(sap) -> Vec<u8>` function.

This would remove the RequestHandover wrapping here and the subsequent pattern matching that comes because of it.

sn_interface/src/messaging/system/proposal.rs
```diff
@@ -40,20 +47,45 @@ impl Proposal {
...
+                        "Couldn't serialise the Proposal '{:?}': {:?}",
+                        self, err
+                    ))
+                })?;
+                let sap2_sig = bincode::serialize(&sap2.sig.public_key).map_err(|err| {
+                    Error::Serialisation(format!(
+                        "Couldn't serialise the Proposal '{:?}': {:?}",
+                        self, err
+                    ))
+                })?;
```
davidrusu, line 86: Does the error string really add much? I personally prefer clearer logic here
```suggestion
                let sap1_sig = bincode::serialize(&sap1.sig.public_key)?;
                let sap2_sig = bincode::serialize(&sap2.sig.public_key)?;
```

sn_node/src/node/messaging/handover.rs
oetyng: Well, sure, but that seems like another issue, another PR.
This code here did not introduce/change the existing logic of this fn, only updated so it handled the Either.

sn_interface/src/messaging/system/proposal.rs
oetyng: This is the error that the code maps bincode serialise error to (see line 60 below, as it does not map the bincode error automatically here.

I just reused the same error from line 60.

If we want to modify/update the errors, that's a bit more than the scope of this refactor I'd say.

This PR intends to change nothing else in logic/handling/results, only get the siblings to be aware of eachother.

oetyng: Yeah, I went with that and reduced the number of declarations back to one again.

joshuef:
bors try

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1223,3 +1220,17 @@ async fn spentbook_spend_with_updated_network_knowledge_should_update_the_node()
 
     Ok(())
 }
+
+fn get_single_sig(proposal: &Proposal) -> Result<Vec<u8>> {
+    match proposal.as_signable_bytes()? {
+        itertools::Either::Left(bytes) => Ok(bytes),
+        itertools::Either::Right(_) => bail!("Invalid expectations! Use another proposal variant."),
+    }
+}
```
davidrusu, line 1229: If this is just for tests, have it panic instead of bail so that we get nicer error's with stack traces when we write wrong tests.

It also reduces symbol noise in the tests.

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1223,3 +1220,17 @@ async fn spentbook_spend_with_updated_network_knowledge_should_update_the_node()
...
+
+fn get_double_sig(proposal: &Proposal) -> Result<(Vec<u8>, Vec<u8>)> {
+    match proposal.as_signable_bytes()? {
+        itertools::Either::Left(_) => bail!("Invalid expectations! Use another proposal variant."),
+        itertools::Either::Right((bytes1, bytes2)) => Ok((bytes1, bytes2)),
+    }
+}
```
davidrusu, line 1236: same here

sn_node/src/node/messaging/agreement.rs
```diff
@@ -125,6 +125,79 @@ impl MyNode {
...
+                        MyNode::write_section_tree(&context);
+                    }
+                    Err(err) => {
+                        error!("Error updating our network knowledge for {their_prefix:?}: {err:?}")
+                    }
+                    _ => {}
+                }
+            }
+            Err(err) => error!("Failed to generate proof chain for a newly received SAP: {err:?}"),
+        }
```
davidrusu, line 197: The excessive matching  hides the logic, any reason we can't be using `?` here?
```suggestion
        parent_section_chain.insert(&parent_key, their_sap.section_key(), sig_over_them.signature)?;
        let their_prefix = their_sap.prefix();
        let section_tree_update = SectionTreeUpdate::new(their_sap, parent_section_chain);
        if self.network_knowledge.update_knowledge_if_valid(section_tree_update, None, &context.name)? {
            info!("Updated our network knowledge for {:?}", their_prefix);
            info!("Writing updated knowledge to disk");
            MyNode::write_section_tree(&context);
        }
```

sn_node/src/node/messaging/proposal.rs
```diff
@@ -173,6 +181,56 @@ impl MyNode {
...
+                        trace!(
+                            "Proposals from {} inserted in aggregator, not enough sig shares yet: {serialised_proposal_1:?} and {serialised_proposal_2:?} {:?}",
+                            sender,
+                            msg_id);
+                    }
+                    (_, Err(error)) | (Err(error), _) => {
+                        error!(
+                            "Failed to add proposal from {}, {:?}: {:?}",
+                            sender, msg_id, error
+                        );
```
davidrusu, line 224: You can save 3 lines by in-lining the variables
```suggestion
                        error!("Failed to add proposal from {sender}, {msg_id:?}: {error:?}");
```

sn_node/src/node/messaging/proposal.rs
```diff
@@ -173,6 +181,56 @@ impl MyNode {
...
+                    }
+                    (Ok(Some(_)), Ok(None)) | (Ok(None), Ok(Some(_))) => {
+                        warn!(
+                            "Unexpected aggregation result from {} {:?}: one sig is aggregated while the other is not. This should not happen.",
+                            sender,
+                            msg_id);
```
davidrusu, line 230:
```suggestion
                        warn!("Unexpected aggregation result from {sender} {msg_id:?}: one sig is aggregated while the other is not. This should not happen.");
```

davidrusu (approved):
👍 comments are mostly nits

oetyng:
bors try

joshuef:
This merge has been removed from main as it introduced CI instability: https://github.com/maidsafe/safe_network/pull/1822#issuecomment-1337509608

PR #1823, opened 2022-12-01
TEST: use qp2p with stream resets after 500ms

joshuef:
alt approach to avoid mem leakage (if it is about open streams dangling, this should clarify that at least)


joshuef:
bors try

PR #1824, opened 2022-12-01
style: applied clippy nightly

b-zee:
Ran nightly by accident so made use of it to apply some fixes anyway.

Mainly lints `iter_kv_map` (e.g. use `keys()`, instead of `iter()`, which iterates over both keys and values where the value is not used) and `needless_borrow`.


davidrusu (approved):
clippy ❤️

PR #1825, opened 2022-12-01
chore(node): don't store client peer sessions

joshuef:

### DIFF
--- a/sn_node/src/comm/mod.rs
+++ b/sn_node/src/comm/mod.rs
@@ -113,6 +113,10 @@ impl Comm {
 
         let sessions = self.sessions.clone();
 
+        // TODO: Is there an optimium we should actually have.
+        // Assuming we dont store clients... can we test for this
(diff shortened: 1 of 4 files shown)


joshuef:
bors try

joshuef:
bors try

RolandSherwin (approved):

sn_node/src/comm/mod.rs
```diff
@@ -113,6 +113,10 @@ impl Comm {
 
         let sessions = self.sessions.clone();
 
+        // TODO: Is there an optimium we should actually have.
+        // Assuming we dont store clients... can we test for this
+        trace!("Sessions known of: {:?}", sessions.len());
```
oetyng, line 118: Hmm.. should correlate with how many members we have in the section roughly? 
And then some for contacts with other sections..

So, we could compare sessions count with our network knowledge basically.

There should be at most
- one per node in our section
- one per Elder in other sections we know of

I think?

PR #1834, opened 2022-12-02
Nr 1: Client to elder msg happy path

oetyng:
- feat(cmds): add `cmd-happy-path` feature
This allows clients to default to a lower impact interaction with
elders, only expanding the impact on failures.
- feat(queries): add `query-happy-path` feature
This allows clients to default to a lower impact interaction with
elders (todo: only expanding the impact on failures).
Adds combined feat for both cmd and query happy path.
- chore(ci): add `msg-happy-path` e2e ci job
- chore(docs): update `sn_client` readme

***

A follow up PR (#1856, currently in draft) adds happy path in the next step of client data cmd; data replication from elders to adults.


.github/workflows/run_pr_checks.yml
```diff
@@ -206,13 +206,13 @@ jobs:
         timeout-minutes: 5
 
       - name: Build all tests before running non ubuntu
-        run: cargo test --no-run --release --features check-replicas --package sn_client
```
joshuef, line 209: I think in the end we'll want this as a separate job run. So we'll have both `strict` data checks, and then a happy path run too.

sn_client/src/connections/messaging.rs
```diff
@@ -91,13 +92,84 @@ impl Session {
         let kind = MsgKind::Client(auth);
         let wire_msg = WireMsg::new_msg(msg_id, payload, kind, dst);
 
-        let send_cmd_tasks = self.send_msg(elders.clone(), wire_msg).await?;
-        trace!("Cmd msg {:?} sent", msg_id);
+        #[cfg(feature = "check-replicas-w-happy-path")]
+        {
+            self.send_to_closest_and_expand(dst_address, elders.clone(), wire_msg)
```
joshuef, line 97: `send_to_closest_and_try_more_on_failure` ?

joshuef (commented):
I realise this is still draft, but made a couple comments.

Also, perhaps out of scope for this initial PR, but just to note:

There's an extra happy path opportunity at the elders too, in the data records. We currently send Cmds to all 4 data replicas and await responses from all. So that can be reduced to just one positive from closest and we return? (the msgs to the rest will need to keep going out too, though).

.github/workflows/run_pr_checks.yml
oetyng: Yes, this combination is included only in the temporary

```
chore(feats): temp change
- This is to test ci with the happy path feature.
```
commit, which is only there to testrun the draft in ci. That commit is not to be kept.

In the previous commit, they are separated.

sn_client/src/connections/messaging.rs
oetyng: I agree `send_to_closest_and_expand` is not very clear. I think `send_to_closest_and_try_more_on_failure` is too verbose though.

oetyng:
> There's an extra happy path opportunity at the elders too, in the data records. We currently send Cmds to all 4 data replicas and await responses from all. So that can be reduced to just one positive from closest and we return? (the msgs to the rest will need to keep going out too, though).

Yep, had a look and seems it's a bit more than trivial to add that one, so I suggest that to be another PR to extend the `cmd-happy-path` feature.

joshuef:
So I was thinking on this, and I think we need to clearly lay out what our (3?) "paths" are (and I think get it in the documentation for where it's relevant).

So we have: 

## "default": This is our standard setup
- Msg all elders for Cmds + Elders wait on All Acks
- Msg three elders for Queries _to one adult_ + wait on all responses

## "strict / data storage verification": this is used for all CI runs right now.
- [client side] Msg three elders for queries _to all adults_ + wait on all responses 

## "happy path"
- [client side] Clients contact one elder for Cmds/queries + Elder waits on all acks? (that is this PR)
- [node side] Clients contact one elder for Cmds/queries + Elder waits on one ack/response? (that is the 1856)

----

So for testing we want:

- `"strict / data storage verification"` which is as `default` but makes sure each adult is storing data.
- client happy path + node happy paths tested (together... would be the least reliable setup, right?).
- client happy path + node happy path churn testing. (are we slowly losing more data).


Is there anything missing there?

I think we can set up some issues to cover all these cases and track that we're not missing/skipping some testing on this front.

sn_client/src/sessions/messaging.rs
```diff
@@ -78,25 +83,115 @@ impl Session {
...
+            }
+            #[cfg(not(feature = "cmd-happy-path"))]
+            {
+                log_line(format!("{elders_len}"));
+                self.send_to(msg_id, elders.clone(), wire_msg).await
+            }
+        }
+    }
+
+    async fn send_to(&self, msg_id: MsgId, elders: Vec<Peer>, wire_msg: WireMsg) -> Result<()> {
```
joshuef, line 118: I was confused reading this at first.

I think we need some docs to outline this vs `send_to_one_or_more`

joshuef:
@oetyng can you add the tests to the `merge.yml` flows too please. For mac/win also

sn_client/src/sessions/messaging.rs
oetyng: I renamed it to `send_msg_and_check_acks`, aiming to perhaps have that be self-documenting?

oetyng:
> @oetyng can you add the tests to the `merge.yml` flows too please. For mac/win also

Added.

joshuef:
bors try

oetyng:
bors try

oetyng:
bors try-

oetyng:
bors try

joshuef:
bors try

.github/workflows/run_pr_checks.yml
```diff
@@ -276,6 +276,116 @@ jobs:
         if: failure()
         continue-on-error: true
 
+  e2e-msg-happy-path:
```
maqi, line 279: just to confirm, this means for each PR, we run e2e tests both with/without the happy-path, right?

sn_client/README.md
```diff
@@ -9,7 +9,35 @@
...
+### "default": This is our standard setup
+(Helps with detecting bugs)
+- Msg _all elders_ for cmds + elders wait on _all acks_ from adults
+- Msg _three elders_ for queries _to one adult_ + wait on _all responses_ from elders
+
+### "check-replicas": this is used for all CI runs right now.
+(Verifies data was stored)
+- Msg _three elders_ for queries _to all adults_ + wait on _all responses_
+
+### "msg-happy-path" [client side]: this is used for sn_client e2e CI runs right now.
```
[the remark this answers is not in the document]
joshuef: They may be. We may want to test them individually? I'm unsure what's best coverage for us as yet. (But we can always merge them down the line?)

.github/workflows/run_pr_checks.yml
joshuef: Yeh, even if happy path ere to become default, we'd still need to be sure everything was being written and returned properly across the board.

sn_client/src/sessions/messaging.rs
```diff
@@ -90,13 +95,117 @@ impl Session {
...
+        }
+    }
+
+    /// This fn will try a happy path,
+    /// successively expanding to all the other elders in case of failure.
+    ///
+    /// 1st attempt: Closest Elder (take 1)
+    /// 2nd attempt: Next closest (skip 1, take 1)
+    /// 3rd attempt: Next 2 closest (skip 2, take 2)
+    /// 4th attempt: Next 3 closest (skip 3, take 3)
```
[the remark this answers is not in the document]
oetyng: I think this will do it:

```
    /// 1st attempt: Closest Elder (take 1) (take index 0)
    /// 2nd attempt: Next closest (skip 1, take 1) (skip idx 0, take idx 1)
    /// 3rd attempt: Next 2 closest (skip 2, take 2) (skip idx 0 and 1, take idx 2 and 3)
    /// 4th attempt: Next 3 closest (skip 4, take 3) (skip idx 0-3 inclusive, take index 4, 5 and 6)
```

sn_client/src/sessions/messaging.rs
```diff
@@ -90,13 +95,117 @@ impl Session {
...
+                .we_have_sufficient_acks_for_cmd(msg_id, elders, send_cmd_tasks)
+                .await
+                .is_ok()
+            {
+                trace!("Cmd {:?} sent", msg_id);
+                return Ok(());
+            }
+        }
+
+        // we expected at least one ack, but got 0
```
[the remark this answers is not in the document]
oetyng: That's true.
But it could be an OK approxm. for now. Requires a bit more thinking to get that number accurate, but it's not a super important value.

maqi (approved):

joshuef:
bors try

PR #1838, opened 2022-12-04
[test] chore(node): remove connection storage altogether

joshuef:
Trying again as we've expanded/changed the flow for a lot of the cmd channel stuff to hopefully be non-blocking.

So perhaps that was a contributing factor to the slowdown/connection batching?


joshuef:
bors try

joshuef:
still problematic

PR #1842, opened 2022-12-04
fix(node): include any relevant full adults in query targets

joshuef:


joshuef:
bors try

joshuef:
Closing as this does not fully solve our "full" adults issue w/r/t data stored. We'll likely proceed to just ignore "fullness" and treat failure to return chunks as faults. This removes a bunch of logic and should simplify things/  get larger nodes into the network

PR #1846, opened 2022-12-05
Always update sap in network knowledge when we receive an AE update.

davidrusu:
Currently we have logic that avoids updating the signed sap in Network-knowledge if we were an elder and we had not received our section key share yet.

This PR removes this logic as it was preventing other changes I was trying to make and it doesn't actually seem to be what we want: If a SAP update has occurred, then hanging on to the old SAP until we finish DKG does little to help the network as the other elders will update soon if they hadn't already, any signature shares this elder is creating with the old section key will only have a potential to be aggregated for that short period while we still have a threshold of old elders around still using the old SAP.

On a more philosophical point, we probably don't want elders to be signing anything after we've agreed to a handover.

### DIFF
--- a/sn_node/src/node/flow_ctrl/tests/mod.rs
+++ b/sn_node/src/node/flow_ctrl/tests/mod.rs
@@ -33,9 +33,7 @@ use sn_interface::{
     elder_count, init_logger,
     messaging::{
         data::{ClientMsg, DataCmd, SpentbookCmd},
-        system::{
-            self, AntiEntropyKind, JoinAsRelocatedRequest, NodeDataCmd, NodeMsg, SectionSigned,
-        },
+        system::{self, AntiEntropyKind, JoinAsRelocatedRequest, NodeDataCmd, NodeMsg},
         Dst, MsgType, WireMsg,
     },
     network_knowledge::{
@@ -45,7 +43,6 @@ use sn_interface::{
     },
     test_utils::*,
     types::{keys::ed25519, PublicKey, ReplicatedData},
-    SectionAuthorityProvider,
 };
 
 use assert_matches::assert_matches;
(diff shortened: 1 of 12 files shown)


joshuef:
bors try

oetyng (approved):

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -812,49 +813,43 @@ async fn handle_demote_during_split() -> Result<()> {
...
-        })
+        }
     };
 
     // Sign the saps.
     let signed_sap_0 = TestKeys::get_section_signed(&sk_set0.secret_key(), sap0);
     let signed_sap_1 = TestKeys::get_section_signed(&sk_set1.secret_key(), sap1);
 
-    let cmd = create_our_elders_cmd(signed_sap_0, signed_sap_1)?;
+    let cmd = create_our_elders_cmd(signed_sap_0, signed_sap_1);
```
RolandSherwin, line 835: The closure is used only once, maybe we can simplify it into a block? I think the other closure got removed in the latest set of changes

sn_node/src/node/mod.rs
```diff
@@ -417,20 +417,24 @@ mod core {
...
         }
 
-        fn initialize_membership(&mut self, sap: SectionAuthorityProvider) -> Result<()> {
-            let key = self
-                .section_keys_provider
-                .key_share(&self.network_knowledge.section_key())?;
+        fn initialize_membership(&mut self, key: SectionKeyShare) -> bool {
+            let sap = self.network_knowledge.signed_sap().value;
+
+            if let Some(m) = self.membership.as_ref() {
```
maqi, line 433: better put a comment here explaining this is for the case that the we got our keyshare AFTER updated network knowledge to new sap

sn_node/src/node/mod.rs
```diff
@@ -439,29 +443,33 @@ mod core {
...
 
-        fn initialize_handover(&mut self) -> Result<()> {
-            let key = self
-                .section_keys_provider
-                .key_share(&self.network_knowledge.section_key())?;
-            let n_elders = self.network_knowledge.section_auth().elder_count();
+        fn initialize_handover(&mut self, key: SectionKeyShare) -> bool {
+            let sap = self.network_knowledge.signed_sap().value;
+
+            if let Some(h) = self.handover_voting.as_ref() {
```
maqi, line 452: same as above, better put a comment here explaining why doing this check

davidrusu:
bors try

sn_node/src/node/flow_ctrl/tests/mod.rs
davidrusu: done

sn_node/src/node/mod.rs
oetyng: Yes, I was looking at this for a while and thinking "what is the connection..". So, seems a log would help.

sn_node/src/node/mod.rs
oetyng: Yes, I was looking at this for a while and thinking "what is the connection..". So, seems a log would help.

sn_node/src/node/mod.rs
oetyng: Yeah I realized that after first moment of wondering. Good stuff.

Issue #1810, opened 2022-11-29
Dummy section tree update on bootstrap shall be removed


maqi:
As described in the Issue https://github.com/maidsafe/safe_network/issues/1798
The dummy section tree update on bootstrap shall be removed to avoid mis-usage

davidrusu:
@maqi do we need both issues? is this a dup of #1798?

Issue #1798, opened 2022-11-28
Remove the dummy section tree update on bootstrap


davidrusu:
This dummy update serves no purpose other than making things compile and could result in bugs down the road.

_Originally posted by @maqi in https://github.com/maidsafe/safe_network/pull/1651#discussion_r1028077376_

PR #1848, opened 2022-12-06
fix: remove dummy SectionTree update

RolandSherwin:
Fixes https://github.com/maidsafe/safe_network/issues/1810 
Fixes https://github.com/maidsafe/safe_network/issues/1798

- fix(section_tree): genesis_sap is required to create the `SectionTree`
  - The fields of the tree are assumed to be in sync. But it is not the case for a newly created tree.
  - This can be fixed by accepting the genesis sap while creating the tree.
- refactor(network_knowledge): rework constructor
  - Allows us to create a minimal `NetworkKnowledge` by just providing the genesis SAP. This will be set as our current `signed_sap`. A new `SectionTree` is also created with the provided SAP.
  - Optionally, we can provide a `SectionTree` which will replace our current tree. In this case, the provided SAP is considered to be part of the tree and is set as our current `signed_sap`

### DIFF
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -154,10 +154,10 @@ impl NetworkKnowledge {
     /// Creates a new `NetworkKnowledge` instance. The prefix is used to choose the
     /// `signed_sap` from the provided SectionTree
     pub fn new(prefix: Prefix, section_tree: SectionTree) -> Result<Self, Error> {
-        let signed_sap = match section_tree.get_signed(&prefix) {
-            Some(signed_sap) => signed_sap.clone(),
-            None => return Err(Error::NoMatchingSection),
-        };
+        let signed_sap = section_tree
+            .get_signed(&prefix)
+            .cloned()
+            .ok_or(Error::NoMatchingSection)?;
         Ok(Self {
             signed_sap,
             section_peers: SectionPeers::default(),
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -200,10 +200,10 @@ impl NetworkKnowledge {
     }
 
     /// update all section info for our new section
-    pub fn relocated_to(&mut self, new_network_knowledge: Self) -> Result<()> {
-        debug!("Node was relocated to {:?}", new_network_knowledge);
-        self.signed_sap = new_network_knowledge.signed_sap();
-        let _updated = self.merge_members(new_network_knowledge.section_signed_members())?;
+    pub fn relocated_to(&mut self, new_name: XorName) -> Result<()> {
+        let signed_sap = self.section_tree().get_signed_by_name(&new_name)?;
+        debug!("Node was relocated to {:?}", signed_sap.section_key());
+        self.signed_sap = signed_sap;
 
         Ok(())
     }
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -151,20 +151,13 @@ pub struct NetworkKnowledge {
 }
 
 impl NetworkKnowledge {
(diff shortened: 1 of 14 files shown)


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -152,19 +152,28 @@ pub struct NetworkKnowledge {
...
+    /// Create a `NetworkKnowledge` with the provided genesis SAP. Optionally provide a
+    /// `SectionTree`, in which case the provided SAP will be considered as the current `signed_sap`
+    ///
+    /// Error if the SAP is not present in the tree.
     pub fn new(
-        section_tree: SectionTree,
-        section_tree_update: SectionTreeUpdate,
+        signed_sap: SectionSigned<SectionAuthorityProvider>,
+        section_tree: Option<SectionTree>,
     ) -> Result<Self, Error> {
```
davidrusu, line 162: This change has some non-intuitive semantics, it has two modes
1. normal case: you pass both a signed sap and a section tree. 
2. genesis case: you pass a signed_sap.

This is not ideal from an API point of view. i.e. it's not clear just from reading the following code that the signed sap is expected to be a genesis sap:

```
NetworkKnowledge::new(sap, None)
```

There's a few alternatives I think would be better:

1. Divide out the two modes into two constructors:
```rust
NetworkKnowledge::genesis(sap: SectionSigned<SAP>)
NetworkKnowledge::new(sap: SectionSigned<SAP>, section_tree: SectionTree)
```

2. take a prefix instead of sap, this makes the user constructor the section tree and pass a prefix as  a reference to where they are in the tree:
```rust
NetworkKnowledge::new(prefix: Prefix, section_tree: SectionTree)
```

I kind of like option 2. The benefits are simpler logic here in the constructor with less error checking:
1. we don't have to check that the signed sap we fetch with `tree.get_signed(prefix)` matches the signed sap passed to the constructor.
2. we don't have to check that the signed sap is a valid genesis sap (if we are in the genesis mode)
3. probably less cloning on the callers side.

RolandSherwin:
bors try

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -152,20 +152,13 @@ pub struct NetworkKnowledge {
...
-            return Err(err);
-        }
-
+    /// Creates a new `NetworkKnowledge` instance. The prefix is used to choose the
+    /// `signed_sap` from the provided SectionTree
+    pub fn new(prefix: Prefix, section_tree: SectionTree) -> Result<Self, Error> {
+        let signed_sap = match section_tree.get_signed(&prefix) {
+            Some(signed_sap) => signed_sap.clone(),
+            None => return Err(Error::NoMatchingSection),
+        };
```
davidrusu, line 161:
```suggestion
        let signed_sap = section_tree.get_signed(&prefix).cloned().ok_or(Error::NoMatchingSection)?;
```

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -47,12 +47,17 @@ pub struct SectionTree {
...
+    pub fn new(genesis_sap: SectionSigned<SectionAuthorityProvider>) -> Result<Self> {
+        if genesis_sap.prefix() != Prefix::default() {
+            return Err(Error::NonGenesisSap);
         }
+
+        let genesis_pk = genesis_sap.public_key_set().public_key();
+        Ok(Self {
+            sections: BTreeMap::from([(Prefix::default(), genesis_sap)]),
+            sections_dag: SectionsDAG::new(genesis_pk),
+        })
```
davidrusu, line 60: 👍

davidrusu (approved):
Looks good, comments are nits 🚢

RolandSherwin:
bors try

RolandSherwin:
bors try

RolandSherwin:
bors cancel

PR #1851, opened 2022-12-06
fix: remove full adults from data storage algos

### LINKED ISSUE #1850: Remove Full Adult special case for data storage/retrievals
Right now we avoid attempting to store on 'full adults' and change retrieval targets when we have them.

This is proving problematic.

Instead we should just remove these caveats and we punish any node not able to store/retrieve data which is should have. 

Thus full nodes will be votes offline eventually.

joshuef:
fixes #1850

### DIFF
--- a/sn_node/src/node/data/capacity.rs
+++ b/sn_node/src/node/data/capacity.rs
@@ -61,15 +61,6 @@ impl Capacity {
             .collect()
     }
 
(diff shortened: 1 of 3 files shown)


joshuef:
marked as do not merge util we have churn test in there

b-zee (approved):

Issue #1841, opened 2022-12-04
Add feat flag to disable joins after prefix X


joshuef:
The ability to prevent node joins after an arbitrary prefix size will be handy to reduce variables in testnets.

(We can kill nodes specifically to allow others to join).

This could bea  simple feat flag we enable in a given built

joshuef:
conceptually fixed by #1853

PR #1853, opened 2022-12-06
chore(node): add feat to limit section size

joshuef:
Allows for one less variable during testnets


joshuef:
fixes aims of #1841 though by checking node count, not prefix len

joshuef:
bors try

RolandSherwin (approved):

PR #1854, opened 2022-12-06
feat(storage): report threshold reached only

oetyng:
- This removes the book keeping of storage level on Elders.
- Makes Adults report only when threshold reached.
- Makes Elders allow joins until split when majority of Adults reached threshold.


sn_node/src/node/mod.rs
```diff
@@ -128,6 +125,7 @@ mod core {
         // Section handover consensus state (Some for Elders, None for others)
         pub(crate) handover_voting: Option<Handover>,
         pub(crate) joins_allowed: bool,
+        pub(crate) joins_allowed_until_split: bool,
```
joshuef, line 128: Can you add some docs to clarify this vs standard `joins_allowed` please.

oetyng:
Superseded by #1926.

PR #1855, opened 2022-12-06
Update sibling on split

oetyng:

### DIFF
--- a/sn_node/src/node/flow_ctrl/tests/mod.rs
+++ b/sn_node/src/node/flow_ctrl/tests/mod.rs
@@ -296,7 +296,7 @@ async fn handle_agreement_on_offline_of_non_elder() -> Result<()> {
     let node_state = NodeState::left(node_state, None);
 
     let proposal = Proposal::VoteNodeOffline(node_state.clone());
-    let sig = TestKeys::get_section_sig_bytes(&sk_set.secret_key(), &proposal.as_signable_bytes()?);
+    let sig = TestKeys::get_section_sig_bytes(&sk_set.secret_key(), &get_single_sig(&proposal)?);
 
     let _cmds = run_and_collect_cmds(Cmd::HandleAgreement { proposal, sig }, &dispatcher).await?;
 
@@ -325,7 +325,7 @@ async fn handle_agreement_on_offline_of_elder() -> Result<()> {
 
     // Handle agreement on the Offline proposal
     let proposal = Proposal::VoteNodeOffline(remove_elder.clone());
-    let sig = TestKeys::get_section_sig_bytes(&sk_set.secret_key(), &proposal.as_signable_bytes()?);
+    let sig = TestKeys::get_section_sig_bytes(&sk_set.secret_key(), &get_single_sig(&proposal)?);
 
     let _cmds = run_and_collect_cmds(Cmd::HandleAgreement { proposal, sig }, &dispatcher).await?;
 
@@ -677,8 +677,7 @@ async fn handle_elders_update() -> Result<()> {
     // current elders and promote the oldest peer.
     let elders_1: BTreeSet<_> = sap1.elders_set();
     let proposal = Proposal::NewElders(sap1.clone());
-    let sig =
-        TestKeys::get_section_sig_bytes(&sk_set0.secret_key(), &proposal.as_signable_bytes()?);
+    let sig = TestKeys::get_section_sig_bytes(&sk_set0.secret_key(), &get_single_sig(&proposal)?);
 
     let cmds = run_and_collect_cmds(
         Cmd::HandleNewEldersAgreement {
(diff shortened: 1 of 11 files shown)


sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1108,3 +1106,19 @@ async fn spentbook_spend_with_updated_network_knowledge_should_update_the_node()
...
+        }
+    }
+}
+
+fn get_double_sig(proposal: &Proposal) -> Result<(Vec<u8>, Vec<u8>)> {
+    match proposal.as_signable_bytes()? {
+        itertools::Either::Left(_) => panic!("Invalid expectations! Use another proposal variant."),
+        itertools::Either::Right((bytes1, bytes2)) => Ok((bytes1, bytes2)),
+    }
+}
```
davidrusu, line 1124:
```suggestion
fn get_double_sig(proposal: &Proposal) -> (Vec<u8>, Vec<u8>) {
    match proposal.as_signable_bytes().expect("Failed to serialize") {
        itertools::Either::Left(_) => panic!("Invalid expectations! Use another proposal variant."),
        itertools::Either::Right((bytes1, bytes2)) => (bytes1, bytes2),
    }
}
```

sn_node/src/node/flow_ctrl/tests/mod.rs
```diff
@@ -1108,3 +1106,19 @@ async fn spentbook_spend_with_updated_network_knowledge_should_update_the_node()
...
 }
+
+fn get_single_sig(proposal: &Proposal) -> Result<Vec<u8>> {
+    match proposal.as_signable_bytes()? {
+        itertools::Either::Left(bytes) => Ok(bytes),
+        itertools::Either::Right(_) => {
+            panic!("Invalid expectations! Use another proposal variant.")
+        }
+    }
+}
```
davidrusu, line 1117:
```suggestion
fn get_single_sig(proposal: &Proposal) -> Vec<u8> {
    match proposal.as_signable_bytes().expect("Failed to serialize") {
        itertools::Either::Left(bytes) => bytes,
        itertools::Either::Right(_) => {
            panic!("Invalid expectations! Use another proposal variant.")
        }
    }
}
```

sn_node/src/node/messaging/proposal.rs
```diff
@@ -173,6 +181,40 @@ impl MyNode {
...
+
+                let res1 = self
+                    .proposal_aggregator
+                    .try_aggregate(&serialised_proposal_1, sig_share);
+                let res2 = self
+                    .proposal_aggregator
+                    .try_aggregate(&serialised_proposal_2, sig_share2);
+                let res = (res1, res2);
+
+                match res {
```
davidrusu, line 200: Unnecessary variable

```suggestion

                match (res1, res2) {
```

sn_node/src/node/messaging/proposal.rs
```diff
@@ -173,6 +181,40 @@ impl MyNode {
...
+                    (Ok(Some(sig1)), Ok(Some(sig2))) => match proposal {
+                        Proposal::HandoverCompleted(SapCandidate::SectionSplit(sap1, sap2)) => cmds
+                            .push(Cmd::HandleNewSectionsAgreement {
+                                sap1,
+                                sig1,
+                                sap2,
+                                sig2,
+                            }),
```
davidrusu, line 208: Do we need this `HandleNewSectionsAgreement` command? Can't we use two `HandleNewEldersAgreement` commands here instead?

```
cmds.push(Cmd::HandleNewEldersAgreement { sap1, sig1 })
cmds.push(Cmd::HandleNewEldersAgreement { sap2, sig2 })
```

sn_node/src/node/messaging/proposal.rs
oetyng: Yeah I tested that first but got some compile error so just didn't spend any more time on it.
But I changed now and there was maybe something else causing it initially, because works fine.

sn_node/src/node/messaging/proposal.rs
oetyng: Seems we do, if we want to update for our own section first. When we go double cmd we make them async and don't know if we update our own section first.

sn_node/src/node/messaging/agreement.rs
```diff
@@ -125,6 +125,79 @@ impl MyNode {
...
+                        MyNode::write_section_tree(&context);
+                    }
+                    Err(err) => {
+                        error!("Error updating our network knowledge for {their_prefix:?}: {err:?}")
+                    }
+                    _ => {}
+                }
+            }
+            Err(err) => error!("Failed to generate proof chain for a newly received SAP: {err:?}"),
+        }
```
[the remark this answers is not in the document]
oetyng: it's not calling the exact same methods there though `update_knowledge_if_valid` vs `update_network_knowledge`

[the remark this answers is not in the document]
oetyng: We can't be calling `update_network_knowledge` because it has validation logic within it that expect us to hold a key share of the added section (which we won't for our sibling).

sn_node/src/node/messaging/proposal.rs
davidrusu: Why would we want to update our own section first?

joshuef:
bors try

davidrusu (approved):

PR #1866, opened 2022-12-07
tests: use isolated config dir for cli tests

jacderida:
- 20996c68e **feat: provide `--config-dir-path` arg for `safe`**

  Add a global `--config-dir-path` argument that will set the location of the config directory. If
  it's not set, the `SAFE_CONFIG_DIR_PATH` environment variable will be used, and if that's not set,
  we will default to `$HOME/.safe`.

  The purpose of this customisation is for integration testing so that each test can use an isolated
  configuration directory, but users can apply it too if they wish.

- fb887ede4 **tests: use isolated config dir for cli tests**

  A function is added to the test utils library for creating a temporary directory that functions as
  the config directory for the CLI. It uses `assert_fs`, so the directory will be automatically
  deleted when the reference to it is dropped. This enables each test case to use its own config
  directory, which resolves issues we've had with multiple instances of `safe` writing/reading the
  config file in a race situation.

  The functions for running `safe` commands were changed to require the config directory to be
  passed. This forces each test case to use this mechanism.

  All the tests were updated to use the same naming convention, with the exception of the `files get`
  tests, which are quite tedious and probably best left with the convention they were using, at least
  for now.

  Some tests could be un-ignored as part of this change.

### DIFF
--- a/sn_cli/src/cli.rs
+++ b/sn_cli/src/cli.rs
@@ -28,7 +28,6 @@ use crate::{
 use clap::{AppSettings::ColoredHelp, Parser};
 use color_eyre::{eyre::eyre, Result};
 use sn_api::{Safe, XorUrlBase};
-use std::env;
 use std::path::PathBuf;
 use tracing::{debug, warn};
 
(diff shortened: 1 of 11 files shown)


sn_cli/src/cli.rs
```diff
@@ -180,22 +181,28 @@ async fn process_commands(mut safe: &mut Safe, args: CmdArgs, config: &mut Confi
...
+///
+/// The reason for customisation of the location is mainly for integration testing, but users could
+/// also apply the customisation if they want to.
+async fn get_config(config_dir_path: Option<PathBuf>) -> Result<Config> {
+    let config_dir_path = if let Some(config_dir_path) = config_dir_path {
+        config_dir_path
+    } else {
+        let mut default_config_path =
+            dirs_next::home_dir().ok_or_else(|| eyre!("Couldn't find user's home directory"))?;
+        default_config_path.push(".safe");
```
bochaco, line 197: It sounds to me that if neither the `SN_CLI_CONFIG_PATH` env var is set, nor `--config-dir-path` arg is passed, then the default path shall be `$HOME/.safe/cli` instead of `$HOME/.safe`. This is supposed to be the path for the CLI config file/s rather than the path where all SAFE apps install their configs or related files....?

bochaco: ...hmmm..never mind, I see this would require additional changes since we use it for finding the network contacts too...so perhaps for a review in the future ....

bochaco (approved):
It'd be nice if we can enable multi-threaded tests in CI for CLI as part of this?

bochaco:
bors try

joshuef:
bors try

PR #1870, opened 2022-12-08
fix(chunks): match on specific error

oetyng:
Matching on any error was wrong, but it was hidden until we actually returned any other error than the expected.
With `limit-client-upload-size`, we started returning more errors, and we botched the logic with it, causing `TooLargeAsSmallFile` error to be returned.
This commit fixes that problem by correctly identifying the error type that signals too small file for self encryption, and erroring out on all other types.


davidrusu (approved):
🚀

PR #1880, opened 2022-12-13
fix(sn_node): update network knowledge AFTER sent join request

maqi:

### DIFF
--- a/sn_node/src/node/bootstrap/join.rs
+++ b/sn_node/src/node/bootstrap/join.rs
@@ -133,26 +133,29 @@ impl<'a> Joiner<'a> {
         self.bootstrap_section_tree(self.join_target_sap()?, response_timeout)
             .await?;
 
-        let mut target_sap = self.join_target_sap()?;
+        let target_sap = self.join_target_sap()?;
         let section_key = target_sap.section_key();
         let msg = NodeMsg::JoinRequest(JoinRequest { section_key });
         self.send(msg, &target_sap.elders_vec(), section_key, false)
             .await?;
 
         loop {
-            let (response, sender) =
-                tokio::time::timeout(response_timeout, self.receive_join_response())
-                    .await
-                    .map_err(|e| {
-                        error!("Failed to receive join response: {:?}", e);
-                        Error::JoinTimeout
-                    })??;
+            let (response, sender) = tokio::time::timeout(
+                response_timeout,
+                self.receive_join_response_and_handle_ae_updates(),
+            )
+            .await
+            .map_err(|e| {
+                error!("Failed to receive join response: {:?}", e);
+                Error::JoinTimeout
+            })??;
+            let target_sap = self.join_target_sap()?;
 
             match response {
                 JoinResponse::Approved { decision } => {
                     info!("{}", LogMarker::ReceivedJoinApproval);
                     if let Err(e) = decision.validate(&target_sap.public_key_set()) {
-                        error!("Dropping invalid join decision: {e:?}");
+                        error!("Failed to validate with {target_sap:?}, dropping invalid join decision: {e:?}");
                         continue;
                     }
 
@@ -196,7 +199,7 @@ impl<'a> Joiner<'a> {
                         trace!("Bootstrapping section tree for retry");
                         self.bootstrap_section_tree(target_sap, response_timeout)
                             .await?;
-                        target_sap = self.join_target_sap()?;
+                        let target_sap = self.join_target_sap()?;
 
                         let new_keypair = ed25519::gen_keypair(
                             &target_sap.prefix().range_inclusive(),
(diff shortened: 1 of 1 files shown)


sn_node/src/node/bootstrap/join.rs
```diff
@@ -152,7 +158,25 @@ impl<'a> Joiner<'a> {
...
+                        )
+                        .await
+                        .map_err(|_| Error::JoinTimeout)??;
+
+                        info!("After request sent, received section tree update: {update:?}");
+
+                        if self.section_tree.update(update)? {
+                            let old_target = target_sap;
+                            target_sap = self.join_target_sap()?;
+                            info!("After request sent, sap changed from {old_target:?} to {target_sap:?}");
```
joshuef, line 177: can we clarify what the `request` is here and above?

joshuef (commented):
Is there a way to test this portion of the join process?

sn_node/src/node/bootstrap/join.rs
```diff
@@ -152,7 +158,25 @@ impl<'a> Joiner<'a> {
...
                     info!("{}", LogMarker::ReceivedJoinApproval);
                     if let Err(e) = decision.validate(&target_sap.public_key_set()) {
-                        error!("Dropping invalid join decision: {e:?}");
+                        error!("Failed to validate with {target_sap:?}, dropping invalid join decision: {e:?}");
+
+                        // A possible network sap change happened,
+                        // try to update knowledge via `NodeMsg::AntiEntropy` once.
+                        let update = tokio::time::timeout(
+                            response_timeout,
+                            self.receive_section_tree_update(),
```
davidrusu, line 167: `receive_section_tree_update()` does not probe the network, it merely waits for an update. What is triggering the network to send us the update?

If it's the accepted node flow that you hope will send us the update, it could be that we dropped it in the  `self.receive_join_response()` call (receive_join_response() skips all non-JoinResponse related messages).


An alternative approach:

- this loop needs to match on two type of messages: JoinResponse and AntiEntropy
- the most direct way would be to delete the `receive_join_response()` helper and use `receive_node_msg()` directly.
- This could further allow us to remove the `JoinResponse::Approval` message entirely since we can simply wait for a valid AntiEntropy update that contains us as a member

sn_node/src/node/bootstrap/join.rs
```diff
@@ -140,19 +140,20 @@ impl<'a> Joiner<'a> {
...
-            let (response, sender) =
-                tokio::time::timeout(response_timeout, self.receive_join_response())
+            let (response, sender, current_sap) =
+                tokio::time::timeout(response_timeout, self.receive_join_response(target_sap))
                     .await
                     .map_err(|e| {
                         error!("Failed to receive join response: {:?}", e);
                         Error::JoinTimeout
                     })??;
+            target_sap = current_sap;
```
davidrusu, line 150: If we just calculate the most up-to-date target here, we don't need to pass the sap to `receive_join_response()` or have `receive_join_response()` return the sap.
```suggestion
            target_sap = self.join_target_sap();
```

sn_node/src/node/bootstrap/join.rs
```diff
@@ -365,11 +366,27 @@ impl<'a> Joiner<'a> {
     }
 
     #[tracing::instrument(skip(self))]
-    async fn receive_join_response(&mut self) -> Result<(JoinResponse, Peer)> {
+    async fn receive_join_response(
```
davidrusu, line 369:
```suggestion
    async fn receive_join_response_and_handle_ae_updates(
```

sn_node/src/node/bootstrap/join.rs
maqi: we cannot do so, as we could waiting for a JoinResponse from sibling section

sn_node/src/node/bootstrap/join.rs
```diff
@@ -679,6 +698,116 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn join_receive_approval_with_newer_sap() {
```
joshuef, line 702: `join_receive_approval_with_newer_sap_proceeds`? I feel like something is missing from the test name here. It's not clear to me what I'll be expecting to happen.

sn_node/src/node/bootstrap/join.rs
```diff
@@ -679,6 +698,116 @@ mod tests {
...
+
+        let state = Joiner::new(node.clone(), send_tx, &mut recv_rx, tree.clone());
+
+        // Create the bootstrap task, but don't run it yet.
+        let bootstrap = async { state.try_join(join_timeout).await.expect("Failed to join") };
+
+        let (new_sap, new_sks, _new_elders, _) = TestSapBuilder::new(Prefix::default()).build();
+        let new_section_key = new_sks.public_keys().public_key();
+
+        let others = async {
```
joshuef, line 731: `others` what? some more comments outlining the test process will be good.

sn_node/src/node/bootstrap/join.rs
maqi: sry, I gave it a second thought, and realize I might made a wrong statement. Will implement the change according to your comment

joshuef (approved):

PR #1887, opened 2022-12-14
fix(storage): write all Register cmds to disk even if one or more fail

bochaco:
- When writting Register cmds log to disk, we log and return the error for any of them failing, but we don't prevent the rest to be written to disk.
- Enable multi-threaded mode for sn_api tests in Bors.
- Some minor improvements to log msgs.


bochaco:
bors try

PR #1888, opened 2022-12-14
chore(messaging): removing unused payload_debug field from msgs

bochaco:
This change also includes a minor refactor within `comm` mod.


bochaco:
bors try

joshuef (approved):

PR #1891, opened 2022-12-15
refactor(node): removing mutex from SendStream

bochaco:
Each `SendStream` is now moved into either `Cmd`s or functions instead of being shared using a `Mutex` around it.


bochaco:
bors try

joshuef (approved):
bors r+

PR #1897, opened 2022-12-16
chore(node): remove the looping system load log

joshuef:


dirvine (approved):
Perfect, this helps reduce complexity, increase performance and therefor makes code more readable (less locks too :-) )

PR #1900, opened 2022-12-16
feat(join): prevent joins from nodes behind NAT

davidrusu:
This PR adds the joining nodes address to the join request.

If the address is different from the connection address, then this node is behind NAT 🙅


davidrusu:
bors try

dirvine (approved):
Perfect

dirvine (approved):
Perfect!! we need this in for next testnet

PR #1903, opened 2022-12-19
fix(node): make initial get/connect blocking

joshuef:
This avoids many many connection attempts to the same node if one is in progress already.

### DIFF
--- a/sn_node/src/comm/peer_session.rs
+++ b/sn_node/src/comm/peer_session.rs
@@ -256,34 +256,39 @@ impl PeerSessionWorker {
 
         let mut link = self.link.clone();
 
-        let _handle = tokio::spawn(async move {
-            // Attempt to get a connection or make one to another node.
-            // if there's no connection here yet, we requeue the job after a wait
-            // incase there's been a delay adding the connection to Comms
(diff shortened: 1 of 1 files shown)


sn_node/src/comm/peer_session.rs
```diff
@@ -256,34 +256,39 @@ impl PeerSessionWorker {
...
 
-                    return;
+                // we await here in case the connection is fresh and has not yet been added
+                tokio::time::sleep(tokio::time::Duration::from_millis(50)).await;
+                if let Err(e) = queue.send(SessionCmd::Send(job)).await {
+                    warn!("Failed to re-enqueue job {id:?} after failed connection retrieval error {e:?}");
                 }
-            };
 
+                return Ok(SessionStatus::Ok);
```
maqi, line 287: just to confirm, so here we queue the job and return saying the session status is ok, right?

joshuef: yup. We re-enqueue and carry on after the small timeout, and this clal would be repeated until we actually are making new connectons.

maqi (approved):

PR #1904, opened 2022-12-19
chore(node): prevent overzealous replication, limiting send msg commands

joshuef:
This adds a hard limit on outgoing msgs and throughput of replication per node

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -35,6 +35,7 @@ use sn_interface::{
 use super::DataStorage;
 use std::sync::Arc;
 use tokio::sync::{mpsc, RwLock};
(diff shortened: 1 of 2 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -207,6 +208,10 @@ impl FlowCtrl {
                                 error!("Failed to enqueue send msg command for replication of data batch to {peer:?}: {error:?}");
                             }
                             data_batch = vec![];
+
+                            // w/ batch size of data_batch_size (mb) at most, this would limit us to:
+                            // 25mb * 2 / s
```
maqi, line 213: this is based on the assumption that each data pushed into the data_batch is always 1mb , right?
if the data size is smaller, the overall speed will be smaller as well, right?

PR #1907, opened 2022-12-19
feat(sn_node): bundle messages according to size and number

maqi:


sn_interface/src/types/mod.rs
```diff
@@ -99,6 +99,15 @@ impl ReplicatedData {
         }
     }
 
+    pub fn size(&self) -> u64 {
+        let length = match self {
+            Self::Chunk(chunk) => chunk.payload_size(),
+            Self::RegisterWrite(_) | Self::SpentbookWrite(_) => 300,
```
joshuef, line 105: where does 300 come from, could/should it be a const?

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -250,3 +251,31 @@ impl FlowCtrl {
         })
     }
 }
+
+static DATA_BUNDLE_SIZE_LIMIT: u64 = 10_000_000;
+static DATA_BUNDLE_ENTRY_LIMIT: usize = 25;
```
joshuef, line 256: I think we could drop entry limit altogether no?

sn_interface/src/types/mod.rs
maqi: I just go through the RegisterCmd and set 300 as it's average size. maybe other value will be better.

I will change it to be an const anyway.

sn_node/src/node/flow_ctrl/mod.rs
maqi: Not sure. maybe the `entry limit` shall still apply to avoid too many register cmds to be bundled together?

joshuef: I'm not sure what would be the problem with filling up with registers though?

maqi: When handling the bundled datas, there will still be message per entry to be sent out.
Bundling too many entries may result in an inner choke over there.
https://github.com/maidsafe/safe_network/blob/main/sn_node/src/node/messaging/node_msgs.rs#L458

joshuef: okay, lets raise the cap though as 25 doe not seem like a lot of registers to me... 250?

joshuef (approved):

PR #1909, opened 2022-12-20
feat(cli): print networks as json

RolandSherwin:


sn_cli/src/subcommands/networks.rs
```diff
@@ -161,7 +165,33 @@ pub async fn networks_commander(
...
             }
         }
-        None => config.print_networks().await,
+        None => {
+            let networks = config.get_networks().await;
+            if OutputFmt::Json == output_fmt {
+                let json = serde_json::to_string(&networks)?;
+                println!("{json}");
+            } else {
+                let mut table = Table::new();
```
bochaco, line 174: I think we should still have a helper function `print_networks` so we don't duplicate this code to print after we called `get_networks()`.

bochaco (changes requested):

bochaco (approved):

bochaco:
bors merge

PR #1916, opened 2022-12-21
fix(log): build tracing `Subscriber` without custom macros + small fixes

RolandSherwin:
- fix(log): use vector of `Layers` to build the `Subscriber`
  - Replaces the macros with a vector containing boxed Layers
- chore(testnet): enable otlp feature flag
- fix(log): prevent panic when we have multiple tracing subscribers
  - When we have the `otlp` feature enabled, we effectively have
  multiple subscribers consuming the logs.
  - Under rare circumstances, we get a panic with the following msg,
  `Format: was already formatted once`
  - Turns out it's because the `itertools::format()` is being called
  multiple times
  - Was not able to reproduce the issue with a minimal setup, but this
  gets rid of the panic.
- chore(log): remove unused event formatting option
  - the `.event_format()` overrides the `.with_thread_names()` option,
  hence remove it

### DIFF
--- a/sn_interface/src/lib.rs
+++ b/sn_interface/src/lib.rs
@@ -142,7 +142,6 @@ pub fn init_logger() {
         tracing_subscriber::fmt::fmt()
             // NOTE: uncomment this line for pretty printed log output.
             //.pretty()
-            .with_thread_names(true)
             .with_ansi(false)
             .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
             .with_target(false)
--- a/sn_node/src/bin/sn_node/log/mod.rs
+++ b/sn_node/src/bin/sn_node/log/mod.rs
@@ -27,9 +27,7 @@ impl TracingLayers {
             } else {
                 Box::new(Targets::new().with_target(current_crate_str(), config.verbose()))
             };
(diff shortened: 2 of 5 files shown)


sn_interface/src/lib.rs
```diff
@@ -142,7 +142,6 @@ pub fn init_logger() {
         tracing_subscriber::fmt::fmt()
             // NOTE: uncomment this line for pretty printed log output.
             //.pretty()
-            .with_thread_names(true)
```
grumbach, line 145: We don't want those anymore?

RolandSherwin: Yeah, the following line `.event_format(LogFormatter::default())` overrides the ` .with_thread_names(true)` option. So I guess we can get rid of it.. We want to enable the `thread_names` inside `LogFormatter` if we want actually want to display them..

RolandSherwin: Not sure if want them though. I can work on enabling it if needed.

grumbach (approved):
Functions feel cleaner than massive macros indeed 👍

b-zee (approved):
Nice work, good to get rid of the macros this way. It's the only time I wrote macros because it got a little complex with types, but I didn't like it.

PR #1921, opened 2022-12-21
refactor(node): serialise the msg header only once when replicating data to holders

bochaco:
- Also removing some minor but unnecessary WireMsg caching, and some objs cloning.


bochaco:
bors try

joshuef (approved):

PR #1922, opened 2022-12-22
fix(log): initialize logging just once

RolandSherwin:
- opentelemetry tracing requires a tokio runtime to be present, hence leave a separate rt running if `otlp` feature is enabled


grumbach (approved):
LGTM

PR #1925, opened 2022-12-26
refactor(node): simplify logic retrieve qp2p config from sn_node::Config

bochaco:
- We used to keep an instance of qp2p::Config within sn_node::Config which required to keep in sync when modifying/reading network related config, we now simply build a qp2p::Config just when is queried from sn_node::Config.

### DIFF
--- a/sn_node/examples/config_handling.rs
+++ b/sn_node/examples/config_handling.rs
@@ -25,13 +25,11 @@ async fn main() -> Result<()> {
     let file_config = Config::new().await?;
 
     // TODO: Uncomment the below lines once we enable reading config from disk
-    // file_config.network_config.local_ip = Some(
-    //     "192.168.0.1"
-    //         .parse()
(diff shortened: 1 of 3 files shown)


bochaco:
bors try

sn_node/examples/config_handling.rs
```diff
@@ -25,13 +25,15 @@ async fn main() -> Result<()> {
     let file_config = Config::new().await?;
 
     // TODO: Uncomment the below lines once we enable reading config from disk
-    // file_config.network_config.local_ip = Some(
-    //     "192.168.0.1"
+    // file_config.local_addr = Some(
```
dirvine, line 28: I know this is commented but seems wrong anyway to have a 192 address there

dirvine (approved):

sn_node/examples/config_handling.rs
bochaco: I'll change it to `127.0.0.1`

PR #1926, opened 2022-12-26
Make elders also store data

oetyng:
- Removes the storage level state and updates.
- Removes the storage distinction between `Elders` and `Adults`, which means that `Elders` and `Adults` are all considered equal nodes, responsible for storing data in the same way.
- Adds a `joins_allowed_until_split` variable + logic, which can now be used by `Elders` to split based on section storage level, without bookkeeping the levels of other nodes. They do this by assuming that the section general level of used space is similar to their own level.

### Follow up PRs:
- The `JoinsAllowedUntilSplit` addition could be split out to its own PR. But it's not a biggie.
- To ensure as close to uniform distribution of nodes as possible (and handle cases where it is not).


oetyng:
bors try

sn_node/src/node/data/records.rs
```diff
@@ -503,8 +498,11 @@ impl MyNode {
             }
         }
 
-        // Whenever there is a replicated data in, we send back a query again
-        if !is_full && has_replicated_data_in {
+        // As long as the data batch is not empty, we send back a query again
+        // to continue the replication process (like pageing).
+        // This means there that there will be a number of repeated `give-me-data -> here_you_go` msg
+        // exchanges, until there is no more data missing on this node.
+        if !is_full && data_batch_is_empty {
```
joshuef: I know this isn't coming from this PR. But I think actually we'll never get back an empty data batch. The flow of "get missing data" returns early if there's nothing to send. 

So I think we can just remove this logic? (Unless we need that msg for some reason I'm missing? in which case we should update the flow there to ensure it's sent)

sn_node/src/node/data/records.rs
```diff
@@ -222,36 +224,68 @@ impl MyNode {
...
-            "{:?} preparing to query adults for data at {:?} with op_id: {:?}",
+            "{:?} preparing to query other nodes for data at {:?} with op_id: {:?}",
             LogMarker::DataQueryReceviedAtElder,
             address,
             operation_id
         );
 
-        let targets = MyNode::target_data_holders(&context, *address.name());
+        let mut targets = MyNode::target_data_holders(&context, *address.name());
+        // since we already checked our local storage above, we do not need to retain ourselves in the list of targets
```
joshuef: I think this breaks the `nth` node checking. As the queried elder may be that `nth` node. But we'll never know if it is or if the data is on the next node.

I think we should probably move this data storage query down around line 305, so if we are nth node, we check our storage (and skip the node msg roundtrip).

joshuef (changes requested):
Still got a couple more commits to check, but so we have the comments thus far...

sn_node/src/node/data/records.rs
oetyng: I think it's a good fail safe, that doesn't rely on code in other places (decouples it). These are the conditions for which it will keep sending the query. It's somewhat documenting as well.

sn_node/src/node/data/records.rs
oetyng: There was a retain done in a previous commit, where we excluded the elder, and that broke the nth node checking.
But then the retain part was removed, and I think the nth node checking no longer breaks with that.

oetyng: Yeah, I can see that this comment was made on the previous commit, where the logic changed in subsequent commit,

sn_node/src/node/data/records.rs
```diff
@@ -513,15 +478,10 @@ impl MyNode {
...
-            // This may return a DatabaseFull error... but we should have reported storage increase
+            // This may return a DatabaseFull error... but we should have reported StorageError::NotEnoughSpace
             // well before this
             match store_result {
-                Ok(level_report) => {
-                    info!("Storage level report: {:?}", level_report);
-                    record_storage_level = Some(level_report);
-                    info!("End of message flow.");
-                }
+                Ok(()) => info!("End of message flow."),
```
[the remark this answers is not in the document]
oetyng: This was also made on a previous commit.
The line is no longer there.

sn_node/src/node/data/records.rs
oetyng: Btw, there was naming bug (not logic bug though) at that specific commit, as the negation operator was used at the assignment. So what we are checking for is `!data_batch_is_empty`. But I accidently named it so that it seems we are checking for it being empty.

oetyng:
bors try

oetyng:
bors try-

sn_node/src/node/mod.rs
```diff
@@ -639,20 +638,12 @@ mod core {
...
-                            .section_auth()
-                            .elders()
-                            .filter(|peer| !old_elders.contains(&peer.name()))
-                            .cloned()
-                            .collect(),
-                        &self.network_knowledge.prefix(),
-                    ),
-                );
-            };
+            // when we split, the range of data that nodes are responsible for is halved,
```
maqi, line 641: not exactly. The network is discreted on xorname. 
Hence the `data` will cling to the node and stay within the same sub-section.
i.e. in most chance, data won't be `halved` but `stay same` in the holder.

oetyng: Ah, you're right.

1st section is responsible for the entire xor-range.
All the nodes together hold all data that is uploaded.

But, the nodes are (supposedly) uniformly distributed, and we have (on avg) equal number on either side of the mid point in the range. 

So, the data is already located in what will later be the child sections.

No data from a future sibling section is held by nodes.

oetyng: This has implications for the design and the assumptions.

Will be back later to continue this.

oetyng: Okay, all good as is. The wider assumptions still hold, as we are adding nodes to actually split, and thereby we get the decrease in used space.

Will remove/change the comment accordingly.

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -395,39 +380,34 @@ impl MyNode {
...
+                let mut cmds = vec![];
+                if full {
                     debug!("[NODE WRITE]: CouldNotStore write gottt...");
-                    let changed = write_locked_node
-                        .set_storage_level(&node_id, StorageLevel::from(StorageLevel::MAX)?);
-                    if changed {
-                        // ..then we accept a new node in place of the full node
-                        write_locked_node.joins_allowed = true;
+                    // Kick out node!
+                    let nodes = BTreeSet::from([node_id.into()]);
```
maqi, line 387: I am not sure, but in case a node can store data due to full, we cann't just simply kick it out?
this will just make the network overall storage situation worse?

oetyng: Hmm. yeah. Should it be just logging node issue here?

I can see that Josh has added a comment like this in a working branch, over the offline proposal:

```
// TODO: Ideally we should only boot it when it's been failing to fulfill
// data... this gives more nodes a chance to step in (once we have preloading)
```

oetyng: Yeah, I replaced that with only logging the issue.
I'm guessing that in the end, the same problem is there. We're removing a node when it's full. We're just adding a few more steps in between.

So, it isn't really a solution to the problem you describe.

But, if we were to primarily act by taking in new nodes when a node reports full (and Elders accumulate that)... then we would avoid that specific problem I think.

I think that perhaps the risk of the reporting to be abused is low.. but that depends on how the joining is designed.
If the joining is designed to pick a replacement in the same sub-range as the lost node, and accept >1 node... then this could be used to gain dominance in the section, by replacing 1 node that you own, with > 1 node that you own (given that you have control over the queue as well).

joshuef (approved):
bors r+

sn_interface/src/messaging/serialisation/wire_msg.rs
```diff
@@ -282,13 +282,8 @@ mod tests {
...
 
         let msg_id = MsgId::new();
-        let pk = crate::types::PublicKey::Bls(dst.section_key);
 
-        let msg = NodeMsg::NodeDataCmd(NodeDataCmd::RecordStorageLevel {
-            node_id: pk,
-            section: pk.into(),
-            level: StorageLevel::zero(),
-        });
+        let msg = NodeMsg::HandoverAE(100);
```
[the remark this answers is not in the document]
oetyng: The test uses an arbitrary msg to test deserialization.
For the value, that is just an old habit to put non-default values, i.e. for this type non-zero values, when checking successful deserialization and similar.

oetyng:
bors try

oetyng:
bors try-

PR #1930, opened 2022-12-27
fix(client): retry only once when check-replicas test query fails due to diff responses

bochaco:
- Re-enabling e2e sn_client tests multi-threaded mode.
- If a connection times out right after we sent a msg, but before we try to finish the stream, since we don’t await for the `finish` result we miss the chance to detect it and retry with new conn. This was made this way to avoid the await, but it seems, that at least for the client, we may want to wait for the `finish` result so we can retry in such scenarios.
- Retrying a test query only once, and only if the failure was due to receiving different responses from the data replicas, which may be due to a temporary out of sync. We've seen this very occasionally in CI with Register queries/entries.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -233,7 +233,7 @@ jobs:
           SN_CMD_TIMEOUT: 90
           SN_QUERY_TIMEOUT: 90
           RUST_LOG: "sn_client=trace,qp2p=debug"
-        run: cd sn_client && cargo test --release --features check-replicas -- --test-threads=1
+        run: cd sn_client && cargo test --release --features check-replicas
         timeout-minutes: 7
 
       - name: Run example app for file API against local network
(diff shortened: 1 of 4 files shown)


bochaco:
bors try

bochaco:
bors try

bochaco:
bors try-

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try-

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

oetyng (approved):

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

PR #1939, opened 2022-12-29
fix(storage): replace modulo op with div

oetyng:
- The operator was a remnant from previous version, that had not yet been replaced.
- Also updates some comments.


joshuef (approved):
bors r+

PR #1941, opened 2022-12-30
Refactor proposals

grumbach:
Refactor the `NodeMsg::Propose` into 4 distinct messages:
- `NodeMsg::RequestHandover`: when nodes finish DKG and request a handover to the current elders (node->elder)
- `NodeMsg::SectionHandoverPromotion`: when elders tell those nodes that they are promoted as elder (elder->node)
- `NodeMsg::SectionSplitPromotion`: when elders tell those nodes that they are promoted as elder in either side of a split (elder->node)
- `NodeMsg::ProposeSectionState`: when elders decide to kick nodes or accept new nodes within a section (elder->elder)

This distinction makes explicit:
- who is signing
- who is receiving/aggregating the signatures

All the changes are in the [refactor: proposal into multiple distinct messages](https://github.com/maidsafe/safe_network/pull/1941/commits/692135f6c803926403cf46ba3b2b747660b04d1d) commit.

The other commits are tidying up and moving code around: large diffs of nothingness!


sn_interface/src/messaging/system/mod.rs
```diff
@@ -60,6 +59,18 @@ pub enum AntiEntropyKind {
...
 }
 
+/// A vote about the state of the section
+/// This can be a result of seeing a node go offline or deciding wether we want to accept new nodes
+/// Anything where we need section authority before action can be taken
+/// Section State Proposals are sent by elders to elders
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum SectionStateVote {
+    /// Vote to remove a node from our section
+    NodeIsOffline(NodeState),
```
oetyng, line 69: There is a fundamental problem here that I mentioned to D.Rusu some months ago (so it is not something new for this PR):

`NodeIsOffline` is voted for with a `NodeState` which contains a `MembershipState`, which can be `Joined`, `Left` or `Relocated`. So we can cast a vote for `NodeIsOffline`, with a state `Joined`, which doesn't make sense?

Just to begin with I see a model mismatch, where we try to do state based and event driven handling at the same time, but never fully settling on what we are doing.
At one side, we are saying "this happened", at another side we are saying "this is our current state".

But the specific problem above I think, is us not having type safety in the system for saying `this is state x`, `this is event e` (that we saw upon that state).

grumbach: Good one! I must admit I just moved and renamed that one without checking the insides of `NodeState` assuming it was correct. It's indeed a type safety issue as you say. 
Let me see what I can do to fix that. 
Since this PR is already quite big and this issue wasn't introduced by this PR, I'm thinking I could do this in a subsequent PR?

joshuef: follow up PR sounds good to me

oetyng: Absolutely, would not have expected you to add it here :)

sn_node/src/node/messaging/handover.rs
```diff
@@ -12,20 +12,132 @@ use crate::node::{
...
+        sig_share: SectionSigShare,
+        sender: Peer,
+    ) -> Result<Vec<Cmd>> {
+        trace!("Handling post DKG handover request {msg_id:?} from {sender:?}: {sap:?}");
+
+        // check sender
+        if !sap.contains_elder(&sender.name()) {
+            trace!("Ignoring invalid handover request {msg_id:?}: not DKG participant");
+            return Ok(vec![]);
+        }
```
oetyng, line 42: Do we vote if we are not ourselves in it? (Does it make sense to check that we are still in the proposed sap?)

sn_node/src/node/messaging/handover.rs
```diff
@@ -12,20 +12,132 @@ use crate::node::{
...
+            .pending_split_sections
+            .entry(membership_gen)
+            .and_modify(|curr| {
+                let _ = curr.insert(signed_sap.clone());
+            })
+            .or_insert_with(|| BTreeSet::from([signed_sap]));
+
+        // if we have reached 2 split SAP candidates for this generation
+        // handle section split (1 to 2)
+        if let [sap1, sap2] = section_candidates_for_gen
```
oetyng, line 123: I'd perhaps use the 0-based indexing, as that coincides with the rest of the references to these further down the flow. (And the prefix bit diff).

sn_node/src/node/messaging/promotion.rs
```diff
@@ -6,122 +6,135 @@
...
-    fn handle_request_handover_agreement(
+    pub(crate) fn handle_section_split_promotion(
         &mut self,
-        sap: SectionAuthorityProvider,
-        sig: SectionSig,
+        msg_id: MsgId,
+        sap1: SectionSigned<SectionAuthorityProvider>,
+        sig_share1: SectionSigShare,
+        sap2: SectionSigned<SectionAuthorityProvider>,
+        sig_share2: SectionSigShare,
```
oetyng, line 80: Maybe use 0-based indexing here as well (and in rest of fn body).
(I know there was already a 1-based used before, further down.)

No biggie and can be updated later.

sn_node/src/node/mod.rs
```diff
@@ -106,22 +106,24 @@ mod core {
...
-        pub(crate) proposal_aggregator: SignatureAggregator,
         // DKG/Split/Churn modules
         pub(crate) dkg_start_aggregator: SignatureAggregator,
         pub(crate) dkg_sessions_info: HashMap<Digest256, DkgSessionInfo>,
         pub(crate) dkg_voter: DkgVoter,
+        pub(crate) elder_promotion_aggregator: SignatureAggregator,
         pub(crate) pending_split_sections:
             BTreeMap<Generation, BTreeSet<SectionSigned<SectionAuthorityProvider>>>,
         pub(crate) relocate_state: Option<Box<JoiningAsRelocated>>,
         // ======================== Elder only ========================
```
oetyng, line 117: Just a note (not something caused by this PR): I think this comment is out of date (or perhaps unclear)? There seems to be elder only related stuff above as well?

oetyng (approved):
Nice work! Great improvement, elegantly solved!

sn_interface/src/messaging/system/mod.rs
```diff
@@ -114,18 +125,44 @@ pub enum NodeMsg {
...
+    /// After Handover consensus, the elders inform the new elder candidates that they are promoted
+    /// The candidates can aggregate the sig_share an obtain SectionSigned proof that they are promoted
+    SectionHandoverPromotion {
+        /// The promoted SAP (signed by themselves)
+        sap: SectionSigned<SectionAuthorityProvider>,
+        /// BLS signature share of an Elder over the sap's pubkey
+        sig_share: SectionSigShare,
+    },
+    /// After Handover consensus, the elders inform the new elder candidates that they are promoted
+    /// The candidates can aggregate the sig_shares an obtain SectionSigned proof that they are promoted
```
maqi, line 150: an => to

sn_interface/src/messaging/system/mod.rs
```diff
@@ -114,18 +125,44 @@ pub enum NodeMsg {
...
+        /// The promoted SAP of section 0 (signed by themselves)
+        sap0: SectionSigned<SectionAuthorityProvider>,
+        /// BLS signature share of an Elder over the sap0's pubkey
+        sig_share0: SectionSigShare,
+        /// The promoted SAP of section 1 (signed by themselves)
+        sap1: SectionSigned<SectionAuthorityProvider>,
+        /// BLS signature share of an Elder over the sap1's pubkey
+        sig_share1: SectionSigShare,
+    },
+    /// A vote about the state of a section to be aggregated in the SectionStateVote aggregator
```
maqi, line 161: maybe make this comment to follow the others' pattern by saying:
Elders votes among themselves regarding SectionStates of node_off or join_allowed

sn_interface/src/network_knowledge/section_authority_provider.rs
```diff
@@ -70,6 +70,7 @@ pub struct SectionAuthorityProvider {
 }
 
 /// `SectionAuthorityProvider` candidates for handover consensus to vote on
+/// Each is signed by their own section key (the one in the SectionAuthorityProvider)
```
maqi, line 73: by their own section key => by their private section key-share

sn_node/src/node/messaging/handover.rs
maqi: This is an inner-section promotion, i.e. the existing elders shall be always stay in the new sap.
if it is not, it means that elder is dead or kicked out.

sn_node/src/node/messaging/handover.rs
```diff
@@ -12,20 +12,132 @@ use crate::node::{
...
+        match self
+            .handover_request_aggregator
+            .try_aggregate(&serialised_sap, sig_share)
+        {
+            Ok(Some(sig)) => {
+                trace!("Handover request {msg_id:?} successfully aggregated");
+                self.handle_request_handover_agreement(sap, sig)
+            }
+            Ok(None) => {
+                trace!("Handover request {msg_id:?} acknowledged, waiting for more...");
```
maqi, line 58: when already aggregated, it will be returned as Ok(None) ?
if so, may worth to be careful with the log

sn_node/src/node/messaging/handover.rs
```diff
@@ -12,20 +12,132 @@ use crate::node::{
...
+        sap: SectionAuthorityProvider,
+        sig: SectionSig,
+    ) -> Result<Vec<Cmd>> {
+        // check if section matches our prefix
+        let equal_prefix = sap.prefix() == self.network_knowledge.prefix();
+        let is_extension_prefix = sap
+            .prefix()
+            .is_extension_of(&self.network_knowledge.prefix());
+        if !equal_prefix && !is_extension_prefix {
+            // Other section. We shouln't be receiving or updating a SAP for
```
maqi, line 80: shouln't => shouldn't

sn_node/src/node/messaging/handover.rs
```diff
@@ -74,15 +189,65 @@ impl MyNode {
...
-            })
+    pub(crate) fn broadcast_handover_completed(
+        &mut self,
+        candidate: SapCandidate,
+    ) -> Result<Vec<Cmd>> {
+        let recipients = candidate.elders();
+        let (others, myself) = self.split_peers_and_self(recipients);
+        let peers = Peers::Multiple(others);
+
+        // sends a promotion message to all of the to-be-Elders with our sig_share over their new pub key
```
maqi, line 200: over their new pub key => over the new sap's public_key

sn_node/src/node/messaging/promotion.rs
```diff
@@ -6,122 +6,135 @@
...
-                sap
-            );
+        trace!("Handling section split promotion message {msg_id:?} by {sender:?} with saps: {sap1:?} {sap2:?}");
+        let our_prefix = self.network_knowledge.prefix();
+        let sig_share_pk1 = &sig_share1.public_key_set.public_key();
+        let sig_share_pk2 = &sig_share2.public_key_set.public_key();
+
+        // Proposal from other sections shall be ignored.
+        if !our_prefix.matches(&sender.name()) {
+            trace!("Ignore promotion message {msg_id:?} from other section");
```
maqi, line 90: just a note, in case the node already got split from previous aggregated SectionSplit,
then the later on handover from old elders may got ignored here.
not an issue but may cause some confusing by the current log.
worth to log our_prefix and the sender together with the log to ease the debugging.

sn_node/src/node/messaging/section_state.rs
```diff
@@ -0,0 +1,152 @@
+// Copyright 2022 MaidSafe.net limited.
```
maqi, line 1: Happy new year
and it's 2023 now  :)

sn_node/src/node/messaging/handover.rs
oetyng: Ah, I was not looking straight. This is already decided, not voting. Not sure how I missed that.

sn_interface/src/network_knowledge/section_authority_provider.rs
grumbach: In this case the aggregation was already done at the previous step, we already have a SectionSigned piece of data (SAP) here.

sn_node/src/node/messaging/promotion.rs
grumbach: So I wanted to do this at first. But there's a catch: if we have sap0 and sap1 we might assume that the sap0 is the one on the 0 side of the split and sap1 on the 1 side. BUT there's nothing that enforces this! So that's why I purposely left it to sap1, sap2 so we don't have these assumptions when reading the code. There are places where it is still sap0, sap1, and it would indeed be prettier to have the same names everywhere.

sn_node/src/node/mod.rs
grumbach: `pending_split_sections` is indeed elder only as well 🤔

sn_node/src/node/messaging/handover.rs
grumbach: When aggregation is finished the aggregator returns the section sig and discards the state, the next received message will start a new round of aggregation so we will indeed get a Ok(None)

grumbach: With that we will indeed probably get `waiting for more...` logs when we are not anymore. What do you suggest?

sn_node/src/node/messaging/section_state.rs
grumbach: #1951

PR #1944, opened 2023-01-02
fix: adapt logging to be less confusing

grumbach:
Move a DKG stalled msg to where it should be! Should declutter the logs.


joshuef (approved):

joshuef:
prev run was udeps fail. bors r+

PR #1945, opened 2023-01-02
Optimize ae probing

oetyng:
This PR reduces ae messaging which is currently causing excess load.
Also, global network section knowledge is refactored to target 1 rand elder in 3 rand sections, every 5 minutes.


joshuef (approved):
bors r+

PR #1946, opened 2023-01-02
ci(fix): use manual cargo udeps

joshuef:
hopefully should avoid random fails due to https://github.com/aig787/cargo-udeps-action/issues/2


oetyng:
bors try

oetyng (approved):

PR #1947, opened 2023-01-02
feat(joining): join within largest empty range

oetyng:
- Generates keys within largest empty range.
- Elders verify that joiner name is within largest range.

This should improve the data uniformity in sections and the network.


oetyng:
Closing for now, as we will get to this feat later on when the node sets are developed.

PR #1948, opened 2023-01-02
fix(network_knowledge): avoid reverifying for partial dag creation

joshuef:
AE back and forth and other messaging can mean we make a good amount of partial_dags. Before this, we were pulling from a verified section_dag and verifying _every_ insert into our fresh partial dag.

This was costly. So now we avoid it.


dirvine (approved):
Yea nice one. I like the 2 versions of insert as opposed to some trickery.

PR #1949, opened 2023-01-02
fix(script): 'all_nodes_joined.sh' script to use all nodes logs instead of only one

bochaco:
The script was using only one of the nodes' logs to find inconsistency in the membership list, but there are cases where this inconsistency is not seen by all nodes by the time we check, thus this PR changes the script to use all nodes' logs to extract a unique list of nodes/members in the network knowledge.


bochaco:
bors try

oetyng (approved):

Issue #1952, opened 2023-01-03
`NodeIsOffline` contains a voted `NodeState` which can contain absurd `Joined`


grumbach:
`NodeIsOffline` is voted for with a `NodeState` which contains a `MembershipState`, which can be `Joined`, `Left` or `Relocated`. So we can cast a vote for `NodeIsOffline`, with a state `Joined`, which doesn't make sense?

Just to begin with I see a model mismatch, where we try to do state based and event driven handling at the same time, but never fully settling on what we are doing.
At one side, we are saying "this happened", at another side we are saying "this is our current state".

But the specific problem above I think, is us not having type safety in the system for saying `this is state x`, `this is event e` (that we saw upon that state).

_Originally posted by @oetyng in https://github.com/maidsafe/safe_network/pull/1941#discussion_r1059334735_
      
`NodeIsOffline` is used for both `Left` (when node is detected offline) and `Relocated` (when node is relocated to another section).

PR #1953, opened 2023-01-03
chore: rename NodeIsOffline NodeStateChange

grumbach:
Help to alleviate #1952


joshuef (approved):
bors r+

oetyng:
Hm, I don't think this alleviates the problem, or does anything to improve the situation.

grumbach:
> Hm, I don't think this alleviates the problem, or does anything to improve the situation.

It doesn't fix the issue that's for sure but I was thinking maybe it would help us not have the wrong assumption that Joins are impossible through it. 
On the other hand it does obfuscate the fact that we only use it for voting nodes off... 
Closing this PR for now until we find a better way. The issue stays open.

PR #1958, opened 2023-01-04
chore(faults): Remove OpId for failed request tracking

joshuef:
We use bidi now, so we can report after any failure, no need for double accounting


bochaco (approved):

bochaco:
bors merge

PR #1960, opened 2023-01-05
chore(sn_node): skip section_tree update earlier for outdated sap

maqi:

### DIFF
--- a/sn_interface/src/network_knowledge/section_tree/mod.rs
+++ b/sn_interface/src/network_knowledge/section_tree/mod.rs
@@ -217,47 +217,47 @@ impl SectionTree {
         let signed_sap = section_tree_update.signed_sap;
         let proof_chain = section_tree_update.proof_chain;
 
-        let incoming_prefix = &signed_sap.prefix();
-        match self.get_signed(incoming_prefix) {
-            Some(sap) if sap == &signed_sap => {
-                // It's the same SAP we are already aware of
-                info!("Dropping SectionTree update since the incoming SAP for prefix '{incoming_prefix}' is not new");
-                return Ok(false);
-            }
-            Some(sap) => {
-                let current_sap_elder_count = sap.elder_count();
-                let proposed_sap_elder_count = signed_sap.elder_count();
-                if proposed_sap_elder_count < current_sap_elder_count {
-                    warn!("Proposed SAP elder count is LESS than current...\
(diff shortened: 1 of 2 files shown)


sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -217,47 +217,47 @@ impl SectionTree {
...
-                    )));
-                }
+
+            // Verify the new SAP with the SAP we know and the proof chain provided.
+            if !proof_chain.has_key(&sap.section_key()) {
+                // This case may happen when both the sender and receiver is about to using
+                // a new SAP. The AE-Update was sent before sender switching to use new SAP,
+                // hence it only contains proof_chain covering the old SAP.
+                // When the update arrives after the receiver got switched to use new SAP,
+                // this error will be complained.
```
oetyng, line 243: I'm a bit unsure what this means.

sn_interface/src/network_knowledge/section_tree/mod.rs
```diff
@@ -275,13 +275,6 @@ impl SectionTree {
                 signed_sap.value
             )));
         }
-        // Make sure the proof chain can be trusted,
-        // i.e. check each key is signed by its parent/predecessor key.
-        if !proof_chain.self_verify() {
```
joshuef, line 280: Do we not still need this verification?

sn_interface/src/network_knowledge/sections_dag.rs
```diff
@@ -301,7 +301,9 @@ impl SectionsDAG {
     /// other
     pub fn merge(&mut self, mut other: Self) -> Result<()> {
         if !other.self_verify() {
-            return Err(Error::InvalidSignature);
+            return Err(Error::UntrustedProofChain(format!(
+                "Proof chain failed self verification: {other:?}",
```
joshuef, line 305: Can we add a specific error variant here 🙇

joshuef (changes requested):

sn_interface/src/network_knowledge/sections_dag.rs
maqi: that `UntrustedProofChain` is already bit specific?
or what do you suggest to use?

joshuef: Not to use a string in an existing error, but rather specific error for this situation. (Ideally we should breakdown UntrustedProofChain into the actual varieties it has, but i know that's not really to do with this PR)

PR #1961, opened 2023-01-05
Fault improvements

joshuef:


oetyng (approved):
This is still greek for me, as I have not looked into the workings of fault detection module.
But as you say network is healthier with this, thumbs up!

joshuef:
Aye I think we need more eyes on this part of the network in general, and to get the tests deeper here. But bit by bit. At least now there's been some voting off of something deemed bad, which was not happening before. 🙇

PR #1965, opened 2023-01-06
Fix store only section conns

joshuef:

### DIFF
--- a/sn_node/src/comm/mod.rs
+++ b/sn_node/src/comm/mod.rs
@@ -26,7 +26,7 @@ use sn_interface::{
 
 use dashmap::DashMap;
 use qp2p::{Endpoint, IncomingConnections};
-use std::{net::SocketAddr, sync::Arc, time::Duration};
+use std::{collections::BTreeSet, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::{
     sync::mpsc::{self, Receiver, Sender},
     task,
@@ -88,6 +88,11 @@ impl Comm {
 
     pub(crate) fn socket_addr(&self) -> SocketAddr {
         self.our_endpoint.public_addr()
+    }
+
+    /// Retain only cached connections for passed members set
+    pub(crate) fn retain_only_peers(&mut self, members: BTreeSet<Peer>) {
+        self.sessions.retain(|p, _| members.contains(p));
     }
 
     /// Fake function used as replacement for testing only.
--- a/sn_node/src/comm/mod.rs
+++ b/sn_node/src/comm/mod.rs
@@ -12,7 +12,7 @@ mod peer_session;
 
 use self::{
     link::Link,
-    listener::{ConnectionEvent, MsgListener},
+    listener::MsgListener,
     peer_session::{PeerSession, SendStatus, SendWatcher},
 };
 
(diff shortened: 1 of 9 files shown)


sn_node/src/comm/mod.rs
```diff
@@ -26,53 +26,75 @@ use sn_interface::{
...
+pub(crate) use link::SendToOneError;
+pub(crate) use listener::ConnectionEvent;
+
 // Communication component of the node to interact with other nodes.
 #[derive(Clone)]
 pub(crate) struct Comm {
     pub(crate) our_endpoint: Endpoint,
     msg_listener: MsgListener,
     sessions: Arc<DashMap<Peer, PeerSession>>,
+    /// Pipe to check that we want to store filters (allows us to check node knowledge eg)
```
oetyng, line 44: > store filters
It's not filters we are storing?

oetyng, line 44: Isn't this to check if we want to store _connections_, rather than filters?

sn_node/src/comm/mod.rs
```diff
@@ -284,30 +306,68 @@ impl Comm {
...
+                connection,
+            })
+            .await
+            .map_err(|_| Error::CouldNotSendToConnectionFilter)?;
+
+        Ok(Some(session))
+    }
+
+    /// Check if a session exists to add connection to
+    /// or passes connection even on to be filtered for relevance before
```
oetyng, line 338: `even` +t

sn_node/src/comm/mod.rs
```diff
@@ -385,11 +447,31 @@ impl Comm {
...
+                        crate::node::Error::CouldNotSendToConnectionFilter
+                    );
+                }
+            }
+            ConnectionEvent::ConnectionClosed { peer, connection } => {
+                comm.remove_conn(&peer, connection).await;
+            }
+        }
+    }
+}
```
oetyng, line 467: missing line break between blocks

oetyng (approved):
Seems like it does its job :+1:.
But I wonder if this whole connection thing could be made simpler somehow. Lots of conditionals and jumping between channels now.

sn_node/src/node/node_starter.rs
```diff
@@ -183,6 +185,29 @@ async fn bootstrap_node(
     )
     .await;
 
+    let filter_node = node.clone();
+
+    // Spawn a task to filter incoming connections and ensure that we only cache our section's connections
```
bochaco, line 190: I may be getting confused, but why are we having this task to filter connections if we are keeping the `Comm` up to date with the set of peers we should keep connections for when calling `retain_only_peers` ? can't the `Comm` perform the filtering already without needing to send it over this channel assuming the node will keep it up to date by calling that function?
I'm also thinking that trying to read the list of network knowledge members for each new conn is a expensive, so if the above makes sense, the `Comm` would be updated with members only when they change, and then it can filter the conns based on the se of members set with `retain_only_peers` ....? ....I know I could be very wrong here and not fully understanding how these two new channels actually work

oetyng:
This test seems to be broken:
`comm::tests::successful_send`
I re-ran it since ci seemed to have timeouts in other places today, but same result, so it seems to be something real there.

sn_node/src/node/node_starter.rs
oetyng: Yeah, I think I agree there @bochaco. I see the same. Sessions are tightly kept up to date, so should be possible to filter it directly based on that. But we were adding new sessions for every new connection that came in.

What we could do, for example, is to keep a separate members list, which we only modify via membership (add when joined, remove when lost). Incoming connections would simply look at that internal list. No channels and faff.

oetyng: Raised a [PR](https://github.com/maidsafe/safe_network/pull/1966) with the suggested approach.

joshuef: I did start in on that road. It leads to a `RwLock` around the comm's members. It mayyy still be preferable to this middleware setup. But aye, comms is cloneable, so any member list must be thread safe. (which, also the retain members wont be doing this properly either...)

joshuef:
Closing this as some lock on members here looks necessary, so we may as well go down that simpler path

PR #1966, opened 2023-01-06
feat(comm): only cache conns of members

oetyng:
- Updates comms on member changes.
- Only adds connections to local cache if from a member.

Supersedes #1965.


joshuef (approved):
bors r+

PR #1971, opened 2023-01-09
Ae probe with bidi

joshuef:

### DIFF
--- a/sn_node/src/node/error.rs
+++ b/sn_node/src/node/error.rs
@@ -33,6 +33,9 @@ pub enum Error {
     /// Client messages always need a response, so should come in on bidi streams
     #[error("No client response stream. Client message came in on a unidirectional stream. All client messages should be bidirectional.")]
     NoClientResponseStream,
+    /// Client messages always need a response, so should come in on bidi streams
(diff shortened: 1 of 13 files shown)


sn_node/src/node/error.rs
```diff
@@ -33,6 +33,12 @@ pub enum Error {
     /// Client messages always need a response, so should come in on bidi streams
     #[error("No client response stream. Client message came in on a unidirectional stream. All client messages should be bidirectional.")]
     NoClientResponseStream,
+    /// We avoid creating connections to entities not in the network. Any nodes/clients etc should initiate bi-di msg channels
```
maqi, line 39: Just a confirmation that for the Relocation node, the second JoinAsRelocatedRequest will be sent using the new name via normal send NodeMsg (not a bi-di channel as I understand?), which at that moment not in the network yet.
Will that mean the elders won't be able to respond then?
If so, could be a headache to refactor.

joshuef: >  Will that mean the elders won't be able to respond then?
This would mean that yep.

But are they responding directly, or after some membership? As post membership, if the node is in the section it's fine (that's the normal join flow there)

joshuef:
Closing, as I think a more general join refactor (integrating it into the join flow) will smooth out a lot of this out and be a  cleaner solution

PR #1974, opened 2023-01-10
fix(sn_node): resolve issues cause the relocation failure

maqi:


sn_interface/src/network_knowledge/node_state.rs
```diff
@@ -147,12 +151,14 @@ impl NodeState {
             }
```
RolandSherwin, line 151: During the validation of the relocation proposal, we check if our prefix matches the prefix of the destination. But this would fail if we want to relocate a node across sections?

PR #1976, opened 2023-01-11
Re-revert commit to check for NAT nodes

b-zee:
This reverts commit c46bb9934d7c12881dcac887ae55fe796027525d.


b-zee:
Closing as currently the address that is sent with the join request will usually be the local socket address (e.g. `0.0.0.0`) and thus offers no reliable way for a node to see if the node is reachable or behind NAT.

PR #1979, opened 2023-01-12
Require total participation for handover requests

grumbach:
An attempt to fix DKG selfish termination by requiring total participation before going to Handover.


joshuef (approved):
bors r+

PR #1980, opened 2023-01-12
Chore track failed proposals

joshuef:

### DIFF
--- a/sn_interface/src/messaging/signature_aggregator.rs
+++ b/sn_interface/src/messaging/signature_aggregator.rs
@@ -8,6 +8,7 @@
 
 use crate::messaging::system::{SectionSig, SectionSigShare};
 use crate::types::keys::ed25519::Digest256;
+use crate::types::Peer;
 use std::collections::{BTreeMap, BTreeSet};
 use thiserror::Error;
 use tiny_keccak::{Hasher, Sha3};
@@ -26,7 +27,7 @@ use tiny_keccak::{Hasher, Sha3};
 #[derive(Debug, Default)]
 pub struct SignatureAggregator {
     /// a map of the hash(payload + bls pubkey) to the signature shares
-    map: BTreeMap<Digest256, BTreeSet<SectionSigShare>>,
+    map: BTreeMap<Digest256, BTreeSet<(Peer, SectionSigShare)>>,
 }
 
 /// AggregatorErrors returned from `SignatureAggregator::add`.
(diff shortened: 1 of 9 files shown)


joshuef:
bors try

sn_interface/src/messaging/signature_aggregator.rs
```diff
@@ -69,17 +71,19 @@ impl SignatureAggregator {
 
         // save the sig share
         let current_shares = self.map.entry(hash).or_default();
-        let _ = current_shares.insert(sig_share.clone());
+        let _ = current_shares.insert((peer, sig_share.clone()));
 
+        // TODO: Should we test for peers sending many shares?
```
grumbach, line 76: I think bls signatures are deterministic unlike ed sigs so if a Peer votes multiple times for the same thing it would result in the same signature. 
If that's the case then NO we don't need such a check.

sn_node/src/node/mod.rs
```diff
@@ -280,6 +285,39 @@ mod core {
             self.info().name()
         }
 
+        /// Given a set of elders that aggregated a proposal, we check our current knowledge and trigger
+        /// fault logging for any nodes that had failed to propse.
+        ///
+        /// This should _over time_ average out and weed out any elderrs that are consistently failing
+        /// to be involved in decisions
+        pub(crate) fn log_proposal_fault_for_any_missing_elders(
```
grumbach, line 293: Say we have 5 elders in Asia and the others in South America, the ones in Asia will most probably get each other's messages before the ones from South America for every vote. Eventually the ones in South America will get kicked out because their messages arrive later, not because they are slow or unreliable, but just because they are slightly further away. Wouldn't that be too punitive?

joshuef: If it's consistently so, yep. But given xorspace isn't geographically mapped, it should be _very_ unlikely for this to occur? (At the moment the fault tracking is still time blocked and faults expire over time too).

joshuef: And enough elders have to agree over this time 🤔 ... Not sure if im being handwavy here, or it is _really_ unlikely.

joshuef: handwavy. even if super unlikely, when it does happen, this is what would happen. good point @grumbach . Hmmm

joshuef:
Closing this as the result is not the intention. We're penalising slightly slower nodes and the unintended consequences here could be centralisation. There's a better way to do this.

PR #1983, opened 2023-01-15
chore(node): remove unused priority

joshuef:


grumbach (approved):
LGTM, I've always wondered how we used priority, I guess this answers my question! 😄

PR #1984, opened 2023-01-15
Feat track and untrack proposal knowledge

joshuef:

### DIFF
--- a/sn_node/src/node/messaging/node_msgs.rs
+++ b/sn_node/src/node/messaging/node_msgs.rs
@@ -352,7 +352,7 @@ impl MyNode {
                     sender,
                     msg_id
                 );
-                node.untrack_node_issue(sender.name(), IssueType::Knowledge);
+                node.untrack_node_issue(sender.name(), IssueType::NetworkKnowledge);
                 node.handle_section_state_proposal(msg_id, proposal, sig_share, sender)
             }
             NodeMsg::DkgStart(session_id, elder_sig) => {
--- a/sn_node/src/node/messaging/node_msgs.rs
+++ b/sn_node/src/node/messaging/node_msgs.rs
@@ -352,7 +352,7 @@ impl MyNode {
                     sender,
                     msg_id
                 );
-                node.untrack_node_issue(sender.name(), IssueType::NetworkKnowledge);
+                node.untrack_node_issue(sender.name(), IssueType::ElderVoting);
                 node.handle_section_state_proposal(msg_id, proposal, sig_share, sender)
             }
             NodeMsg::DkgStart(session_id, elder_sig) => {
--- a/sn_node/src/node/messaging/node_msgs.rs
+++ b/sn_node/src/node/messaging/node_msgs.rs
@@ -338,7 +338,7 @@ impl MyNode {
                 sig_share,
             } => {
                 let mut node = node.write().await;
-                debug!("[NODE WRITE]: ProposeSectionState write gottt...");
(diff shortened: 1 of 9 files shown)


sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -276,6 +276,7 @@ impl MyNode {
             NodeMsg::HandoverVotes(votes) => {
                 let mut node = node.write().await;
                 debug!("[NODE WRITE]: handover votes write gottt...");
+                node.untrack_node_issue(sender.name(), IssueType::Knowledge);
```
grumbach, line 279: Why untrack here in Handover?

sn_node/src/node/messaging/section_state.rs
```diff
@@ -52,6 +53,13 @@ impl MyNode {
         // broadcast the proposal to the recipients
         let mut cmds = vec![];
         let (other_peers, myself) = self.split_peers_and_self(recipients);
+
+        for node in &other_peers {
+            // log a knowledge issue for each that we're proposing
+            // when they vote, this will be untracked
+            self.log_node_issue(node.name(), IssueType::Knowledge);
```
grumbach, line 60: Wouldn't SectionStateProposal be better than Knowledge for the IssueType?

grumbach (commented):
I like this debt collection idea. Couple of questions here and there.

sn_node/src/node/messaging/section_state.rs
joshuef: So we already have knowledge, I'm not sure if we need a new tracker?

Currrently, we add a knowledge issue if AE is out of sync. So this could fall in that same bucket? (We'd only want a different bucket if the weight of the issue were to change)

sn_node/src/node/messaging/node_msgs.rs
joshuef: Ah, I was thinking to track handover votes in a similar fashion to section proposals, but have clearly forgotten to add the initial track :| 

Perhaps that doesn't make sense though?

sn_node/src/node/mod.rs
```diff
@@ -390,11 +390,11 @@ mod core {
                 // So, shall only track the side that we are in as well.
                 if zero_dkg_id.elders.contains_key(&self.info().name()) {
                     for candidate in zero_dkg_id.elders.keys() {
-                        self.log_node_issue(*candidate, IssueType::Dkg);
+                        self.track_node_issue(*candidate, IssueType::Dkg);
                     }
                 } else if one_dkg_id.elders.contains_key(&self.info().name()) {
                     for candidate in one_dkg_id.elders.keys() {
-                        self.log_node_issue(*candidate, IssueType::Dkg);
+                        self.track_node_issue(*candidate, IssueType::Dkg);
```
grumbach, line 397: We are tracking ourselves too here?

sn_node/src/node/mod.rs
```diff
@@ -447,7 +447,7 @@ mod core {
                 };
                 // track init of DKG
                 for candidate in session_id.elders.keys() {
-                    self.log_node_issue(*candidate, IssueType::Dkg);
+                    self.track_node_issue(*candidate, IssueType::Dkg);
```
grumbach, line 450: Here too

sn_node/src/node/mod.rs
joshuef: pushed a commit to avoid that inside of `track_node_issue`

grumbach (approved):
LGTM

PR #1986, opened 2023-01-16
refactor(node): pass NodeContext within Cmd::UpdateNetworkAndHandleValidClientMsg

bochaco:
This also checks if the network knowledge provided in a `SpentbookCmd::Spend` cmd has any update against the `NodeContext`, before trying to update the actual node's network knowledge by triggering the internal `Cmd::UpdateNetworkAndHandleValidClientMsg`.

### DIFF
--- a/sn_node/src/node/messaging/client_msgs.rs
+++ b/sn_node/src/node/messaging/client_msgs.rs
@@ -24,7 +24,7 @@ use sn_interface::{
         system::{NodeDataResponse, OperationId},
         AuthorityProof, ClientAuth, MsgId,
     },
-    network_knowledge::section_keys::build_spent_proof_share,
+    network_knowledge::{section_keys::build_spent_proof_share, SectionTreeUpdate},
     types::{
         log_markers::LogMarker,
         register::{Permissions, Policy, Register, User},
@@ -124,7 +124,7 @@ impl MyNode {
     /// the `data_copy_count()` nodes, then we will send a wiremsg
     /// to ourselves, among the msgs sent to the other holders.
     pub(crate) async fn handle_valid_client_msg(
-        context: NodeContext,
+        mut context: NodeContext,
         msg_id: MsgId,
         msg: ClientMsg,
         auth: AuthorityProof<ClientAuth>,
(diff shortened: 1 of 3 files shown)


sn_node/src/node/messaging/client_msgs.rs
```diff
@@ -167,25 +167,35 @@ impl MyNode {
...
-                        signed_sap,
-                        msg_id,
-                        msg: updated_client_msg,
-                        origin,
-                        send_stream,
-                        auth,
-                    };
-                    return Ok(vec![update_command]);
+                    let name = context.name;
+                    let there_was_an_update = context.network_knowledge.update_knowledge_if_valid(
```
joshuef, line 171: we're reusing the `context` below, but we've modified it here. 

It looks like in the update+handle func we don't need the changed info atm, but our context is not what we started with. I think it's cleaner to do a context clone and run `update_knowledge_if_valid` on _that_ so our context remains unchanged as we pass it on. (that way there's no confusion down the line on the node state vs that in the context)

joshuef (changes requested):

sn_node/src/node/messaging/client_msgs.rs
joshuef:
```suggestion
                    let there_was_an_update = context.clone().network_knowledge.update_knowledge_if_valid(
```

joshuef: the `mut` above on context probs needs remove there too.

bochaco: I think we actually want the context changed, the msg being handled carries an update for our node which is meant to be required to properly process the msg, so wee need to update our node's network knowledge to the process it, therefore if we didn't have the context we would be update the node's state and then process the msg atop the new state. Alternatively if we cloned the context to check if there is an update, we would want to get a new context instance out of the node's state after we apply the update. The context and node's state is being updated as part o the msg handling process, so I don't think it matters what was the initial state/context but what the handling process should be work on.....?

joshuef (approved):
bors r+

PR #1988, opened 2023-01-17
fix(relocation): skip the propose NodeIsOffline flow for relocations.

davidrusu:
Currently node relocations first go through a NodeIsOffline proposal flow. This flow was originally intended to check that enough elders agree that a node is faulty, i.e. detecting network connection errors, etc.

It's not needed for relocations since the relocation details has all the information we need to verify if a relocation request is valid.

This PR drops the NodeIsOffline proposal in favor of simply triggering the membership change directly.

This should allow us to more easily retry failed relocations


davidrusu:
bors try

oetyng (approved):

PR #1993, opened 2023-01-18
Misc Release Process Updates

jacderida:
- 2badf68d6 **ci: add arm builds to merge workflow**

  We seen a failure on a release because an ARM build failed. The failure occurred because the `cross`
  utility (which we use for cross compilation) wouldn't build correctly. Strangely, I wasn't able to
  reproduce this problem, but I think it highlights the need for building on ARM before we merge. If
  we are going to release on every merge, we will need to check that we can build on ARM.
  
- bfc96acdc **ci: slack notifications for workflow failures**

  We didn't become aware of our latest release workflow failure for weeks after the event, so I've
  enabled notifications for failures of either the version bumping or release workflows.
  
- d14f34bde **ci: add sn_comms crate to the release process**

  This new crate is added to all the necessary mechanisms for the release process.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -11,7 +11,7 @@ concurrency:
   group: "main-modifying-workflow"
 
 env:
-  CARGO_INCREMENTAL: 0  # bookkeeping for incremental builds has overhead, not useful in CI.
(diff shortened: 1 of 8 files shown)


joshuef (approved):
bors r+

PR #1977, opened 2023-01-11
test(sn_node): relocation + membership tests

RolandSherwin:
Depends on https://github.com/maidsafe/safe_network/pull/1982

PR #1998, opened 2023-01-19
fix(sn_node): fix a deadlock during relocation

RolandSherwin:
Fixes a deadlock that causes test failures in https://github.com/maidsafe/safe_network/pull/1977


joshuef (approved):
bors r+

PR #2002, opened 2023-01-19
chore(node): remove readlock where we have context

joshuef:


oetyng (approved):
One thing that hit me, as I read first file change, before I scrolled down to second one, is that it could be unclear what on the context is possible to mutate and what is not as it is a snapshot.
Any channels of course would still send it over to mutating parts. And that's fine, all good. Just a thought that hit me about clarity wrt the context not being only a snapshot (it doesn't have to be only that).

joshuef:
aye that's a fair Q @oetyng . Not sure how best to address that, but something we can think on 👍

PR #2004, opened 2023-01-19
fix(relocation): always run the DKG checks on membership decision

davidrusu:

### DIFF
--- a/sn_node/src/node/messaging/membership.rs
+++ b/sn_node/src/node/messaging/membership.rs
@@ -18,7 +18,6 @@ use crate::node::{
 use bls::Signature;
 use sn_consensus::{Decision, Generation, SignedVote, VoteResponse};
 use sn_interface::{
-    elder_count,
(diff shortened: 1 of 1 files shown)


davidrusu:
bors try

sn_node/src/node/messaging/membership.rs
```diff
@@ -223,17 +222,7 @@ impl MyNode {
...
-        let we_have_enough_elders = self.network_knowledge.elders().len() == elder_count();
-        let everything_joined_was_relocated = joining_nodes
-            .iter()
-            .all(|(node_state, _sig)| node_state.previous_name().is_some());
-        let we_should_look_for_new_elders =
-            !(we_have_enough_elders && everything_joined_was_relocated);
-        if we_should_look_for_new_elders {
-            cmds.extend(self.trigger_dkg()?);
-        }
+        cmds.extend(self.trigger_dkg()?);
```
maqi, line 225: The change will trigger DKG even having enough elders, will that be a bit waste?

joshuef (approved):
bors r+

PR #2008, opened 2023-01-20
ci: add check for expected number of elders after startup

joshuef:

### DIFF
--- a/sn_comms/src/lib.rs
+++ b/sn_comms/src/lib.rs
@@ -190,49 +190,28 @@ impl Comm {
         msg_id: MsgId,
         bytes: UsrMsgBytes,
     ) -> Result<()> {
-        let watcher = self.send_to_one(peer, msg_id, bytes).await;
+        let watcher = self.send_to_one(peer, msg_id, bytes).await?;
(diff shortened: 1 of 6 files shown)


joshuef:
bors try

sn_comms/src/lib.rs
```diff
@@ -190,49 +190,28 @@ impl Comm {
...
-                    } else {
-                        if should_remove {
-                            // do cleanup of that peer
-                            let perhaps_session = sessions.remove(&peer);
-                            if let Some((_peer, session)) = perhaps_session {
-                                session.disconnect().await;
-                            }
-                        }
-                        Err(Error::FailedSend(peer))
+            let (send_was_successful, should_remove) = Self::is_sent(watcher, msg_id, peer).await?;
```
RolandSherwin, line 201: Not directly related, but the `Result` in the `is_sent` method is redundant and can be removed..

joshuef: sorted

RolandSherwin (approved):
The changes look good to me!

PR #2009, opened 2023-01-20
No incoming conn cached

joshuef:
follow up to #2008


sn_comms/src/listener.rs
```diff
@@ -144,9 +129,9 @@ impl MsgListener {
 
         trace!(%conn_id, %remote_address, "{}", LogMarker::ConnectionClosed);
 
-        // if the connection was from a (non client) peer was cached, we shall remove it
-        if let Some(peer) = node_conn_cached {
-            trace!("Removing connection {conn_id} with node {peer} from cache");
+        // remove this closed connection
+        if let Some(peer) = established_peer {
+            trace!("Removing connection {conn_id} with node {established_peer:?} from cache");
```
bochaco, line 134: I'm not understanding how we could remove a conn from a peer if we never add such a conn to the cache?

bochaco: ...it sounds like we assume connections will be kept in memory till we drop the `send_stream` we've sent on `MsgFromPeer` event, but we don't have the connection in our cache to then remove it later on...?...

sn_comms/src/listener.rs
```diff
@@ -83,15 +80,12 @@ impl MsgListener {
                             continue;
                         }
                     };
-                    let mut is_from_client = false;
-                    let mut is_node_join_msg = false;
+
+                    // let mut is_node_join_msg = false;
```
bochaco, line 84: let's remove this

sn_comms/src/lib.rs
```diff
@@ -267,7 +245,7 @@ impl Comm {
                 peer.name(),
                 error,
             );
-                let _peer = self.sessions.remove(&peer);
+                let _peer = self.outgoing_sessions.remove(&peer);
```
bochaco, line 248: Shouldn't we use `remove_conn` every time we try to remove a session otherwise we are removing all connections to that peer?

sn_comms/src/lib.rs
```diff
@@ -178,7 +176,8 @@ impl Comm {
     pub async fn update_members(&mut self, members: BTreeSet<Peer>) {
         let new_members = members.clone();
         *self.members.write().await = members;
-        self.sessions.retain(|p, _| new_members.contains(p));
+        self.outgoing_sessions
+            .retain(|p, _| new_members.contains(p));
```
bochaco, line 180: perhaps we want to remove sessions to peers which are not members anymore ?

sn_comms/src/listener.rs
joshuef: It could have been a member previously, and so perhaps should be cleared

sn_comms/src/lib.rs
```diff
@@ -190,49 +189,28 @@ impl Comm {
...
-                    warn!("Accessed a disconnected peer: {peer}. This is potentially a bug!");
-
-                    let _peer = sessions.remove(&peer);
-                    error!(
-                        "Sending message (msg_id: {msg_id:?}) to {peer} failed \
-                        as we have disconnected from the peer. (Error is: {error})",
-                    );
-                    Err(Error::FailedSend(peer))
-                }
+                Err(Error::FailedSend(peer))
```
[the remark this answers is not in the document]
joshuef: We do use it in `node_msgs.rs`: https://github.com/maidsafe/safe_network/blob/main/sn_node/src/node/messaging/node_msgs.rs#L63 to track faults

sn_comms/src/listener.rs
joshuef: Ah sorry, misread this. Aye, cos we never add anything we're listening to to the cache, this is not needed. 👍

joshuef:
bors try

joshuef:
bors try

joshuef:
that bors try fail is due to differing replica returns, which is out of scope for this PR

sn_comms/src/lib.rs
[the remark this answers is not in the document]
bochaco: You are not getting the `Result` from the task spawned by `send_out_bytes`, that will be lost/thrown-away, the `send_out_bytes` function is returning `Ok(())` right after it spawned the task, it doesn't return what the spawned task results in. The only errors you can obtain is those returned by the call `send_out_bytes` does to `send_to_one`.

bochaco (approved):

bochaco:
bors merge

PR #2010, opened 2023-01-20
fix(comm): add relocating peer to `Comm`

RolandSherwin:


sn_node/src/node/messaging/joining_nodes.rs
```diff
@@ -103,6 +103,13 @@ impl MyNode {
             return None;
         };
 
+        // Add the peer to our Comm to facilitate communication
+        node.write()
```
joshuef, line 107: MIssing a note of `NODE WRITE` here

sn_comms/src/lib.rs
```diff
@@ -174,11 +175,28 @@ impl Comm {
         self.our_endpoint.close()
     }
 
-    /// Updates cached connections for passed members set only.
-    pub async fn update_members(&mut self, members: BTreeSet<Peer>) {
-        let new_members = members.clone();
-        *self.members.write().await = members;
-        self.sessions.retain(|p, _| new_members.contains(p));
+    /// Updates cached connections for the current members and the members undergoing relocation
+    pub async fn update_members(&mut self, network_knowledge: &NetworkKnowledge) {
```
joshuef, line 179: I think we're crossing crate boundaries here. 

This is node world stuff, which members are valid and communicable.

If we need a wrapper func on comms call to do this thats fine (and something we can add tests to actually)

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -33,6 +33,11 @@ impl SectionPeers {
         self.members.values().cloned().collect()
     }
 
+    /// Returns set of archived members, i.e those that've left our section
+    pub(super) fn archived_members(&self) -> BTreeSet<SectionSigned<NodeState>> {
+        self.archive.values().cloned().collect()
```
joshuef, line 38: are all archived members valid forever?? This seems like a heavy handed solution. Can we not just add the actual node in question to the comms members?

joshuef (changes requested):

sn_interface/src/network_knowledge/section_peers.rs
RolandSherwin: Ah! I was planning to add the details about this.. Any members that are older than 5 (`ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE`) churn events are removed from the list.. 
Maybe this is too high of a number still? If so, I will make the changes to add just the node that we want..

joshuef: yeh if elders dont churn for a whiiile this could be a big list

RolandSherwin:
Closing due to breaking changes to relocation.

PR #2016, opened 2023-01-22
chore(node): only clone membership when needed

joshuef:
this should vastly reduce allocations coming from membership.clone() (which accounts for the vast majority of alloc in a runnning node just now

### DIFF
--- a/sn_node/src/node/flow_ctrl/periodic_checks.rs
+++ b/sn_node/src/node/flow_ctrl/periodic_checks.rs
@@ -57,10 +57,15 @@ impl FlowCtrl {
     pub(super) async fn perform_periodic_checks(&mut self) {
         let (context, membership_context) = {
             let read_locked_node = self.node.read().await;
-            (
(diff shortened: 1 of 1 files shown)


sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -57,10 +57,15 @@ impl FlowCtrl {
...
+            let membership = if self.timestamps.last_vote_check.elapsed() > MISSING_VOTE_INTERVAL {
+                read_locked_node.membership.clone()
+            } else {
+                None
+            };
+
+            (&read_locked_node.context(), membership)
         };
 
         if !context.is_elder {
```
RolandSherwin, line 71: I assume that the adult periodic check is still a WIP, so maybe we can optimize this further once it lands?

RolandSherwin (approved):
PR looks good to me, assuming that the adult check will be enabled soon..

Issue #2011, opened 2023-01-20
Make relocation details verifiable.


davidrusu:
Nodes are relocated between sections based on a `ChurnId`. These churn Id's decide two things:
1. which nodes will be relocated to a new section.
2. which section the node will be relocated to

So if they are forgeable, a faulty elder could arbitrarily force any node to relocate to a section of their choice. We need to ensure that this can't happen.

Currently the NodeState struct looks like this:

```
pub struct NodeState {
    peer: Peer,
    state: MembershipState,
    previous_name: Option<XorName>,
}

pub enum MembershipState {
    Joined,
    Left,
    Relocated(RelocationDst), // <-- what we are looking at in this issue
}

pub struct RelocationDst(XorName)
```

We need to replace the RelocationDst with something verifiable, like a RelocationTrigger that holds the verifiable event that triggered this relocation.

The end state would look something like this:

```
pub struct NodeState {
    peer: Peer,
    state: MembershipState,
    previous_name: Option<XorName>,
}

pub enum MembershipState {
    Joined,
    Left,
    Relocated(RelocationTrigger>),
}

pub struct RelocateTrigger(Decision<NodeState>)

impl RelocateTrigger {
    fn dst(&self) -> XorName; // derived from signed node state
}
```

b00f:
Without going more into details, It looks to me that `NodeState` **has** a  `MembershipState` and `MembershipState` **has** a `NodeState`. This kind of  circular references (Foo has Bar and Bar has Foo ) might cause some difficulties in future. 
Don't you think we can assign some behavior to the  `NodeState` for `XorName` and `verification`? Also we can decouple the `Decision` from `RelocateTrigger`.

PR #2018, opened 2023-01-22
refactor(relocation): use existing join flow

oetyng:
This PR precedes a PR addressing [this issue](https://github.com/maidsafe/safe_network/issues/2011).

Relocation was previously a separate request/response flow.
That flow is now removed with this PR.

Now we use the existing join flow, with an optional parameter of a `RelocationProof`.
This allows the relocation to use the same `ae` flow that already was implemented for join.

Additional changes to note:
- Only adults are relocated.
- The source section no longer increments the age in their vote for `Relocated` of the relocating node.
Instead, the joining node increments its own age, as it generates a new name after it receives the membership decision. It then tries to join the dst section with this new name including the incremented age. The dst section validates that the joiner age is one higher than its previous age (as given by its previous name).


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -332,6 +345,19 @@ impl NetworkKnowledge {
...
         &self,
         name: &XorName,
+    ) -> Option<SectionSigned<SectionAuthorityProvider>> {
+        let closest_sap = self
+            .section_tree
+            .closest(name, Some(&self.prefix()))
+            // In case the only prefix is ours, we fallback to it.
+            .unwrap_or(self.section_tree.get_signed(&self.prefix())?);
+        Some(closest_sap.clone())
+    }
```
davidrusu, line 355: Use this helper below in `closest_signed_sap_with_chain()` to get the closest sap, i.e.:

```rust
    pub fn closest_signed_sap_with_chain(
        &self,
        name: &XorName,
    ) -> Option<(&SectionSigned<SectionAuthorityProvider>, SectionsDAG)> {
        let closest_sap = self.closest_signed_sap()?;
        ...
    }
```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -199,12 +200,24 @@ impl NetworkKnowledge {
...
+
     /// update all section info for our new section
     pub fn relocate_to(&mut self, new_name: XorName) -> Result<()> {
         let signed_sap = self.section_tree().get_signed_by_name(&new_name)?;
         debug!("Node was relocated to {:?}", signed_sap.section_key());
         self.signed_sap = signed_sap;
-
         Ok(())
     }
```
davidrusu, line 222: These two functions are doing very similar jobs, do we need both?

oetyng: No, the below shouldn't be used, let me check why that is there.

oetyng: ah, part of the crate public interface, that's why. removing it now

sn_interface/src/network_knowledge/node_state.rs
```diff
@@ -200,60 +181,89 @@ impl NodeState {
...
-        let dst_section_key = network_knowledge
-            .section_auth_by_name(&dst)
-            .map_or_else(|_| genesis_key, |section_auth| section_auth.section_key());
+#[derive(Clone, PartialEq, Serialize, Deserialize, custom_debug::Debug)]
+pub struct RelocationProof {
+    info: RelocationInfo,
+    #[serde(with = "serde_bytes")]
+    #[debug(with = "Self::fmt_ed25519")]
+    peer_sig: Signature,
+    peer_key: PublicKey,
```
davidrusu, line 217: It's unclear who's signature this is. Is this the node signing it's own relocation proof?

If so, how about we call it `node_sig`.

And, we can drop the `peer_key` since it should be derivable from `info.signed_relocation.value.name`. (and removes a validation check).

But I may be missing something here.

sn_interface/src/network_knowledge/node_state.rs
```diff
@@ -200,60 +181,89 @@ impl NodeState {
...
+pub struct RelocationDst(XorName);
+
+impl RelocationDst {
+    pub fn new(name: XorName) -> Self {
+        Self(name)
+    }
+
+    pub fn name(&self) -> &XorName {
+        &self.0
+    }
```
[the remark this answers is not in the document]
oetyng: Yeah, I agree, good point. Will fix!

sn_interface/src/network_knowledge/node_state.rs
oetyng: We can clarify the naming there I think.
But the key is needed for verifying the sig.
The XorName can be derived from the key, but not the other way around, so we can't verify the sig with the XorName.

oetyng: The relocating node is signing it's new name, but a sig over an xorname could be anything, it's missing context, so I thought it OK to sign the entire section signed relocated state, as to be sure that it was a relocation. Probably not entirely necessary, since it would be hard to misuse, but I thought it a good practice anyway to include context in a sig.

oetyng: Node/Peer btw is something I've since long been wondering if there is any distinction.

oetyng: I will add some more comments here.

sn_interface/src/network_knowledge/node_state.rs
```diff
@@ -200,60 +181,107 @@ impl NodeState {
...
+        let serialized_state = bincode::serialize(&self.info.signed_relocation.value)
+            .map_err(|_err| Error::InvalidSignature)?;
+        if !self.info.signed_relocation.sig.verify(&serialized_state) {
+            Err(Error::InvalidSignature)
         } else {
-            return false;
-        };
+            Ok(())
+        }
+    }
```
davidrusu, line 271: We need to also check that `the self.info.signed_relocation.name == XorName::from(self.self_old_key)`

oetyng: Good catch!

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -302,17 +269,16 @@ impl MyNode {
...
-                        debug!("[NODE READ]: relocation_state read got");
-                        if relocate_state_present {
-                            let mut node = node.write().await;
-                            debug!("[NODE WRITE]: relocation_state write got...");
-                            node.relocate_state = None;
+                        if decision
+                            .proposals
+                            .keys()
+                            .filter(|n| n.state() == MembershipState::Joined)
+                            .all(|n| n.previous_name().is_some())
```
davidrusu, line 280: Do you mean `any()` here? As in, if there was a node we accepted who had a previous name, then there was a relocation that completed.

```suggestion
                            .any(|n| n.previous_name().is_some())
```

oetyng: Yeah IDK, I reused that bit of the code.

oetyng: Yeah, it's from ln 265.

davidrusu (approved):
Looks great 🚀

PR #2021, opened 2023-01-23
fix(comms): terminate PeerSessionWorker when dropping a PeerSession

bochaco:
- Minor refactoring, and removing unnecessary private `SessionStatus` type.
- Making sure we invoke `PeerSession::disconnect` on any session being dropped in order to not leak `PeerSessionWorker` tasks.


bochaco:
bors try

PR #2022, opened 2023-01-23
refactor(comms): removing Comm::members and unnecessary private types

bochaco:
- We now use Comm::sessions as the list of members we keep sessions with, which is updated only when the user wants to update the set of known peers.
- Removing unnecessary PeerSession's SessionStatus, disconnect fn, and SessionCmd.
- Never remove sessions from Comm::sessions unless the set of known members is updated/changed by the user. Even if we failed to send using all peer session's connections, we keep the session since it's been set as a known and connectable peer.

### DIFF
--- a/sn_comms/src/lib.rs
+++ b/sn_comms/src/lib.rs
@@ -58,14 +58,14 @@ use self::{
 
 use sn_interface::{
     messaging::{MsgId, WireMsg},
-    types::{log_markers::LogMarker, Peer},
+    types::Peer,
 };
 
-use qp2p::{Endpoint, IncomingConnections, SendStream, UsrMsgBytes};
+use qp2p::{Endpoint, SendStream, UsrMsgBytes};
 
 use dashmap::DashMap;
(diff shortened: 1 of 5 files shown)


bochaco:
bors try

joshuef (approved):
nice tidy up

bochaco:
bors try

sn_comms/src/lib.rs
```diff
@@ -166,10 +161,19 @@ impl Comm {
 
     /// Updates cached connections for passed members set only.
     pub async fn update_valid_comm_targets(&mut self, members: BTreeSet<Peer>) {
```
oetyng, line 163: async can be removed here

oetyng: I'm going to raise a PR fixing stuff right on top of this, so I can take it there if you want

bochaco:
bors try

bochaco:
bors try

bochaco:
bors merge

PR #2024, opened 2023-01-24
feat(relocation): add retry for relocating node

oetyng:
When a node is joining with a relocation proof, they could be rejected by one reason only: `AddressNotReachable`. This is because the other reject reason, `JoinsDisallowed`, does not apply to relocated nodes, making this state `unreachable`.

Address not reachable is an irrecoverable error, and therefore a node has to change their port configuration before starting a node again. However, this specific issue should be validated long before a relocation, by which this state is also considered `unreachable`.

Beyond those, and tampered proofs, it's just regular timeouts. 

So this feature allows a joining node with a relocation proof, to retry when the join times out.

### DIFF
--- a/sn_node/src/node/flow_ctrl/periodic_checks.rs
+++ b/sn_node/src/node/flow_ctrl/periodic_checks.rs
@@ -19,6 +19,7 @@ use std::{collections::BTreeSet, sync::Arc, time::Duration};
 use tokio::{sync::RwLock, time::Instant};
 
 const PROBE_INTERVAL: Duration = Duration::from_secs(300);
+const RELOCATION_TIMEOUT_SECS: Duration = Duration::from_secs(60);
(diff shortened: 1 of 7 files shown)


sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -165,6 +166,31 @@ impl FlowCtrl {
             cmds.extend(self.vote_out_faulty_nodes().await);
         }
 
+        // This check keeps relocation retrying if it times out.
+        if let Some((proof, last_try)) = &context.relocation_proof {
```
davidrusu, line 170: Put the `last try` timestamp into the timestamps struct above: `PeriodicChecksTimestamps` so that we follow the same structure as the other periodics.

davidrusu (approved):
Looks good 🚢

PR #2040, opened 2023-01-27
Remove section state vote

oetyng:
This layer of voting is redundant, as membership votes are enough both for voting
offline and setting joins allowed.


grumbach (commented):
I'm always in favour or removing code and flows. I also agree that the sectionstate proposal flows are redundant. 
But membership currently doesn't have the functionality that we removed here :
- refusing to allow joins
- refusing to reject a node

PR #2041, opened 2023-01-27
fix(storage): make data exists not be an error

oetyng:
- The data storage is idempotent, and it's not an error that data already exists.


joshuef (approved):
bors r+

PR #2043, opened 2023-01-28
refactor(comms): remove use of dashmap

oetyng:
This removes dashmap use for sessions in comms.

It also makes comms error handling and issue reporting focused to one place.

Additional thing to note is that Elders now relay responses back to client as soon as they are received.
The way the requests are designed now, that would only be more than one response when storing data,
since queries from a Client, only go out to one data holder from the Elders.

### DIFF
--- a/sn_comms/src/lib.rs
+++ b/sn_comms/src/lib.rs
@@ -50,21 +50,67 @@ mod peer_session;
 
 pub use self::error::{Error, Result};
 
-use self::{listener::MsgListener, peer_session::PeerSession};
+use self::peer_session::PeerSession;
 
 use sn_interface::{
-    messaging::{MsgId, WireMsg},
+    messaging::{
+        data::ClientDataResponse as ClientResponse, system::NodeMsgType, Dst, MsgId, MsgKind,
+        MsgType, WireMsg,
+    },
     types::Peer,
 };
 
 use qp2p::{Endpoint, SendStream, UsrMsgBytes};
 
-use dashmap::DashMap;
-use std::{collections::BTreeSet, net::SocketAddr, sync::Arc};
-use tokio::sync::mpsc::Sender;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    net::SocketAddr,
+};
+use tokio::{
+    sync::mpsc::{self, Receiver, Sender},
+    task,
+};
 
 /// Standard channel size, to allow for large swings in throughput
 static STANDARD_CHANNEL_SIZE: usize = 100_000;
+
+/// Events from the comm module.
+#[derive(Debug)]
+pub enum CommEvent {
+    /// A msg was received.
+    Msg(MsgFromPeer),
+    /// A send error occurred.
+    Error {
+        /// The sender/recipient that failed.
+        peer: Peer,
+        /// The failure type.
+        error: Error,
+    },
+}
+
+/// Internal comm cmds.
+#[derive(custom_debug::Debug)]
+enum CommCmd {
+    Send {
+        msg_id: MsgId,
+        peer: Peer,
+        #[debug(skip)]
+        bytes: UsrMsgBytes,
+    },
+    SetTargets(BTreeSet<Peer>),
+    SendWithBiResponse {
+        peer: Peer,
+        msg_id: MsgId,
+        bytes: UsrMsgBytes,
+    },
+    SendAndRespondOnStream {
+        msg_id: MsgId,
+        msg_type: NodeMsgType,
+        #[debug(skip)]
+        node_bytes: BTreeMap<Peer, UsrMsgBytes>,
+        dst_stream: (Dst, SendStream),
+    },
+}
 
 /// A msg received on the wire.
 #[derive(Debug)]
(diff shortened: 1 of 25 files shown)


joshuef:
> Additional thing to note is that Elders now relay responses back to client as soon as they are received. That also puts the burden on the client to sort out if they have enough responses or not.

Are we able to separate this changeset out into a separate PR please?

sn_node/src/node/messaging/streams.rs
```diff
@@ -139,257 +118,100 @@ impl MyNode {
 
     /// Sends a msg via comms, and listens for any response
     /// The response is returned to be handled via the dispatcher (though a response is not necessarily expected)
-    pub(crate) async fn send_msg_enqueue_any_response(
+    pub(crate) fn send_msg_with_bi_response(
```
joshuef, line 121: originally i made this as it's not guaranteed that we'll see _any_ response here.

I find the name change somewhat confusing. we're sending a response? are we expecting one?

joshuef:
bors try

sn_comms/src/lib.rs
```diff
@@ -157,85 +195,347 @@ impl Comm {
...
-            .await
-            .map_err(|err| {
-                error!("Failed sending {msg_id:?} to {peer:?}: {err:?}");
-                Error::FailedSend(peer)
-            })?;
-        debug!("Peer response from {peer:?} is in for {msg_id:?}");
-        WireMsg::from(adult_response_bytes).map_err(|_| Error::InvalidMessage)
-    }
+#[tracing::instrument(skip_all)]
+fn send_with_bi_response(
```
joshuef, line 380: aye this naming is still not clear to me. (I think some docs would help)

is it still basically? `send_out_bytes_to_peer_and_return_response`? but looking for something shorter? 

`send_and_return_response` perhaps? (that it's `bi` feels like an impl detail)

sn_comms/src/lib.rs
```diff
@@ -157,85 +195,347 @@ impl Comm {
...
+            }
         }
-    }
+    });
+}
+
+/// Verify what kind of response was received, and if that's the expected type based on
+/// the type of msg sent to the nodes, then return the corresponding response to the client.
+#[tracing::instrument(skip_all)]
+fn map_to_client_response(
```
joshuef, line 475: This feels like we're blurring crate boundaries here.

Feels like we're wanting this to respond faster to the client on each/every response on the bidi... and so we don't have to leak the `ResponseStream` out?

(That whole func in node is also ugly, tbh).

If we're wanting to send direct to client... Then I'd say we make the client capabale of handling NodeDataResponse. Or the nodes just send out ClientDataResponse, and we punt it straight on (I'm actually looking at this just now from another angle, but it seems sensible).

The whole mapping just feels superfluous. And if we're just directly resending `WireMsg`, then we loose that and the blurred boundary?

joshuef (changes requested):
Overall I really like this approach for comms.

Some niggly bits would be nice to get sorted w/r/t some naming feeling less clear for me, and that blurry boundary. Would be good to know what you think about the passthrough idea there

sn_node/src/node/messaging/streams.rs
oetyng: I changed to `send_and_return_response`.

sn_comms/src/lib.rs
oetyng: I changed to `send_and_return_response`.

oetyng:
bors try

b-zee:
Looks solid to me. Partially agree with the comment from @joshuef on crate boundaries, though  in general the interface boundaries are better. The session handling seems more solid now, with a clear idea on where sessions are managed.

b-zee (commented):
Looks solid to me. Partially agree with the comment from @joshuef on crate boundaries, though in general the interface boundaries are better. The session handling seems more solid now, with a clear idea on where sessions are managed.

oetyng:
> Looks solid to me. Partially agree with the comment from @joshuef on crate boundaries, though in general the interface boundaries are better. The session handling seems more solid now, with a clear idea on where sessions are managed.

Yep, the crate boundary there was very consciously crossed, because I couldn't get there easily without it. I see it as a temporary thing, seemed like an acceptable trade-off for now, which will get sorted.

joshuef:
>  I see it as a temporary thing, seemed like an acceptable trade-off for now, which will get sorted.

I think we should be able to remove that mapping altogether now we just forward messages on?

sn_node/src/node/messaging/streams.rs
```diff
@@ -113,216 +96,79 @@ impl MyNode {
...
-    let mut wire_msg = WireMsg::new_msg(msg_id, payload, kind, dst);
-    let _bytes = wire_msg.serialize_and_cache_bytes()?;
-
-    let mut send_tasks = vec![];
-    for target in targets {
-        dst.name = target.name();
-        let bytes_to_node = wire_msg.serialize_with_new_dst(&dst)?;
+        context
+            .comm
+            .send_and_respond_on_stream(msg_id, node_bytes, targets_len, dst_stream);
```
b-zee, line 168: The `targets_len` should be equal to the size of the `node_bytes` BTreeMap? If so, then we can deduce the expected amount of responses from that.

sn_comms/src/lib.rs
```diff
@@ -120,85 +155,348 @@ impl Comm {
...
+            }
+        };
+    });
+}
+
+#[tracing::instrument(skip_all)]
+fn send_and_respond_on_stream(
+    msg_id: MsgId,
+    peer_bytes: BTreeMap<Peer, (Option<PeerSession>, UsrMsgBytes)>,
+    expected_targets: usize,
```
b-zee, line 379: This parameter `expected_targets` should always be equal to the `len()` of `peer_bytes`?

sn_node/src/node/messaging/streams.rs
oetyng: not when we filter mapped on serialization in streams.rs

sn_node/src/node/messaging/streams.rs
```diff
@@ -113,216 +96,79 @@ impl MyNode {
...
-            .boxed(),
+        let node_bytes: BTreeMap<_, _> = targets
+            .into_par_iter()
+            .filter_map(|target| {
+                let dst = Dst {
+                    name: target.name(),
+                    section_key: context.network_knowledge.section_key(),
+                };
+                wire_msg
+                    .serialize_with_new_dst(&dst)
```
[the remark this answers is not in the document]
oetyng: it's because we need to send a response to the client anyway, so can just as well filter map here, and in comms that's where same logic will be used to send the error response to client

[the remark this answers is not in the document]
oetyng: i.e. there is no shorter path to send the error to client :)

sn_comms/src/lib.rs
oetyng: nopes, due to filter mapped on serialization in streams.rs

PR #2046, opened 2023-01-30
feat(sn_node): membership vote cache changed votes to recast

maqi:


maqi:
so far closed as there is doubt the cache could result in outdated votes which causes issue.

PR #2045, opened 2023-01-30
feat(node): cache for data storage

bochaco:

PR #1996, opened 2023-01-19
refactor: implement new qp2p version

b-zee:
This introduces quite some changes to the API that hopefully simplifies
much of the internals of the node and client.

PR #2047, opened 2023-01-31
Reduced timeouts with cached storage

b-zee:
This branch is rebased on top of #2045 to see if disk usage is a bottleneck.

The rest of the branch is the same as #1996 except for the timeout which is put at the default value of qp2p, which is 10s. This currently results in CI having trouble to pass some tests where the nodes get sort of overloaded by all the concurrent tests. This is why the timeout is set to higher values to pass these test.


b-zee:
bors try

PR #2052, opened 2023-01-31
refactor: moves serialize call to where needed

oetyng:
- We were serializing before hitting the branch where it was needed.
- Also aligns fn name withy doc. The cases when no cmd is generated are the unreachable and invalid msg cases.


joshuef (approved):
bors r+

PR #2055, opened 2023-02-01
Force DKG termination upon receiving SAP update

### LINKED ISSUE #2028: DKG: Treat SAP change as a proxy for acks for dkg parts (removes need to cache old instances)
We should be able to force DKG to generate the key share when we've seen a SAP change.
Take a look at this function here:

https://github.com/maidsafe/sn_sdkg/blob/main/src/sdkg.rs#L367-L386

We can call `generate()` as soon as we have all the Part's but we don't because we can't trust that all the parts were from honest nodes. To protect against this, We do one more round of ACKS to make sure that all parties have seen the same parts.

But, if we see a SAP change, this means that a threshold of nodes have called `generate()`  to generate a key and then used it to sign the SAP change. That means at least one honest node has seen all ACKS over all parts.

Therefore, if we see a SAP change. We can safely call `generate()` on our DKG instance to get our key share even if we haven't seen all the ACK's.

This would remove the need to cache old DKG instances, we'd just keep our DKG instances until we see a SAP change.
Once we see a SAP change, we call `generate()`  and see if the generated public key set matches the one in the SAP.
- If it does, we are now an elder :tada: so store the `sk_share` from the `generate()` call.
- If it doesn't match, then another concurrent DKG instance won the race :disappointed: so throw away the key share.

grumbach:
- Fixes #2028


sn_node/src/node/mod.rs
```diff
@@ -604,7 +614,10 @@ mod core {
                         ))?);
                     }
                 } else {
-                    warn!("We're an elder but are missing our section key share, delaying elder state initialization until we receive it: sap={sap:?}");
+                    error!(
+                        "We're an elder but are missing our section key share for section key:{:?}",
+                        sap.section_key()
+                    );
```
grumbach, line 620: Turned this into an error as it should not be happening anymore!

grumbach: I did wonder if that was necessary but throwing the error above doesn't help since we can't handle it above either. 
Actually I can't really think of a scenario where that might happen without faulty nodes and leaking private keys or A BUG in the code.
It's certainly an error we have to keep track of. 
Restarting the node? So we can notice this error? I was thinking it might help to panic here in a testnet just to make sure it never happens. But maybe not on prod.

sn_node/src/node/mod.rs
```diff
@@ -537,29 +537,29 @@ mod core {
...
-            let new_section_key = new.network_knowledge.section_key();
-            let new_prefix = new.network_knowledge.prefix();
-            let old_prefix = old.network_knowledge.prefix();
-            let old_section_key = old.network_knowledge.section_key();
+        /// Updates DKG state on section change
+        /// Force terminates ongoing sessions to obtain section key
+        /// Discards past sessions
+        fn update_dkg_on_section_change(&mut self) {
+            let current_chain_len = self.network_knowledge.section_chain_len();
+            let prev_chain_len = current_chain_len - 1;
```
davidrusu, line 545: Are you sure the current chain len is never 0?

To be safe, we could rewrite the check to

```
                    .filter(|(_, info)| info.session_id.section_chain_len + 1 == current_chain_len)
```

sn_node/src/node/messaging/dkg.rs
```diff
@@ -705,6 +705,39 @@ impl MyNode {
...
+                    session_id.sh(),
+                );
+                // Add our new keyshare to our cache, we will then use
+                // it to sign any msg that needs section agreement.
+                let key_share = SectionKeyShare {
+                    public_key_set: new_pubs,
+                    index: our_id.into(),
+                    secret_key_share: new_sec,
+                };
+                self.section_keys_provider.insert(key_share);
```
davidrusu, line 724: We want to only keep this key share IFF the public_key_set matches the current sap. Otherwise, we should drop the key share.

You can modify this function to take the expected public_key_set as argument:
```
pub(crate) fn force_dkg_termination(&mut self, session_id: &DkgSessionId, expected_pks: PublicKeySet) {
```
And then check for it here:

```
if key_share.public_key_set == expected_pks {
    self.section_keys_provider.insert(key_share);
}
```

davidrusu (commented):
Looks good, just a few changes needed so that we don't pollute the key cache with invalid keys.

sn_node/src/node/mod.rs
grumbach: Chain len should be 1 at genesis, then 2 at the first SAP update. Worse case scenario if for some reason it becomes 0, we have an underflow and none of the DKG sessions match the `prev_chain_len`, which is fine. 
As both ways are safe I'm thinking of keeping the most readable one.

sn_node/src/node/mod.rs
joshuef: We can error out all the way up to the `create_node` layer, and so if it happens, we'll just restart with a fresh node at least?

davidrusu (approved):

PR #2064, opened 2023-02-03
Fix Misc User Experience Issues

jacderida:
- 1c83fd880 **fix: pull the correct version of sn_node**

  We added a new crate but forgot to account for the new version in the `node install` command, so it
  was using sn_client's version number.

- 41fc52222 **feat: only submit traces if environment var is set**

  We will only configure the tracing infrastructure if the `otlp` feature is enabled *and* the
  `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable is set.


joshuef (approved):
bors r+

PR #2065, opened 2023-02-03
chore(node): use parking lot and setup deadlock detection

joshuef:

### DIFF
--- a/sn_node/src/bin/sn_node/main.rs
+++ b/sn_node/src/bin/sn_node/main.rs
@@ -61,6 +61,33 @@ fn main() -> Result<()> {
         (rt, guard)
     };
 
+    {
+        use parking_lot::deadlock;
+        use std::thread;
+
+        // Create a background thread which checks for deadlocks every 3s
+        let _handle = thread::spawn(move || loop {
+            thread::sleep(Duration::from_secs(3));
+            let deadlocks = deadlock::check_deadlock();
+            if deadlocks.is_empty() {
+                continue;
+            }
+
+            println!("{} deadlocks detected", deadlocks.len());
+            warn!("{} deadlocks detected", deadlocks.len());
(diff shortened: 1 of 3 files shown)


joshuef:
bors try

joshuef:
bors try

sn_node/src/bin/sn_node/main.rs
```diff
@@ -61,6 +61,34 @@ fn main() -> Result<()> {
         (rt, guard)
     };
 
+    {
+        // only for #[cfg]
```
RolandSherwin, line 65: Is the deadlock detection thread meant to be enabled as an optional feature? If so, I think the flag is absent here...

joshuef: At the moment, I think we just enable it outright. We can disable w/ feat flag later when we're confident we're not seeing it.

oetyng (approved):
LGTM. Just the doc / sleep time diff, left a comment there.

sn_node/src/bin/sn_node/main.rs
```diff
@@ -61,6 +61,33 @@ fn main() -> Result<()> {
         (rt, guard)
     };
 
+    {
+        use parking_lot::deadlock;
+        use std::thread;
+
+        // Create a background thread which checks for deadlocks every 10s
+        let _handle = thread::spawn(move || loop {
+            thread::sleep(Duration::from_secs(3));
```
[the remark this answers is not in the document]
joshuef: ah, 10 was the copypasta. 3 was thinking, it'd be nice to know more _sooner_. (I think this is how often we report of a stall... if it cleans itself up before, then nothing is reported?)

i'll correct the msg 👍 thanks

joshuef:
bors try

joshuef:
bors try

PR #2069, opened 2023-02-06
fix(client): ignore successfully stopped stream on

b-zee:
If we call finish, but the node has already signalled `stop(0)`, then
our bytes are already received successfully, so we do not need to error
out or retry.


joshuef (approved):
bors r+

PR #2073, opened 2023-02-07
chore: remove unused async

RolandSherwin:
- These were missed by clippy since the `#[instrument]` macro interferes with it


joshuef (approved):
bors r+

Issue #2075, opened 2023-02-07
Remove SAP `members` interface


joshuef:
It's often erroneously used in place of `membership`. We should remove [this calls](https://github.com/maidsafe/safe_network/blob/main/sn_interface/src/network_knowledge/section_authority_provider.rs#L184-L186) or make it a private helper, and use the membership struct itself whenever we need to choose adults/members etc. (fault tolerance, data targets etc)

PR #2077, opened 2023-02-07
chore(sn_node): remove SAP::members function from the public API

maqi:


maqi:
closed as decided to obselete the issue https://github.com/maidsafe/safe_network/issues/2075
and to go towards the approach as described within https://github.com/maidsafe/safe_network/issues/2081

PR #2079, opened 2023-02-07
chore: remove redundant enum variant

oetyng:
- `ClientDataResponse` variant `CommunicationIssues` was not differentiated on client, and was carrying the exact same error enum as the variant `NetworkIssue`.


oetyng:
> Canceled.

Bah, didn't notice merge was going on.

PR #2084, opened 2023-02-08
refactor(node): send response to joiner

b-zee:
TryJoin wasn't previously answered over the bi-stream. This adds a
'Considering' response that is sent back over the same stream.

### DIFF
--- a/sn_interface/src/messaging/system/join.rs
+++ b/sn_interface/src/messaging/system/join.rs
@@ -37,6 +37,8 @@ pub enum JoinResponse {
     Approved(Decision<NodeState>),
     /// Join was rejected
     Rejected(JoinRejectReason),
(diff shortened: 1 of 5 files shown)


sn_interface/src/messaging/system/join.rs
```diff
@@ -37,6 +37,8 @@ pub enum JoinResponse {
     Approved(Decision<NodeState>),
     /// Join was rejected
     Rejected(JoinRejectReason),
+    /// Join is being considered
+    Considering,
```
joshuef, line 41: `UnderConsideration` perhaps chimes better with the current naming

joshuef (changes requested):

joshuef (approved):
bors r+

PR #2086, opened 2023-02-09
fix(node): prevent endlessly forwarding client msg if we're data holder

joshuef:


sn_node/src/node/messaging/mod.rs
```diff
@@ -82,7 +82,12 @@ impl MyNode {
 
         // alternatively we could flag in msg kind for this...
         // todo: this peer is actually client + forwarder ip....
-        let is_for_us = wire_msg.dst().name == context.name || msg_kind.is_client_spend();
+
+        // we've forwadedd it to ourselves as we're the holder. This prevents a loop.
```
oetyng, line 86: a typo

oetyng (approved):

PR #2088, opened 2023-02-09
tests(node): fix using the response stream

b-zee:
The other side expects us to send the UnderConsideration message,
otherwise it will get unexpected errors because the stream is not
responded to


joshuef (approved):
bors r+

b-zee:
Superseded by #2100
Thanks @RolandSherwin!

Issue #2081, opened 2023-02-08
membership/network knowledge: Merge Section Peers and Membership


davidrusu:
Currently we have 2 different view of section membership:
1. NetworkKnowledge::section_peers() - a set of section-signed nodestates 
2. Memberhip::section_members() - bootstrap members + a list of membership churn decisions

These two views can differ sometimes due to different rules for how these two membership sets are maintained.

Ideally we move to a single representation and I think that should be the the Membership view but moved to NetworkKnowledge.

That is, SectionPeers will become something like:

```rust
struct SectionPeers {
    bootstrap_members: BTreeSet<NodeState>,
    decisions: Vec<Decision<NodeState>>,
    archive: BTreeMap<XorName, Decision<NodeState>>,
}

impl SectionPeers {
    fn members(&self) -> BTreeSet<NodeState> {
        let mut members = self.bootstrap_members.clone()
        for decision in self.decisions.iter() {
            // ... apply decision to `members` set.
        }
    }
}
```
And Membership can then read the current section members state from NetworkKnowledge so we can drop the `Membership::{history, bootstrap_members}` field and just deal with driving the consensus flow.

Doing all this in one PR is probably too large a task so I think we can break it down into a few steps:

- [x] Change SectionPeers and AntiEntropy to deal with Decisions instead of SectionSigned, but keep everything else the same.
```rust
struct SectionPeers {
    members: BTreeMap<XorName, SectionSigned<NodeState>>,
    archive: BTreeMap<XorName, SectionSigned<NodeState>>,
}
// to
pub(super) struct SectionPeers {
    members: BTreeMap<XorName, Decision<NodeState>>,
    archive: BTreeMap<XorName, Decision<NodeState>>,
}

// and

AntiEntropyKind::Update { members: BTreeSet<SectionSigned<NodeState>> }
// to
AntiEntropyKind::Update { members: BTreeSet<Decision<NodeState>> }
```
- [x] Then we can make the change to SectionPeers to derive membership state from the bootstrap_members and decisions:
``` rust
struct SectionPeers {
    bootstrap_members: BTreeSet<NodeState>,
    decisions: Vec<Decision<NodeState>>,
    archive: BTreeMap<XorName, Decision<NodeState>>,
}
```
- [ ] Then finally delete the `Membership::{bootstrap_members, history}` fields and instead read that state from NetworkKnowledge.

I probably missed some steps along the way but I think that's the broad strokes.

maqi:
[...]

PR #2089, opened 2023-02-09
feat(sn_node): refactor SectionPeers to use Decision

maqi:
Being the first step as described within the issue https://github.com/maidsafe/safe_network/issues/2081 , the work of this PR relies on couple of TODO points that shall get addressed in the following PRs :
1, the util function of `section_signed_to_decision` shall get removed once complete the full transition
2, Decision requires the PublicKeySet to validate, however NetworkKnowledge only retains the PublicKey info currently. Need to re-enable the validation in the following PR
3, The current `update`, `merge_members` procedure is based on `node_state` as atomic. However `Decision` may contain multipe of `node_state` within one. Need to double check this won't incur any trouble.
4, The current `members` function return list of `Decision`, which may contain `Join` and `Left` in one Decision, or multiple Joins. This need to be double checked in the following PRs.
5,

### DIFF
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -44,7 +44,12 @@ use crate::{
 use bls::PublicKey as BlsPublicKey;
 use section_peers::SectionPeers;
 use serde::Serialize;
-use std::{collections::BTreeSet, iter, net::SocketAddr};
+use sn_consensus::Decision;
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    iter,
+    net::SocketAddr,
+};
 use xor_name::{Prefix, XorName};
 
 /// The secret key for the genesis DBC.
@@ -154,6 +159,17 @@ pub struct NetworkKnowledge {
     section_tree: SectionTree,
 }
 
+// Convert SectionSigned to Decision
+fn section_signed_to_decision(section_signed: SectionSigned<NodeState>) -> Decision<NodeState> {
+    let mut proposals = BTreeMap::new();
+    let _ = proposals.insert(section_signed.value, section_signed.sig.signature);
+    Decision::<NodeState> {
+        votes: BTreeSet::new(),
+        proposals,
+        faults: BTreeSet::new(),
+    }
+}
+
 impl NetworkKnowledge {
     /// Creates a new `NetworkKnowledge` instance. The prefix is used to choose the
     /// `signed_sap` from the provided SectionTree
@@ -188,10 +204,12 @@ impl NetworkKnowledge {
         for peer in network_knowledge.signed_sap.elders() {
             let node_state = NodeState::joined(*peer, None);
             let sig = create_first_sig(&public_key_set, &secret_key_share, &node_state)?;
-            let _changed = network_knowledge.section_peers.update(SectionSigned {
-                value: node_state,
-                sig,
-            });
+            let _changed = network_knowledge
+                .section_peers
+                .update(section_signed_to_decision(SectionSigned {
+                    value: node_state,
+                    sig,
+                }));
         }
 
         let section_key_share = SectionKeyShare {
@@ -411,29 +429,19 @@ impl NetworkKnowledge {
 
     /// Try to merge this `NetworkKnowledge` members with `peers`.
     /// Checks if we're already up to date before attempting to verify and merge members
-    pub fn merge_members(&mut self, peers: BTreeSet<SectionSigned<NodeState>>) -> Result<bool> {
+    pub fn merge_members(&mut self, peers: BTreeSet<Decision<NodeState>>) -> Result<bool> {
         let mut there_was_an_update = false;
-        let our_current_members = self.section_peers.members();
 
-        for node_state in &peers {
-            if our_current_members.contains(node_state) {
-                // we already know of this one, so nothing to do here.
-                continue;
-            }
-            trace!(
-                "Attempting to update section members. Name: {:?}, new state: {:?}",
-                node_state.name(),
-                node_state.state()
-            );
-            if !node_state.verify(&self.section_chain()) {
-                error!(
-                    "Can't update section member, name: {:?}, new state: {:?}",
-                    node_state.name(),
-                    node_state.state()
-                );
-            } else if self.section_peers.update(node_state.clone()) {
-                there_was_an_update = true;
-            }
+        for decision in &peers {
+            // TODO: Decisioin requires the PublicKeySet to validate,
+            //       however NetworkKnowledge only retains the PublicKey info
+            // if !decision.validate(&self.section_chain()).is_ok() {
+            //     error!("Can't update decision {decision:?}");
+            //     continue;
+            // }
+
+            trace!("Updating decision {decision:?}",);
+            there_was_an_update |= self.section_peers.update(decision.clone());
         }
 
         self.section_peers.retain(&self.prefix());
@@ -458,7 +466,9 @@ impl NetworkKnowledge {
             return false;
         }
 
-        let updated = self.section_peers.update(node_state);
+        let updated = self
+            .section_peers
+            .update(section_signed_to_decision(node_state));
         trace!("Section member state, name: {node_name:?}, updated: {updated}");
 
         updated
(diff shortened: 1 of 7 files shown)


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -408,27 +426,33 @@ impl NetworkKnowledge {
...
+            //       will cause any trouble.
+            for node_state in decision.proposals.keys() {
+                if our_current_members.contains(node_state) {
+                    // we already know of this one, so nothing to do here.
+                    trace!(
+                        "Failed to update decision {decision:?}, as already know {:?} new_state {:?}",
+                        node_state.name(),
+                        node_state.state()
+                    );
+                    continue;
```
davidrusu, line 451: This `continue` seems to do nothing, can we drop it?

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -71,32 +93,40 @@ impl SectionPeers {
...
-
-        match (self.members.entry(node_name), new_state.state()) {
-            (Entry::Vacant(entry), MembershipState::Joined) => {
-                // unless it was already archived, insert it as current member
-                if self.archive.contains_key(&node_name) {
-                    false
-                } else {
-                    entry.insert(new_state);
+    pub(super) fn update(&mut self, new_decision: Decision<NodeState>) -> bool {
+        // TODO: currently only support Decision contains one NodeState.
```
davidrusu, line 97: This TODO seems wrong/outdated, you have a loop here that deals with every proposal in the decision. So it seems we are supporting decisions with multiple node states already.

maqi: although there is loop along all the proposals, the return value of `update` only support the `last entry` or the `first failed to update entry`.
So, the `TODO` point still valid ?

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -21,47 +19,71 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
...
 }
 
 impl SectionPeers {
     /// Returns set of current members, i.e. those with state == `Joined`.
-    pub(super) fn members(&self) -> BTreeSet<SectionSigned<NodeState>> {
+    pub(super) fn members(&self) -> BTreeSet<NodeState> {
+        // TODO: ensure Decision has no Joined and Left combined within one
+        let mut node_state_list = BTreeSet::new();
+        for decision in self.members.values() {
+            for node_state in decision.proposals.keys() {
```
davidrusu, line 34: I think these TODO's can be resolved in this PR with a filter.
```suggestion
            for node_state in decision.proposals.keys().filter(|node| matches!(node.state, MembershipState::Joined) {
```

sn_interface/src/network_knowledge/mod.rs
maqi: Ah, this shall be a `continue to the outer layer loop`, not `the current layer loop`, I will try to figure out how to do it properly

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -151,6 +156,17 @@ pub struct NetworkKnowledge {
...
+// Convert SectionSigned to Decision
+fn section_signed_to_decision(section_signed: SectionSigned<NodeState>) -> Decision<NodeState> {
+    let mut proposals = BTreeMap::new();
+    let _ = proposals.insert(section_signed.value, section_signed.sig.signature);
+    Decision::<NodeState> {
+        votes: BTreeSet::new(),
+        proposals,
+        faults: BTreeSet::new(),
+    }
+}
```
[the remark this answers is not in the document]
maqi: ok, will do

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -21,47 +19,71 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
...
     }
 
     /// Returns set of archived members, i.e those that've left our section
-    pub(super) fn archived_members(&self) -> BTreeSet<SectionSigned<NodeState>> {
-        self.archive.values().cloned().collect()
+    pub(super) fn archived_members(&self) -> BTreeSet<NodeState> {
+        // TODO: ensure Decision has no Joined and Left combined within one
+        let mut node_state_list = BTreeSet::new();
+        for decision in self.archive.values() {
+            for node_state in decision.proposals.keys() {
```
davidrusu, line 53: Again, just add a filter here and we can drop the TODO's
```suggestion
            for node_state in decision.proposals.keys().filter(|node| node.state != MembershipState::Joined) {
```

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -408,27 +426,33 @@ impl NetworkKnowledge {
...
+            //     continue;
+            // }
+
+            // TODO: check whether one decision may contain multiple node_state
+            //       will cause any trouble.
+            for node_state in decision.proposals.keys() {
+                if our_current_members.contains(node_state) {
+                    // we already know of this one, so nothing to do here.
+                    trace!(
+                        "Failed to update decision {decision:?}, as already know {:?} new_state {:?}",
```
davidrusu, line 447: This log may confuse people.. the `section_peers.update(decision)` will loop over the proposals and handle each one. So we may get a partial update if some proposals are new.

I'd drop the entire for loop if you don't need it.

sn_interface/src/network_knowledge/section_peers.rs
maqi: Well, for `SectionPeers`, the `self.members` shall supposed to already only contain `Joined` when insertion, so here shall no need to carry out a filter.

sn_interface/src/network_knowledge/section_peers.rs
davidrusu: I noticed that as I kept reading but I think that's going to cause problems, I made some suggestions further down to pull `updated` out of the loop

sn_interface/src/network_knowledge/section_peers.rs
maqi: Same comment as above https://github.com/maidsafe/safe_network/pull/2089#discussion_r1107092806

sn_interface/src/network_knowledge/mod.rs
maqi: as mentioned at https://github.com/maidsafe/safe_network/pull/2089#discussion_r1107063842 ,
the looping layer is incorrectly `continued`, I will update this block of code to make the log valid.

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -71,32 +93,40 @@ impl SectionPeers {
...
+                    //  remove it from our current members, and insert it into our archive
+                    let _ = entry.remove();
+                    let _ = self.archive.insert(node_name, new_decision.clone());
+                    true
+                }
+            };
+            if !updated {
+                error!("Failed to update {new_decision:?} as conflict with existring records");
+                return updated;
             }
```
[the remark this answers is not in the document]
maqi: In theory, any update attempt shall get passed and undertaken.
A rejected update attempt indicates a potential incorrect action, hence the `error` log got chosen? that's what I understand why use `error` log here.
Though `re-sent an old decision that we've already processed` could be one reason, but there could be other cases?

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -71,32 +93,40 @@ impl SectionPeers {
...
-                if self.archive.contains_key(&node_name) {
-                    false
-                } else {
-                    entry.insert(new_state);
+    pub(super) fn update(&mut self, new_decision: Decision<NodeState>) -> bool {
+        // TODO: currently only support Decision contains one NodeState.
+        for new_state in new_decision.proposals.keys() {
+            let node_name = new_state.name();
+
+            let updated = match (self.members.entry(node_name), new_state.state()) {
```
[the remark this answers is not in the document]
maqi: The reason this `updated` is kept inside the `for clause`, is that it's based on the assumption that `Decision contains just one proposal(NodeState)`.
I think we expect the `update` is atomic, i.e. update of a Decision is succeeded on whole, and there shall be no `partial update` ? 
That's why I put a TODO point in this function, instead of move the `updated outside of loop` to support a potential `partial update`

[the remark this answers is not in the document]
davidrusu: We do regularly see decisions with multiple proposals. It's not much work to adapt the logic to handle the multiple proposals properly so can we do that to avoid issues.

joshuef:
bors try

sn_interface/src/network_knowledge/mod.rs
davidrusu: Can we drop this loop now that we are handling multiple proposals in a decision?

maqi:
bors try

sn_interface/src/network_knowledge/mod.rs
maqi: This is to avoid unnecessary updates of existing entries during `merge` , so I think may still required?

davidrusu (approved):
Looks good 🚢

davidrusu:
r+

PR #2090, opened 2023-02-09
fix(node): send response to the client upon missing response from a peer

bochaco:
When an Elder is processing a client request/msg, and it timed out due to not receiving a response from a data holder, a response needs to be sent to the client rather than simply dropping the stream.


bochaco:
bors try

PR #2094, opened 2023-02-13
fix: verifiable relocation trigger

### LINKED ISSUE #2011: Make relocation details verifiable.

b00f:
## Summary

In this PR:
- `CurhnId` moved to network_knowledge
- `RelocationDst` replaced with `RelocationTrigger` 
- `RelocationTrigger` has decision and therefore it is verifiable
- Test are updated to use trigger (result of decision)

## Issues

It fixes #2011 

## TODO

Perhaps we can mock the decision (consensus) for faster tests.

### DIFF
--- a/sn_interface/Cargo.toml
+++ b/sn_interface/Cargo.toml
@@ -11,14 +11,14 @@ repository = "https://github.com/maidsafe/safe_network"
 version = "0.19.7"
 
 [features]
-test-utils=["proptest"]
+test-utils = ["proptest"]
 
 [dependencies]
 base64 = "~0.13.0"
 bincode = "1.3.1"
 bls = { package = "blsttc", version = "8.0.1" }
 bytes = { version = "1.0.1", features = ["serde"] }
-crdts = { version = "7.2", default-features=false, features = ["merkle"] }
+crdts = { version = "7.2", default-features = false, features = ["merkle"] }
 custom_debug = "~0.5.0"
 ed25519 = { version = "1.2.0", features = ["serde_bytes"] }
 ed25519-dalek = { version = "1.0.0", features = ["serde"] }
@@ -30,10 +30,10 @@ itertools = "~0.10.0"
 lazy_static = "1"
 multibase = "~0.9.1"
 num_cpus = "1.13.0"
-proptest = { version ="1.0.0", optional =true }
+proptest = { version = "1.0.0", optional = true }
 qp2p = "~0.36.2"
 rand = "~0.8.5"
-rand-07 = { package = "rand", version = "0.7.3" } # required till ed25519-dalek upgrades to rand v0.8
+rand-07 = { package = "rand", version = "0.7.3" }                                    # required till ed25519-dalek upgrades to rand v0.8
 rayon = "1.5.1"
 rmp-serde = "1.0.0"
 self_encryption = "~0.28.0"
@@ -52,7 +52,7 @@ tiny-keccak = { version = "2.0.2", features = ["sha3"] }
 tracing = "~0.1.26"
 tracing-core = "~0.1.21"
 tracing-subscriber = { version = "0.3.1", features = ["env-filter", "json"] }
-uluru="3.0.0"
+uluru = "3.0.0"
 xor_name = "~5.0.0"
 
 [dependencies.tokio]
@@ -66,7 +66,7 @@ proptest = { version = "1.0.0" }
 
 [dev-dependencies.cargo-husky]
 version = "1.5.0"
-default-features = false # Disable features which are enabled by default
+default-features = false                                           # Disable features which are enabled by default
 features = ["precommit-hook", "run-cargo-clippy", "run-cargo-fmt"]
 
 [package.metadata.cargo-udeps.ignore]
--- a/sn_interface/src/network_knowledge/mod.rs
+++ b/sn_interface/src/network_knowledge/mod.rs
@@ -20,12 +20,11 @@ pub mod test_utils;
 #[cfg(any(test, feature = "test-utils"))]
 pub use section_tree::test_utils as test_utils_st;
 
+use self::node_state::ChurnId;
 pub use self::{
     errors::{Error, Result},
     node_info::MyNodeInfo,
-    node_state::{
-        MembershipState, NodeState, RelocationDst, RelocationInfo, RelocationProof, RelocationState,
-    },
+    node_state::{MembershipState, NodeState, RelocationInfo, RelocationProof, RelocationState},
     section_authority_provider::{SapCandidate, SectionAuthUtils, SectionAuthorityProvider},
     section_keys::{SectionKeyShare, SectionKeysProvider},
     section_tree::{SectionTree, SectionTreeUpdate},
@@ -585,10 +584,35 @@ fn create_first_sig<T: Serialize>(
     })
 }
 
+// Returns the number of trailing zero bits of the bytes slice.
+pub fn trailing_zeros(bytes: &[u8]) -> u32 {
+    let mut output = 0;
+
+    for &byte in bytes.iter().rev() {
+        if byte == 0 {
+            output += 8;
+        } else {
+            output += byte.trailing_zeros();
+            break;
+        }
+    }
+
+    output
+}
+
+// Relocation check - returns whether a member with the given age is a candidate for relocation on
+// a churn event with the given churn id.
+pub fn relocation_check(age: u8, churn_id: &ChurnId) -> bool {
+    // Evaluate the formula: `churn_id % 2^age == 0` Which is the same as checking the churn_id
+    // has at least `age` trailing zero bits.
+    trailing_zeros(&churn_id.0) >= age as u32
+}
+
 #[cfg(test)]
 mod tests {
     use super::{supermajority, NetworkKnowledge};
     use crate::{
+        network_knowledge::trailing_zeros,
         test_utils::{gen_addr, prefix, TestKeys, TestSapBuilder, TestSectionTree},
         types::NodeId,
     };
@@ -598,6 +622,18 @@ mod tests {
     use proptest::prelude::*;
     use rand::thread_rng;
     use xor_name::XorName;
+
+    #[test]
+    fn byte_slice_trailing_zeros() {
+        assert_eq!(trailing_zeros(&[0]), 8);
+        assert_eq!(trailing_zeros(&[1]), 0);
+        assert_eq!(trailing_zeros(&[2]), 1);
+        assert_eq!(trailing_zeros(&[4]), 2);
+        assert_eq!(trailing_zeros(&[8]), 3);
+        assert_eq!(trailing_zeros(&[0, 0]), 16);
+        assert_eq!(trailing_zeros(&[1, 0]), 8);
+        assert_eq!(trailing_zeros(&[2, 0]), 9);
+    }
 
     #[test]
     fn supermajority_of_small_group() {
--- a/sn_interface/src/network_knowledge/node_state/relocation.rs
+++ b/sn_interface/src/network_knowledge/node_state/relocation.rs
@@ -6,8 +6,6 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::fmt::{self, Formatter};
-
 use super::NodeState;
 use crate::messaging::system::SectionSigned;
 use crate::network_knowledge::{Error, Result};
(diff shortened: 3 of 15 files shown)


sn_node/src/node/messaging/relocation.rs
```diff
@@ -316,8 +318,16 @@ mod tests {
                 }
             })
         {
-            initialize_relocation(dispatcher, relocation_node_old_name, prefix1, msg_counter)
-                .await?;
+            let sk_set = env.get_secret_key_set(prefix1, None);
+            let age = relocation_node_old_name[XOR_NAME_LEN - 1];
+            let relocation_trigger = create_relocation_trigger(&sk_set, age + 1);
```
b00f, line 323: @joshuef here we create relocation trigger. It is repeated for all nodes (17 nodes) since it is brute forcing to create trigger, the test is slow.

joshuef: Ah right. This slowness is coming from the PR, right? Is there another way to do this rather than brute forcing? Passing the relocation trigger in after one run perhaps?

b00f: Not sure how we can fix it. Mocking decision looks the good option here, but needs refactoring the code.

joshuef: why does it need to happen for all 17 nodes? Can't we do it once in the test and then pass it in?

b00f: @joshuef I moved it out of the loop, now is faster.

joshuef:
bors try

sn_interface/src/network_knowledge/node_state/relocation.rs
```diff
@@ -6,7 +6,7 @@
 // KIND, either express or implied. Please review the Licences for the specific language governing
 // permissions and limitations relating to use of the SAFE Network Software.
 
-use std::fmt::{self, Formatter};
+use std::fmt::{self, Display, Formatter};
```
oetyng, line 9: Nitpick:

Third party libs would - by our loose and (afaik) not entirely verbalized conventions - form a section of lines furthest down.
On top of that is our `sn_*` libs, then `crate`, `super` and lastly `self`.

So that basically means the usings are structured according to how close to this file they are.

```
use self::Abc;
use self::Bce;

use super::Cde;
use super::Def;

use crate::Efg;
use crate::Fgh;

use sn_ghi::Hij;
use sn_ijk::Jkl;

use std::fmt::Display;
use xor_name::XorName;
```

(Note that, as can be seen by the above use of `xor_name`, all non `sn_*` libs are treated as third party, even though they are created/maintained by maidsafe.)

oetyng: Usually when there is just one `super` for example, I'd not add an empty line between that and a `crate` use.

Usually we seem to not combine the `crate` or `super` usings, and I think having them separate improves readability.

All other though are combined. I.e.:
```
use serde::{Deserialize, Serialize};
```

But remember these are quite loose conventions, and I would guess that not everyone think it matters much, down to the details described above. Those are a mix of my preferences and what I think is followed by most.

Usings in tests are even more loosely approached generally.

sn_interface/src/network_knowledge/node_state/relocation.rs
```diff
@@ -15,20 +15,56 @@ use crate::network_knowledge::{Error, Result};
...
     /// We are relocating to the section that matches `dst_name`.
-    pub fn new(dst_name: XorName) -> Self {
-        Self(dst_name)
+    pub fn new(decision: Decision<NodeState>) -> Self {
+        Self(decision)
     }
 
-    pub fn name(&self) -> &XorName {
-        &self.0
+    ///
```
oetyng, line 45: Maybe add a few words here.

sn_interface/src/network_knowledge/node_state/relocation.rs
```diff
@@ -15,20 +15,56 @@ use crate::network_knowledge::{Error, Result};
...
+        for sig in self.0.proposals.values() {
+            content_parts.push(sig.to_bytes().to_vec());
+        }
+
+        XorName::from_content_parts(
+            Vec::from_iter(content_parts.iter().map(|v| v.as_slice())).as_slice(),
+        )
+    }
+
+    ///
```
oetyng, line 58: Maybe add a few words here.

sn_node/src/node/relocation.rs
```diff
@@ -236,6 +181,7 @@ mod tests {
     // from an unspecified random data using an unspecified random `SecretKey`. That is OK because
     // the relocation algorithm doesn't care about whether the signature is valid. It only
     // cares about its number of trailing zeros.
+    #[allow(dead_code)]
     fn signature_with_trailing_zeros(trailing_zeros_count: u32) -> bls::Signature {
```
oetyng, line 185: Do we need to keep this fn?

sn_node/src/node/relocation.rs
```diff
@@ -253,6 +199,7 @@ mod tests {
         })
     }
 
+    #[allow(dead_code)]
     fn gen_signature_with_trailing_zeros(trailing_zeros_count: u32) -> bls::Signature {
```
oetyng, line 203: Do we need to keep this fn?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -598,10 +599,35 @@ fn create_first_sig<T: Serialize>(
     })
 }
 
+// Returns the number of trailing zero bits of the bytes slice.
+pub fn trailing_zeros(bytes: &[u8]) -> u32 {
```
oetyng, line 603: Is there any special reason this is moved to `network_knowledge`?

oetyng (approved):
LGMT just a few questions and minor points.

sn_node/src/node/flow_ctrl/tests/cmd_utils.rs
```diff
@@ -162,7 +162,12 @@ impl<'a> ProcessAndInspectCmds<'a> {
     }
 
     pub(crate) async fn next(&mut self) -> crate::node::error::Result<Option<&Cmd>> {
-        let mut next_index = self.index_inspected + 1;
+        let mut next_index = if self.index_inspected == usize::MAX {
```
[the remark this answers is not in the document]
b00f: Nope, unless we define `isize` and set it to -1

sn_interface/src/network_knowledge/mod.rs
b00f: Otherwise we need to duplicate the code. It is used by `relocation_check` in `network_knowledge`. Any suggestion?

sn_node/src/node/relocation.rs
b00f: Yes, unless we delete them, they are not used by any other modules.

sn_interface/src/network_knowledge/node_state/relocation.rs
b00f: Done

sn_interface/src/network_knowledge/test_utils.rs
```diff
@@ -274,3 +274,29 @@ where
...
+) -> (RelocationTrigger, Decision<NodeState>) {
+    use super::relocation_check;
+
+    loop {
+        let node_state = NodeState::joined(gen_peer(MIN_ADULT_AGE), Some(xor_name::rand::random()));
+        let decision = section_decision(sk_set, node_state.clone());
+        let relocation_trigger = RelocationTrigger::new(decision.clone());
+        let churn_id = relocation_trigger.churn_id();
+
+        if relocation_check(age, &churn_id) && !relocation_check(age + 1, &churn_id) {
```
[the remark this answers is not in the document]
b00f: I am not sure why. This code is only uses by tests, probably better to keep it as it is.

sn_interface/src/network_knowledge/node_state/relocation.rs
b00f: Done, Let me know if it needs to be amended

sn_interface/src/network_knowledge/node_state/relocation.rs
b00f: Done, Let me know if it needs to be amended

sn_interface/Cargo.toml
```diff
@@ -30,18 +30,18 @@ itertools = "~0.10.0"
...
+rand-07 = { package = "rand", version = "0.7.3" }                                    # required till ed25519-dalek upgrades to rand v0.8
 rayon = "1.5.1"
 rmp-serde = "1.0.0"
 self_encryption = "~0.27.5"
 serde = { version = "1.0.111", features = ["derive", "rc"] }
 serde_bytes = "~0.11.5"
 serde_json = "1.0.53"
 signature = "1.1.10"
-sn_consensus = "3.2.1"
+sn_consensus = { version = "3.2", git = "https://github.com/maidsafe/sn_consensus" }
```
davidrusu, line 44: let's drop the git reference now that sn_consensus is published.
```suggestion
sn_consensus = "3.2.1"
```

sn_interface/src/network_knowledge/test_utils.rs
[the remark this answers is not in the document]
davidrusu: I think the idea is that this helper generates a relocation trigger that will only relocate nodes of a specific age i.e. age=4

Useful in tests to target a specific node.

sn_node/Cargo.toml
```diff
@@ -55,7 +60,7 @@ rand-07 = { package = "rand", version = "~0.7.3" }
 rayon = "1.5.1"
 rmp-serde = "1.0.0"
 self_encryption = "~0.27.5"
-sn_consensus = "3.2.1"
+sn_consensus = { version = "3.2", git = "https://github.com/maidsafe/sn_consensus" }
```
davidrusu, line 63: ditto, drop the git ref
```suggestion
sn_consensus = "3.2.1"
```

sn_node/src/node/flow_ctrl/tests/cmd_utils.rs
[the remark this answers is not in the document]
davidrusu: Try `let mut next_index = self.index_inspected.wrapping_add(1);`

See docs: https://doc.rust-lang.org/std/primitive.u64.html#method.wrapping_add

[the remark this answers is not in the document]
davidrusu: Sorry, I read the code further and I think this introduces a bug no?

If `index_inspected` is usize::MAX, it'll be set to `0` here. Then on line [174](https://github.com/maidsafe/safe_network/pull/2094/files#diff-150691660f12f76fba4707f7d381fae6aeee5aabcc1e2aea83832f1e9edc8c72R179) below, it subtracts 1 so we underflow. I don't think that's what's intended.

sn_node/src/node/relocation.rs
b00f: @joshuef I tried `cfg(test)` but I got this error on clippy: 
```
function `signature_with_trailing_zeros` is never used
```

sn_node/src/node/flow_ctrl/tests/cmd_utils.rs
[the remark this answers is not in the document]
b00f: @davidrusu done

sn_node/src/node/relocation.rs
davidrusu: Yeah, if it's no longer used, remove it

davidrusu (approved):
Looks like you've got some conflicts again, and you can remove those unused functions with the `allow(dead)`.

Besides that, looks great 👍

davidrusu (approved):

joshuef (approved):
approved after rebase 👍

joshuef (approved):

PR #2103, opened 2023-02-15
Reduce qp2p async

joshuef:

### DIFF
--- a/sn_client/src/sessions/listeners.rs
+++ b/sn_client/src/sessions/listeners.rs
@@ -35,18 +35,20 @@ struct MsgResent {
 impl Session {
     #[instrument(skip_all, level = "debug")]
     async fn read_resp_from_recvstream(
-        recv_stream: &mut RecvStream,
+        recv_stream: RecvStream,
         peer: Peer,
         correlation_id: MsgId,
     ) -> Result<(MsgId, ClientDataResponse), Error> {
+        let stream_id = recv_stream.id();
         let bytes = recv_stream.read().await?;
+
         match WireMsg::deserialize(bytes)? {
             MsgType::ClientDataResponse { msg_id, msg } => Ok((msg_id, msg)),
             msg => {
                 warn!(
                     "Unexpected msg type received on {} from {peer:?} in response \
                     to {correlation_id:?}: {msg:?}",
-                    recv_stream.id()
+                    stream_id
                 );
                 Err(Error::UnexpectedMsgType {
                     correlation_id,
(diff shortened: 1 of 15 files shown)


joshuef:
bors try

joshuef:
bors try

joshuef:
do not merge until qp2p is updated

sn_client/src/sessions/listeners.rs
```diff
@@ -90,8 +94,7 @@ impl Session {
             let stream_id = recv_stream.id();
```
bochaco, line 94: we can remove this as it's now done in L80

sn_comms/src/listener.rs
```diff
@@ -48,50 +50,76 @@ pub(crate) async fn listen_for_msgs(
...
-        match result {
-            Ok((msg_bytes, send_stream)) => {
-                let stream_info = if let Some(stream) = &send_stream {
-                    format!(" on {}", stream.id())
-                } else {
-                    "".to_string()
-                };
-                debug!(
+    loop {
+        let msg = incoming_msgs.next_with_stream().await.transpose();
```
bochaco, line 54: Isn't this going into an endless loop when the there is no more msgs returned by `next_with_stream`, i.e. when it returns `None` ?
...plus, you are reserving a slot in the `comm_events` in such case which won't be used ?
Why not simply keep the `while let Some(...) = incoming_msgs.next_with_stream().await` ?

joshuef: hmm, there is a big difference in that the current setup seems to feedback to qp2p, whereas the `while` loop does not. (we're back in the land of 1000s of 100ms reads again instead of a handful).

I'm honestly not sure i can see _why_ that's the case just now though as what you say makes sense to me 🤔

joshuef: Okay been poking at this. Looks like the "improvement" here, and the feedback I'd thought we'd gained was more just about this delay here. Even with the concurrency super limited, we're still not feeding back "enough" (if there is an enough....) 

Digging more

joshuef: Right, i see what I was going for, which was attempting to wait before we poll the stream... which is the feedbaack we need. But because this kicks off only when we get a msg, we cannot reserve capacity over that... 

What actually may be needed is qp2p (for this concept at least) to advise "msg in", and only then do we read w/ capacity...

Orrr. we're throttling all processing on _some basis_... which feels perhaps more complex? Though perhaps what's needed for a stress level 🤔 

Going to think on this more

joshuef:
going to close this as it does more than we want atm, we could just set things up for feedback to comms, without going into permits for now

PR #2104, opened 2023-02-15
refactor(node): pass public addr in --first

b-zee:
The genesis node will output a network contact file which contains its own address. This public address was specified by a separate flag but was removed, making the genesis incapable of producing a proper contacts file. This changes the `--first` flag to be used to pass in that public address, which should be done for all genesis nodes, so the proper external address can be advertised.

See https://github.com/maidsafe/sn_launch_tool/pull/136 for relevant changes in the sn_launch_tool after which this PR can be turned into review mode.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4186,9 +4186,9 @@ dependencies = [
 
 [[package]]
 name = "sn_launch_tool"
-version = "0.13.0"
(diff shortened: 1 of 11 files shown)


joshuef:
bors try

joshuef (approved):
bors r+

Issue #2101, opened 2023-02-15
Moving age calculation into the interface (common) crate


b00f:
We can move the logic behind calculating age from `XorName` to somewhere that all crates have access to it (like `sn_interface`)

```rust
pub fn calc_age(name: &XorName) -> u8 {
     name()[XOR_NAME_LEN - 1]
 } 
```

joshuef:
is this code replicated over the whole code base? 

In general if you see a quick win for something like this you can fire up a PR with the change directly and we can look at it there / see any issues more clearly. 🙇

RolandSherwin:
Fixed - https://github.com/maidsafe/safe_network/pull/2106

PR #2106, opened 2023-02-16
fix: adding a helper func to calculate the age

b00f:
Adding a new helper (inline function) to calculate the age from peer name.

Fixing #2101

### DIFF
--- a/sn_interface/src/types/mod.rs
+++ b/sn_interface/src/types/mod.rs
@@ -42,7 +42,7 @@ pub use keys::{
 pub use peer::Peer;
 
 use serde::{Deserialize, Serialize};
-use xor_name::XorName;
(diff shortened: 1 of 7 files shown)


sn_interface/src/types/mod.rs
```diff
@@ -126,3 +126,8 @@ impl ReplicatedData {
         }
     }
 }
+
+#[inline]
+pub fn calc_age(xor_name: &XorName) -> u8 {
+    xor_name[XOR_NAME_LEN - 1]
+}
```
RolandSherwin, line 133: LGTM! But do you think `sn_interferace/types/utils/mod.rs` would be a better location for the function?

b00f: @RolandSherwin done.

RolandSherwin (approved):

PR #2108, opened 2023-02-16
refactor: call nodes after check still in section

oetyng:
- It is an unnecessary work started if node was removed from section.


joshuef (approved):
bors r+

PR #2111, opened 2023-02-16
chore: removing unnecessary results

oetyng:
This is probably a remainder from old code where the paths had potential errors.


joshuef (approved):
bors r+

PR #2112, opened 2023-02-16
fix: extend cleanup to registers

oetyng:
As nodes fill up they check if there's any data they are no longer responsible for, and clean it up.

It was missing the removal of registers.

NB: This can be made more granular, by looking at close nodes instead of prefix.


joshuef (approved):
bors r+

PR #2113, opened 2023-02-16
feat(client): sn_client to not have any default cmd/query timeout value set

bochaco:
sn_client is now set to send msgs to the network without any default timeout values, the user of the lib can still set them with SN_CMD_TIMEOUT and SN_QUERY_TIMEOUT env vars according to its own requirements.


bochaco:
bors try

bochaco:
bors try

bochaco:
bors try

joshuef (approved):
bors r+

PR #2114, opened 2023-02-16
chore: feat gate data

oetyng:
This gives a first simple version of a `payment-network`.
Storing data needs enabling of the `data-network` feature.

NB: This does not gate it in `sn_node`, thus with a custom client, data operations could still be available.
As the current wallet impl is dependent on the data storage, we cannot gate data without updating the wallet to work without it. That is a larger change suitable for a follow up PR.


oetyng:
bors try

oetyng:
bors try

joshuef (approved):

PR #2124, opened 2023-02-22
chore(sn_node): refactor SectionPeers to using Decision

maqi:


sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -101,39 +125,44 @@ impl SectionPeers {
...
                 }
             };
         }
+
+        if shall_push_to_decisions {
+            self.decisions.push(new_decision);
+        }
+
         updated
     }
```
davidrusu, line 161: Decisions need to be processed in order by generation, so you can reduce this function to something like:

```rust
pub(super) fn update(&mut self, decision: Decision<NodeState>) -> Result<bool> {
    if new_decision.generation()? == self.decisions.len() + 1 {
        self.decisions.push(new_decision);

        for archived_node in decision.proposals.iter().filter(|n| matches!(n.state(), MembershipState::Left)) {
           self.archive.insert(archived_node.name(), decision.clone());
        }

        Ok(true)
    } else {
        // TODO: request AE update with previous decisions.
        Ok(false)
    }
}
```

To make this work, I think we may need to also change the `AntiEntropyKind::Update(..)` method to be `AntiEntropyKind::Update(Vec<Decision>)`

maqi: > To make this work, I think we may need to also change the AntiEntropyKind::Update(..) method to be AntiEntropyKind::Update(Vec<Decision>)

The current Update sends `SectionPeers` which contains `Vec<Decision>` already. 
We can have a later on PR to optimisze it, but currently it shall be OK as long as the included content got properly applied.

davidrusu: Confusingly, when I traced the `SectionPeers` type, it was actually a type alias of `BTreeSet<SectionSig<NodeState>`.

maqi: It's an alias of `BTreeSet<Decision<NodeState>>;` actually ?
I will double check maybe `Vec<Decision<NodeState>>` will be better, might rename it as well.
OR, I might just replace it with the proper `SectionPeers` to be used.

davidrusu: I think just inlining `Vec<Decision<NodeState>>` into the Update message is good enough

sn_interface/src/network_knowledge/errors.rs
```diff
@@ -68,4 +68,6 @@ pub enum Error {
     Consensus(#[from] sn_consensus::Error),
     #[error("An archived node attempted to rejoin the section")]
     ArchivedNodeRejoined,
+    #[error("Request AE update")]
+    RequestAEUpdate,
```
oetyng, line 72: What is the actual error here?
Requesting an AE update isn't an error?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -200,17 +189,14 @@ impl NetworkKnowledge {
...
-            let node_state = NodeState::joined(*peer, None);
-            let sig = create_first_sig(&public_key_set, &secret_key_share, &node_state)?;
-            let _changed = network_knowledge
-                .section_peers
-                .update(section_signed_to_decision(SectionSigned {
-                    value: node_state,
-                    sig,
-                }));
-        }
+        let bootstrap_members = network_knowledge
```
oetyng, line 192: We only expect a single Elder here, right? the first node.

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -200,17 +189,14 @@ impl NetworkKnowledge {
...
+            .signed_sap
+            .elders()
+            .map(|peer| NodeState::joined(*peer, None))
+            .collect();
+        network_knowledge
```
oetyng, line 197: What's happening here?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -427,51 +422,47 @@ impl NetworkKnowledge {
...
-            // }
-
-            trace!("Updating decision {decision:?}",);
-            there_was_an_update |= self.section_peers.update(decision.clone());
-        }
-
-        self.section_peers.retain(&self.prefix());
-
-        Ok(there_was_an_update)
+    /// This function shall only get called at the time of SAP change
```
oetyng, line 425: Also good for docs: _Why_ is it only called at the time of SAP change, and what does it mean in the domain to reset the bootstrap members?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -427,51 +422,47 @@ impl NetworkKnowledge {
...
+        //       however NetworkKnowledge only retains the PublicKey info
+        // if !decision.validate(&self.section_chain()).is_ok() {
+        //     error!("Can't update decision {decision:?}");
+        //     continue;
+        // }
+        Ok(self.section_peers.update_peers(peers))
+    }
+
+    /// Update one decision. Returns whether it actually updated.
+    pub fn update_decision(&mut self, decision: Decision<NodeState>) -> Result<bool> {
```
oetyng, line 444: I thought by the name that the decision itself was updated, but there was no mut ref and nothing updated.

So this is a check telling us if we need to update a decision?

Name perhaps: `shall_update_decision` or `decision_update_needed` ?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -427,51 +422,47 @@ impl NetworkKnowledge {
...
+        // TODO: Decisioin requires the PublicKeySet to validate,
+        //       however NetworkKnowledge only retains the PublicKey info
+        // if !decision.validate(&self.section_chain()).is_ok() {
+        //     error!("Can't update decision {decision:?}");
+        //     continue;
+        // }
+        Ok(self.section_peers.update_peers(peers))
+    }
+
+    /// Update one decision. Returns whether it actually updated.
```
oetyng, line 443: Are we actually updating a decision?

It seems we are updating our section peers with a new decision?

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -520,9 +511,9 @@ impl NetworkKnowledge {
         self.section_peers.members()
     }
 
-    /// Returns joined members with Decision info.
-    pub fn section_members_with_decision(&self) -> BTreeSet<Decision<NodeState>> {
-        self.section_peers.section_members_with_decision()
+    /// Returns joined members at the specific generation.
```
oetyng, line 514: Nit pick: Aren't members just simply members?
So when a node has left, it is not a member.

I don't think we need an extra qualifier like _joined_ beause that introduces doubts on what exactly "member" means. Can a member not be a part of the section?

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,74 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
 const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 3;
 
 /// Container for storing information about (current and archived) members of our section.
-#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
 pub(super) struct SectionPeers {
-    members: BTreeMap<XorName, Decision<NodeState>>,
+    /// Initial members snapshot at the time of SAP change
+    bootstrap_members: BTreeSet<NodeState>,
```
oetyng, line 29: Could we call this `initial_members` instead?

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,74 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
...
+    /// Decisions received during the current SAP so far
+    decisions: Vec<Decision<NodeState>>,
+    /// Left Decisions archived
```
oetyng, line 32: Are these archived decisions on leaves only?

I think the comment and field can be clearer on that.

(Perhaps also some comment on why we archive leaves specifically and nothing else.)

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,74 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
 const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 3;
 
 /// Container for storing information about (current and archived) members of our section.
-#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
 pub(super) struct SectionPeers {
```
oetyng, line 27: Note: 
There seems to be some more domain logic expressed with `SectionPeers` now.
`SectionPeers` seems to be an instance that is reflecting the span of a single SAP, continuously updated with Adult changes.
I'm thinking of how that can be made clearer with the naming.

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,74 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
...
+    bootstrap_members: BTreeSet<NodeState>,
+    /// Decisions received during the current SAP so far
+    decisions: Vec<Decision<NodeState>>,
+    /// Left Decisions archived
     archive: BTreeMap<XorName, Decision<NodeState>>,
 }
 
 impl SectionPeers {
+    /// Reset the bootstrap members when SAP changed.
+    pub(crate) fn reset_bootstrap_members(&mut self, new_bootstrap_members: BTreeSet<NodeState>) {
```
oetyng, line 38: Okay, I can see now from reading though more code (considering my previous questions on meaning of this fn), that this is a re-instantiation of section peers.

I think it would actually be better for clarity to just replace the instance.
Like so:
`self.section_peers = SectionPeers::new(initial_elders);`

oetyng: Alternatively rename the reset to:

`transition_to_new_elders` (because those are always Elders, right? Never Adults?)

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,74 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
...
+            .cloned()
+            .collect()
+    }
+
+    /// Returns section decisions since last SAP change
+    pub(crate) fn section_decisions(&self) -> Vec<Decision<NodeState>> {
+        self.decisions.clone()
+    }
+
+    /// Returns joined members at the specific generation.
```
oetyng, line 56: Same as previous comment about members and `joined` qualifier.

oetyng: Members are just members. Other nodes are not members.
The impl details is that we identify them as those that saw the event Joined but not the event Left or Relocated.

(Now, `NodeState` is actually in our domain a `NodeEvent` (`Left`, `Joined`, `Relocated` are _past tense_, things that _happened_, i.e. events), and a `NodeState` would be a left fold of such events (replaying them to get current state), giving for example `NodeState::Member`, `NodeState::PreviousMember`, `NodeState::Other`).

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -293,15 +350,17 @@ mod tests {
         );
         let node_relocated_joins = section_signed_to_decision(node_relocated_joins);
 
-        assert!(!section_peers.update(node_left_joins));
-        assert!(!section_peers.update(node_relocated_joins));
+        assert!(!section_peers.update(node_left_joins)?);
+        assert!(!section_peers.update(node_relocated_joins)?);
 
         assert_lists(section_peers.archive.values(), &[node_left, node_relocated]);
```
oetyng, line 356: Wrt the comment on the `archive` field of `SectionPeers` (saying `/// Left Decisions archived`), this code here seems to imply that the archive contains the events of nodes `joined` decided, but we put them in `archive` after we have seen the events of them `left` decided as well as `relocated` decided.

sn_interface/src/network_knowledge/errors.rs
maqi: we are behind hence request an AE update, I will expand the display msg to make it more explicit

sn_interface/src/network_knowledge/mod.rs
maqi: yes. it's just the API is like that.

sn_interface/src/network_knowledge/mod.rs
maqi: > Why is it only called at the time of SAP change

The member changes between the SAP change will be recorded via updates, i.e. stored Decisions

> bootstrap members

Any idea of a better naming?

sn_interface/src/network_knowledge/mod.rs
maqi: Yeah some wording need to be optimized, I will change the naming bit.

sn_interface/src/network_knowledge/mod.rs
maqi: I think it's just to highlight, as there we have `archived_members`
As `member` within `SectionPeers` also have the meaning of `once being a member`

sn_interface/src/network_knowledge/section_peers.rs
maqi: To be more accurate, it's actually `decision contains a leaving node_state (left or relocated)` to be archived.
The reason is to avoid node's rejoin using the same id.
I will update the comment little bit

sn_interface/src/network_knowledge/section_peers.rs
maqi: > to just replace the instance

We still need the `archived_members` from the previous SectionPeer

> rename the reset to:  transition_to_new_elders (because those are always Elders, right? Never Adults?)

No. at the time of the SAP change, all members (elders + adults) at that time becomes `bootstrap_emmbers/initial_members`

sn_interface/src/network_knowledge/section_peers.rs
maqi: https://github.com/maidsafe/safe_network/pull/2124#discussion_r1119911069

sn_interface/src/network_knowledge/section_peers.rs
maqi: `archive` is a btreemap with key of node_name, i.e. although the Decision may contain multiple node_states, the entry is only about one node.
and we only interested in the `node_state of that entry's node_name within the Decision` to be left.

sn_interface/src/network_knowledge/errors.rs
oetyng: `AEOutdated` is the actual error then.
`RequestAEUpdate` is an imperative (something we must do, i.e. a cmd) resulting from us being outdated.

sn_interface/src/network_knowledge/mod.rs
oetyng: Maybe we can just pick that single elder, and make it clear in the naming that we only have and expect one?

sn_interface/src/network_knowledge/mod.rs
oetyng: Yeah I think `bootstrap` is a concept we have specifically for joining a network, and it makes things confusing to reuse the word for other things.

sn_interface/src/network_knowledge/mod.rs
oetyng: Maybe `archived` is simpler and more to the point if just being called `previous_members`. That way there is no way to explain that `archived` means they were previously members.

sn_interface/src/network_knowledge/section_peers.rs
oetyng: I think it would be more specific and clear what we want and are doing if we call a fn `update_members` for example, and pass in the new set `members`.

(there is some elder change, and there could be an adult change as well if it was due to promotion)

I think that would make it all much clearer.

sn_interface/src/network_knowledge/section_peers.rs
oetyng: What I mean is that the archive actually contains `Joined` states (that's what we are comparing against in the test, and are adding to our set for comparison).

So here in the test, it looks like the state is supposed to be (or can be, nothing clear) `Joined`in the archive.

But in the comment on the archive field (and your comments here?) seem to say they should only be `Left`?

maqi: these two lines of assertion no long valid, and there is now a TODO comment put in to explain the reason.

> the archive actually contains Joined states 

It will only happen when a Decision holds multiple node_states, and one of it is a `Left` or `Relocated` . 
Also, for `archive` , for an entry's decision value, we only interested in the node_state that its name matches the entry's name.

sn_interface/src/network_knowledge/mod.rs
```diff
@@ -198,17 +187,10 @@ impl NetworkKnowledge {
...
-            let node_state = NodeState::joined(*peer, None);
-            let sig = create_first_sig(&public_key_set, &secret_key_share, &node_state)?;
-            let _changed = network_knowledge
-                .section_peers
-                .update(section_signed_to_decision(SectionSigned {
-                    value: node_state,
-                    sig,
-                }));
-        }
+        let initial_members: BTreeSet<_> = iter::once(NodeState::joined(peer, None)).collect();
```
davidrusu, line 190: I find the `from_iter(..)` pattern more clear:
```suggestion
        let initial_members = BTreeSet::from_iter([NodeState::joined(peer, None)]);
```
Either way works 👍

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -19,33 +23,68 @@ const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 5;
 const ELDER_CHURN_EVENTS_TO_PRUNE_ARCHIVE: usize = 3;
 
 /// Container for storing information about (current and archived) members of our section.
-#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd)]
+#[derive(Clone, Default, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
```
davidrusu, line 26: I don't think we need the `Serialize, Deserialize` derives anymore since we are no longer including the full SectionPeers in the AEupdate

sn_interface/src/network_knowledge/section_peers.rs
```diff
@@ -78,45 +117,65 @@ impl SectionPeers {
...
+
+        // TODO: validate all the incoming Decisions are signed by current SAP
+        if self.decisions.len() < peers.len() {
+            // Only update to the newer
+            self.decisions = peers;
+            there_was_an_update = true;
+        }
+
+        there_was_an_update
+    }
```
davidrusu, line 174: From slack convo, we want to remove this function since we only want one way to update the section peers decisions.

sn_interface/src/network_knowledge/mod.rs
maqi: `NodeState::joined(peer, None)` is not an iter, hence has to call some `into_iter` if want to using `from_iter` .
so still prefer to use `: BTreeSet<_>` in this case.

sn_node/src/node/messaging/handover.rs
```diff
@@ -330,7 +318,9 @@ impl MyNode {
     fn check_elder_handover_candidates(&self, sap: &SectionAuthorityProvider) -> Result<()> {
         // in regular handover the previous SAP's prefix is the same
         let previous_gen_sap = self.get_sap_for_prefix(sap.prefix())?;
-        let members = self.get_members_at_gen(sap.membership_gen())?;
+
+        // use the current members
+        let members = self.get_current_members()?;
```
davidrusu, line 323: This will cause Handover failures which will lead to failures in elder promotions, you need to use the members at the time that the Sap candidates were created. (i.e. revert back to `members_at_gen(sap.membership_gen())` logic)

sn_node/src/node/messaging/handover.rs
```diff
@@ -356,7 +346,7 @@ impl MyNode {
         // and the same ancestor prefix
         let prev_prefix = sap1.prefix().popped();
         let previous_gen_sap = self.get_sap_for_prefix(prev_prefix)?;
-        let members = self.get_members_at_gen(sap1.membership_gen())?;
+        let members = self.get_current_members()?;
```
davidrusu, line 349: same here.

davidrusu (approved):

PR #2141, opened 2023-02-24
Remove joins allowed voting

grumbach:
Removes joins_allowed voting, refactors section proposals


joshuef:
Local testing: 60 nodes (|10mb large), normal section size. 33 join. One client test run (putting... 170mb of data...) and we're up to 47 nodes joined... Will need to poke a bit more, but on the surface of it, this seems like a fine simplification.

joshuef (approved):

PR #2147, opened 2023-02-28
fix: ae updates and affects split into sap and member updates

joshuef:
We were only updating members if the SAP changed. We were attempting to terminate DKG runs if members changed.

Now those two flows are distinct

### DIFF
--- a/sn_node/src/node/messaging/dkg.rs
+++ b/sn_node/src/node/messaging/dkg.rs
@@ -821,7 +821,7 @@ impl MyNode {
         // it to sign any msg that needs section agreement.
         self.section_keys_provider.insert(key_share.clone());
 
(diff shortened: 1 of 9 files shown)


sn_node/src/node/messaging/dkg.rs
```diff
@@ -821,7 +821,7 @@ impl MyNode {
         // it to sign any msg that needs section agreement.
         self.section_keys_provider.insert(key_share.clone());
 
-        let mut cmds = self.update_on_section_change(&self.context()).await?;
+        let mut cmds = self.update_on_sap_change(&self.context()).await?;
```
grumbach, line 824: Great rename!

grumbach (approved):
Good fix!

grumbach (approved):
🚢

PR #2148, opened 2023-02-28
Improve relcoation logs

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/cmds.rs
+++ b/sn_node/src/node/flow_ctrl/cmds.rs
@@ -203,8 +203,6 @@ pub(crate) enum Cmd {
 
 impl Cmd {
     pub(crate) fn send_msg(msg: NodeMsg, recipients: Peers, context: NodeContext) -> Self {
-        let msg_id = MsgId::new();
-        debug!("Sending msg {msg_id:?} to {recipients:?}: {msg:?}");
         Cmd::send_network_msg(NetworkMsg::Node(msg), recipients, context)
     }
 
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -75,6 +75,7 @@ pub(crate) struct FlowCtrl {
     cmd_sender_channel: Sender<(Cmd, Vec<usize>)>,
     fault_channels: FaultChannels,
     timestamps: PeriodicChecksTimestamps,
+    started_relocating: Option<XorName>,
 }
 
 impl FlowCtrl {
@@ -124,6 +125,7 @@ impl FlowCtrl {
             cmd_sender_channel: cmd_sender_channel.clone(),
             fault_channels,
             timestamps: PeriodicChecksTimestamps::now(),
+            started_relocating: None,
         };
 
         let _handle = tokio::task::spawn(flow_ctrl.process_messages_and_periodic_checks());
--- a/sn_node/src/node/flow_ctrl/periodic_checks.rs
+++ b/sn_node/src/node/flow_ctrl/periodic_checks.rs
@@ -21,8 +21,8 @@ use std::{collections::BTreeSet, sync::Arc, time::Duration};
 use tokio::{sync::RwLock, time::Instant};
 
 const PROBE_INTERVAL: Duration = Duration::from_secs(300);
-const REQUEST_TO_RELOCATE_TIMEOUT_SEC: Duration = Duration::from_secs(5);
-const JOIN_AS_RELOCATED_TIMEOUT_SEC: Duration = Duration::from_secs(5);
+const REQUEST_TO_RELOCATE_TIMEOUT_SEC: Duration = Duration::from_secs(30);
+const JOIN_AS_RELOCATED_TIMEOUT_SEC: Duration = Duration::from_secs(30);
 const MISSING_VOTE_INTERVAL: Duration = Duration::from_secs(5);
 const MISSING_DKG_MSG_INTERVAL: Duration = Duration::from_secs(5);
 // const SECTION_PROBE_INTERVAL: Duration = Duration::from_secs(300);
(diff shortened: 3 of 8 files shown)


sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -178,7 +178,8 @@ impl MyNode {
                 let mut node = node.write().await;
                 trace!("[NODE WRITE]: RelocationRequest write gottt...");
                 trace!("Handling RelocationRequest msg from {sender}: {msg_id:?}");
-                Ok(node.handle_relocation_request(relocation_node, relocation_trigger)?)
+                Ok(node
+                    .handle_relocation_request_from_member(relocation_node, relocation_trigger)?)
```
oetyng, line 182: > handle_relocation_request_from_member
can there be such requests from non-members?

sn_node/src/node/mod.rs
```diff
@@ -337,6 +337,18 @@ mod core {
             ))
         }
 
+        /// Generates a AE probe for our section elders
+        pub(crate) fn generate_section_probe_msg(context: &NodeContext) -> Cmd {
+            // Get prefixes of other sections.
```
oetyng, line 342: Is this a copy paste comment? Doesn't seem to match the code.

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -272,38 +287,30 @@ impl FlowCtrl {
...
-    //             issue: sn_fault_detection::IssueType::AeProbeMsg,
-    //         });
-    //     }
+        // Send a probe message to an elder
+        debug!("Going to health check elders");
+
+        let elders = context.network_knowledge.elders();
+        for elder in elders {
+            // we track a knowledge issue
+            // whhich is countered when an AE-Update is
```
oetyng, line 302: whhich => which

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -131,9 +150,15 @@ impl FlowCtrl {
                         NodeMsg::TryJoin(Some(proof.clone())),
                     ));
                 } else {
```
oetyng, line 152: I always try to keep number of negations as few as possible, so the if clause here would be reversed to

`if context.network_knowledge.is_section_member(&context.name) {`

with line 153-163 in it, and current ln 145-151 within the `else` clause.

sn_node/src/node/messaging/node_msgs.rs
joshuef: nodes joining from other sections was what I thought this was doing. Rename is for clarity on that

oetyng: I think they are handled via the join flow?

Looking that up now.

joshuef: > was supposed to be used as a verifiable proof of the trigger actually giving the relocation of that node

Was it? I thought it was just that the trigger was verifiable from the churnId now? cc @b00f , what's the case there?

oetyng: It was another issue, that's why I removed that part of the comment.

So, the refactor that introduced the node driving the process did not include the change to have the trigger verifiable.

oetyng: I'm raising a PR to clarify this flow with some name updates.

b00f: BeginRelocating and RelocationRequest sounds a bit wired for me also. I think we need to give a better name (if we can't merge them). By the way, relocation trigger has an instance of the decision that is the proof. AFAIK, to verify relocation we need the trigger. So here both items have triggers.

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -269,38 +283,30 @@ impl FlowCtrl {
...
-    //             issue: sn_fault_detection::IssueType::AeProbeMsg,
-    //         });
-    //     }
+        // Send a probe message to an elder
+        debug!("Going to health check elders");
+
+        let elders = context.network_knowledge.elders();
+        for elder in elders {
+            // we track a knowledge issue
+            // which is countered when an AE-Update is
```
b-zee, line 298: This comment doesn't seem to finish the

b-zee: Sorry, this seems like it was already in the code. I thought you added it 😀

sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -97,7 +97,16 @@ impl FlowCtrl {
     async fn enqueue_cmds_for_node_periodic_checks(&mut self, context: &NodeContext) -> Vec<Cmd> {
         let mut cmds = vec![];
 
-        // check if we have been asked to prepare for relocation
+        // here we specifically ask for AE prob msgs and manually
+        // track faults for missing AE, and theerefore hopefully keep members updated...
```
b-zee, line 101:
```suggestion
        // track faults for missing AE, and therefore hopefully keep members updated...
```

b-zee (approved):

b-zee (approved):

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -75,6 +75,7 @@ pub(crate) struct FlowCtrl {
     cmd_sender_channel: Sender<(Cmd, Vec<usize>)>,
     fault_channels: FaultChannels,
     timestamps: PeriodicChecksTimestamps,
+    started_relocating: Option<XorName>,
```
RolandSherwin, line 78: Sorry, I can't really figure out what the above field is doing. Is this a placeholder for a future change?

joshuef: in periodic checks, we log when we start a relocation process (only the first time)

sn_node/src/node/flow_ctrl/cmds.rs
```diff
@@ -203,8 +203,6 @@ pub(crate) enum Cmd {
 
 impl Cmd {
```
RolandSherwin, line 204: Please ignore this. Testing github auto-merge

RolandSherwin, line 204: Please ignore this as well. Testing github auto-merge

PR #2157, opened 2023-03-01
DBC reasons

grumbach:
Adds reasons in DBCs

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -162,12 +162,14 @@ impl Safe {
         let is_spent = spent_transactions.into_iter().any(|tx_hash| {
             let shares_for_current_tx = spent_proof_shares
                 .iter()
-                .filter(|share| share.content.transaction_hash == tx_hash);
+                .cloned()
+                .filter(|share| share.content.transaction_hash == tx_hash)
+                .collect();
 
             verify_spent_proof_shares_for_tx(
                 public_key,
                 tx_hash,
-                shares_for_current_tx,
+                &shares_for_current_tx,
                 &proof_key_verifier,
             )
             .is_ok()
@@ -278,6 +280,7 @@ impl Safe {
         wallet_url: &str,
         amount: &str,
         owner_public_key: Option<bls::PublicKey>,
+        reason: Option<Hash>,
     ) -> Result<Dbc> {
         debug!(
             "Reissuing DBC from wallet at {} for an amount of {} tokens",
@@ -289,6 +292,7 @@ impl Safe {
                 [(amount.to_string(), owner_public_key)]
                     .into_iter()
                     .collect(),
+                reason,
             )
             .await?;
 
@@ -309,6 +313,7 @@ impl Safe {
         &self,
         wallet_url: &str,
         outputs: Vec<(String, Option<bls::PublicKey>)>,
+        reason: Option<Hash>,
     ) -> Result<Vec<Dbc>> {
         let mut total_output_amount = Token::zero();
         let mut outputs_owners = Vec::<(Token, OwnerOnce)>::new();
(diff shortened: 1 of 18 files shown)


sn_api/src/app/wallet.rs
```diff
@@ -1036,6 +1067,26 @@ mod tests {
...
+            .await?;
+
+        let pk = bls::SecretKey::random().public_key();
+        let just_any_xorname = XorName::from_content(&[1, 2, 3, 4]);
+        let any_reason = DbcReason::from(just_any_xorname);
+        let output_dbc = safe
+            .wallet_reissue(&wallet_xorurl, "1", Some(pk), any_reason)
+            .await?;
+
+        assert_eq!(Some(any_reason.into()), output_dbc.reason());
```
grumbach, line 1085: Usage example with XorName as reason (eg: for data payments)

sn_cli/src/subcommands/wallet.rs
```diff
@@ -74,6 +74,10 @@ pub enum WalletSubCommands {
         /// A file path to store the content of the reissued DBC.
         #[clap(long = "save")]
         save: Option<PathBuf>,
+        /// The reason why this DBC is spent
+        /// (Used for data payments: currently not yet implemented)
```
oetyng, line 78: I'd say this is used for more than data payments.

If the content of this field will let us know that the reason is a data payment, then it will also be able to tell us that it is _not_ a data payment. And since the only thing _not_ being a data payment, is a _transfer_, it means it will also be used for _transfer_ payments.

Otherwise, if that is to be derived from some other place, we would have potentially 2 conflicting sources of truth. Better keep it to one place.

So... with that in mind, it might also be good to think of what representations we will actually encode in this field, and how those will connect with the respective logic.

sn_client/src/api/spentbook_apis.rs
```diff
@@ -129,6 +132,7 @@ mod tests {
     use crate::Client;
 
     use sn_dbc::{rng, DbcTransaction, Hash, OwnerOnce, TransactionBuilder};
+    use sn_interface::dbcs::DbcReason;
```
oetyng, line 135: nitpick, but collapse with `ln 36`

sn_interface/src/dbcs/mod.rs
```diff
@@ -7,13 +7,18 @@
...
-    RevealedInput, SpentProofContent, SpentProofShare, Token, TransactionBuilder,
+    rng, Dbc, Error as DbcError, Hash, IndexedSignatureShare, Owner, OwnerOnce, PedersenGens,
+    RevealedCommitment, RevealedInput, SpentProofContent, SpentProofShare, Token,
+    TransactionBuilder,
 };
 
 use std::{fmt::Debug, result};
 use thiserror::Error;
 
+mod reasons;
```
oetyng, line 18: nitpick again, but we would normally put both `mod` declarations and `pub` usings on top of the rest of the usings.

sn_interface/src/dbcs/reasons.rs
```diff
@@ -0,0 +1,55 @@
...
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use serde::{Deserialize, Serialize};
+use sn_dbc::Hash;
+use xor_name::XorName;
+
+use super::Error;
```
oetyng, line 13: nitpick again :)
so order of use declarations: self, super, crate, sn_*, 3rd party,

oetyng (approved):
LGTM

oetyng (approved):

PR #2161, opened 2023-03-01
fix(node): node to update its own members list upon being relocated

bochaco:
This also adds an improvement to the relocation process where we try to finalise an ongoing relocation both when we receive the join approval, or with our periodic checks. 
This prevents from having unnecessary delays in updating the node's internal state, i.e. if the join approval is received we update our state and members right away, without having to wait till the periodic check's timer goes off. If, on the contrary, we missed the join approval msg but we've got an AE msg where we learned we've been approved, then we'll perform the node's internal state update with the periodic check as usual.

### DIFF
--- a/sn_node/src/node/flow_ctrl/periodic_checks.rs
+++ b/sn_node/src/node/flow_ctrl/periodic_checks.rs
@@ -117,7 +117,7 @@ impl FlowCtrl {
         // check if we can request for relocation
         // The relocation_state will be changed into `JoinAsRelocated` once the request has been
         // approved by the section
-        if let Some(RelocationState::PreparingToRelocate(trigger)) = &context.relocation_state {
+        if let RelocationState::PreparingToRelocate(trigger) = &context.relocation_state {
             if self.timestamps.request_to_relocate_check.elapsed() > REQUEST_TO_RELOCATE_TIMEOUT_SEC
             {
                 info!(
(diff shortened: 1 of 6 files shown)


sn_node/src/node/flow_ctrl/periodic_checks.rs
```diff
@@ -143,26 +143,19 @@ impl FlowCtrl {
         }
 
         // check if we can join the dst section
-        if let Some(RelocationState::ReadyToJoinNewSection(proof)) = &context.relocation_state {
+        if let Some(proof) = context.relocation_state.proof() {
```
oetyng, line 146: I think it was neat and good to remove the Option as you did. We could be more consistent if we do just as on line 121, or indeed as it was before, and call:

```
if let Some(RelocationState::ReadyToJoinNewSection(proof)) = &context.relocation_state {
```

Or what do you think?
Then we don't need the extra

```
impl RelocationState {
    pub fn proof(&self) -
```

stuff either.

oetyng: I see the `fn proof() `is reused further down, but it makes it clearer I think to make it explicit that it is when `RelocationState::ReadyToJoinNewSection(proof)` that we call `node.finalise_relocation`.

sn_node/src/node/messaging/node_msgs.rs
```diff
@@ -205,6 +204,11 @@ impl MyNode {
                             target_sap.prefix(),
                         );
 
+                        // If we were relocating finaise it
```
oetyng, line 207: Small typo: `finaise` => `finalise`

joshuef (approved):

PR #2162, opened 2023-03-02
chore(node): membership validation logic allows relocations even when…

joshuef:
… joined...

### DIFF
--- a/sn_interface/src/network_knowledge/node_state/mod.rs
+++ b/sn_interface/src/network_knowledge/node_state/mod.rs
@@ -93,7 +93,7 @@ impl NodeState {
         }
     }
 
-    pub fn validate(
+    pub fn validate_node_state(
         &self,
         prefix: &Prefix,
         members: &BTreeMap<XorName, Self>,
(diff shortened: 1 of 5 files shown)


sn_interface/src/network_knowledge/node_state/mod.rs
```diff
@@ -111,16 +111,19 @@ impl NodeState {
...
-                } else if let Some(existing_node) = members
-                    .values()
-                    .find(|n| n.peer().addr() == self.peer().addr())
-                {
+                } else if let Some(existing_node) = members.values().find(|n| {
+                    // reject if we find a node that is marked as currently 'Joined' AND has same socket.
+                    // (This check is needed to allow nodes that have been Relocated to rejoin this section
+                    // if they've been sent to our section again)
+                    n.peer().addr() == self.peer().addr()
+                        && matches!(n.state, MembershipState::Joined)
```
grumbach, line 123: `members` above come from the call to `self.section_members` which already filters for `MembershipState::Joined`

joshuef: orly 🤔

grumbach (approved):
LGTM

PR #2169, opened 2023-03-02
feat(rewards): integrate reward key to node

oetyng:
This is the first in a number of steps to introduce resource royalties, aka rewards.

In this PR we add the reward key as an essential part of a node/peer, here called `RewardPeer`.
Membership votes will thus be over not only name and address, but also the reward key.
The reward key is by that also following the node on relocations.

Follow up PRs:
2. Elders confirming that spends contain outputs destined for them.
3. Implementing a commonly known deterministic transfer/store cost and distribution to nodes.
4. Including transfer/store cost in outputs when spending DBCs.
5. Elders confirming that the amounts in those outputs are sufficient.


sn_interface/src/types/peer.rs
```diff
@@ -89,3 +117,116 @@ impl PartialOrd for Peer {
         Some(self.cmp(other))
     }
 }
+
+/// A Peer with name, derived from its `PublicKey`, and an address.
```
oetyng, line 121: I'll update this comment.

sn_interface/src/types/peer.rs
```diff
@@ -89,3 +117,116 @@ impl PartialOrd for Peer {
...
+    pub fn addr(&self) -> SocketAddr {
+        self.peer.addr()
+    }
+
+    /// Returns the age.
+    pub fn age(&self) -> u8 {
+        self.peer.age()
+    }
+
+    pub fn reward_key(&self) -> bls::PublicKey {
```
oetyng, line 150: I'll add a clarifying comment here.

sn_interface/src/types/peer.rs
```diff
@@ -89,3 +117,116 @@ impl PartialOrd for Peer {
...
+
+    #[cfg(any(test, feature = "test-utils"))]
+    pub fn random_w_key(public_key: ed25519_dalek::PublicKey) -> Self {
+        Self {
+            peer: Peer::random_w_key(public_key),
+            reward_key: bls::SecretKey::random().public_key(),
+        }
+    }
+
+    pub fn random_w_addr(addr: SocketAddr) -> Self {
```
oetyng, line 178: I agree that it is not stringent to not put these under the feat flag, but it's the use of rand addr generation that requires it.

I can probably think of a better solution, but it's late now :)

sn_node/src/node/flow_ctrl/tests/network_builder.rs
```diff
@@ -140,7 +140,10 @@ impl<R: RngCore> TestNetworkBuilder<R> {
             let socket_addr: SocketAddr = (Ipv4Addr::LOCALHOST, 0).into();
             let (comm, rx) = Comm::new(socket_addr, None).expect("failed to create Comm");
             let mut node = node.clone();
-            node.addr = comm.socket_addr();
+            node.peer = RewardPeer::new(
```
oetyng, line 143: This might look funny, but there is a cause:

> Currently, the TestSAP does not create nodes with a working Comm / Reciever<CommEvent> and hence we don't expect the Comms to be passed into sap_pre_built
Just some utils mismatch thing, it will be fixed to make it coherent.

sn_node/src/node/messaging/join_section.rs
```diff
@@ -212,18 +209,18 @@ mod tests {
...
-        let joiner_peer = joining_node.info().peer();
-        let cmd = MyNode::handle_join(
-            &mut adult,
-            &adult_context,
-            joiner_peer,
+        let joiner_peer = joining_node.info().reward_peer();
+        let cmds = MyNode::handle_join(
+            joiner_peer.peer(),
+            JoinDetails::New(joiner_peer.reward_key()),
             MsgId::new(),
```
oetyng, line 216: Actually.. I think only MsgId::new() is ever passed in here.. so we can probably remove that parameter to this fn.

sn_node/src/node/messaging/join_section.rs
```diff
@@ -212,18 +209,18 @@ mod tests {
...
             None,
-            None,
+            &mut adult,
```
oetyng, line 218: I have in another branch to make this a call on the node instance instead of passing it in as arg. It would shorten + simplify the signature.

sn_node/src/node/mod.rs
```diff
@@ -93,11 +91,11 @@ mod core {
     }
 
     pub(crate) struct MyNode {
+        pub(crate) peer: RewardPeer,
```
oetyng, line 94: Some duplication, agreed (the name is derived from keypair, and addr can be gotten from Comms), but this is a fairly simple way for the logic in the larger scope. And we removed the addr field.

sn_node/src/node/mod.rs
```diff
@@ -796,7 +796,7 @@ mod core {
                     if state.is_relocated()
                     // && state.sig.public_key == context.network_knowledge.section_key()
```
oetyng, line 797: Ah, this can be removed I see..

sn_interface/src/types/peer.rs
```diff
@@ -89,3 +117,116 @@ impl PartialOrd for Peer {
         Some(self.cmp(other))
     }
 }
+
+/// A Peer with name, derived from its `PublicKey`, and an address.
+#[derive(Copy, Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub struct RewardPeer {
```
joshuef, line 123: So i find myself tripping up over this naming. 

What's a `RewardPeer` vs a `Peer`? 

It's that it's a node essentially? providing a reward key?

I'm not entirely sure what the name should be, but my feel is that including `Peer` in there is causing confusion (im confused :P) 

And i guess, what is a node w/ reward key vs any other? So maybe we vote over `Nodes`? 🤷 I dunno. 🧠 ⚡ and my brain isn't particularly ⚡ atm

oetyng: Yeah, I have been thinking about these things a long ti.e, specifically what this "Peer" thing is. To me it looks like a NodeId. The name and the socketaddr is what uniquely identifies a node, am identifyer, i.e. id. And that's how it's used in the code.

Now, I tried first to add the reward key to Peer struct. But that meant that it was included in situations where it wasn't needed, i e. Where we merely want to reference a node by id.

So I added a struct instead, to find the pmaces where it was absolutely needed to include the reward key, for the logic to hold.

Just picked RewardPeer since it is a peer with rewardkey, and was simple for now, and because I think the naming Peer is wrong to begin with, so I thought I can't make that name/struct very same before addressing that.
But the three parts xorname, socketaddr and rewardkey need to go together into `NodeState` for the logic to hold.
I thought it meant least initial co fusion to state that the new type there is just as before, a Peer, but with rewardkey. 

So... PeerWithRewardKey? Just thought it was too long.

I.M.O. `Peer` should be renamed to `NodeId`.

But l such a name change as a follow up, or going in before.

oetyng: And here is that rename: https://github.com/maidsafe/safe_network/pull/2172

sn_node/src/node/messaging/join_section.rs
oetyng: Nopes, that was not the case.

sn_interface/src/network_knowledge/test_utils.rs
```diff
@@ -37,21 +37,32 @@ pub fn gen_addr() -> SocketAddr {
...
 }
 
 // Generate a Peer with the given age and prefix
 pub fn gen_peer_in_prefix(age: u8, prefix: Prefix) -> Peer {
     let name = ed25519::gen_name_with_age(age);
-    Peer::new(prefix.substituted_in(name), gen_addr())
+    Peer::random_w_name(prefix.substituted_in(name))
+}
+
+// Generate a Peer with the given age
```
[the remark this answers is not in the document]
oetyng: mm that was a copy paste from the above, will fix all of them

grumbach (commented):
Nice one changing `Peer` to `NodeId` I think it makes more sense indeed. 
On the other hand, `PaidNodeId` is confusing and the field/var name `node_id` is being use interchangeably for `NodeId` and `PaidNodeId` adding to the confusion. I'd be in favour of renaming that one to something completely different. Why not `NetworkNode`, `PeerNode`, or even `Peer`?
Overall LGTM

sn_interface/src/network_knowledge/node_state/mod.rs
```diff
@@ -146,7 +146,11 @@ impl NodeState {
...
     }
 
-    pub fn node_id(&self) -> &NodeId {
+    pub fn node_id(&self) -> NodeId {
+        NodeId::new(self.name(), self.addr())
+    }
+
+    pub fn paid_node_id(&self) -> &PaidNodeId {
         &self.node_id
     }
```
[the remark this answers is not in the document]
oetyng: Even with the rename to `RewardedNode`, in some places it will make sense to call a variable `id` or `node_id`, since it is still a node id, and used like that, it just has the reward key as well. `RewardedNode` is not a great name either since it's still an ID more than a NODE.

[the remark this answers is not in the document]
oetyng: I think `RewardNodeId` makes most sense here so far..

PR #2171, opened 2023-03-03
Chore update logs around relocation

joshuef:


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -309,7 +307,17 @@ impl FlowCtrl {
     ) {
         let _handle = tokio::task::spawn(async move {
             while let Some(event) = incoming_msg_events.recv().await {
-                debug!("CommEvent received: {event:?}");
+                let capacity = cmd_channel_for_msgs.capacity();
+
+                if capacity < 30 {
+                    warn!("CmdChannel capcity severely reduced");
```
oetyng, line 313: Typo: `capcity`

I think we could probably scour the code though and do a single big PR fixing all typos all over :)

oetyng (dismissed):

PR #2172, opened 2023-03-03
refactor: distinguish client and node ids

oetyng:
This removes the ambiguous `Peer` concept and creates a clearer path for client and node related logic, while finally resolving the node/peer ambiguity. Also, the peer struct is and has always been an identifier for a client or node. For that reason it is now named accordingly.

A peer could previously be both a node and a client (id). But a client doesn't have an age, and we don't want node logic paths muddled. So now we only have `ClientId` and `NodeId`, and in the few places where we don't know which of it it is (or it is shared code) it is now called a `Participant` (as in participating in the network). That name can probably be improved, but is an OK start.

This PR will give a better structure in the introduction of reward key in the [current reward key integration PR](https://github.com/maidsafe/safe_network/pull/2169).

### DIFF
--- a/sn_interface/src/types/identities/client_id.rs
+++ b/sn_interface/src/types/identities/client_id.rs
@@ -14,16 +14,16 @@ use std::{
 };
 use xor_name::XorName;
 
-use super::utils::calc_age;
+use super::Participant;
 
-/// A Peer with name, derived from its `PublicKey`, and an address.
+/// The id is the name, derived from its `PublicKey`, and the address of a client.
 #[derive(Copy, Clone, Debug, serde::Serialize, serde::Deserialize)]
-pub struct Peer {
+pub struct ClientId {
     name: XorName,
     addr: SocketAddr,
 }
 
-impl Peer {
+impl ClientId {
     pub fn new(name: XorName, addr: SocketAddr) -> Self {
         Self { name, addr }
     }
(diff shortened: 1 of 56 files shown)


sn_interface/src/types/identities/client_id.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    pub fn addr(&self) -> SocketAddr {
+        self.addr
+    }
+
+    pub fn from(sender: Participant) -> Self {
+        Self::new(sender.name, sender.addr)
+    }
+
+    // pub fn from(addr: SocketAddr, public_key: ed25519_dalek::PublicKey) -> Self {
```
joshuef, line 43: tiny cleanup here

sn_interface/src/types/identities/mod.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    pub fn from_node(id: NodeId) -> Participant {
+        Self::new(id.name(), id.addr())
+    }
+
+    pub fn from_client(id: ClientId) -> Participant {
+        Self::new(id.name(), id.addr())
+    }
+
+    // pub fn from(addr: SocketAddr, public_key: ed25519_dalek::PublicKey) -> Self {
```
joshuef, line 51: here too

sn_node/src/node/connectivity.rs
```diff
@@ -9,49 +9,48 @@
...
+use sn_interface::types::Participant;
 
 use std::collections::BTreeSet;
 use xor_name::XorName;
 
 impl MyNode {
-    /// Handle error in communication with peer.
-    pub(crate) fn handle_comms_error(&self, peer: Peer, error: sn_comms::Error) {
+    /// Handle error in communication with node.
+    pub(crate) fn handle_comms_error(&self, dst: Participant, error: sn_comms::Error) {
```
joshuef, line 19: i'd watch out calling this `dst` as we have a `Dst` struct. Could just be `participant`?

sn_node/src/node/flow_ctrl/cmds.rs
```diff
@@ -170,21 +169,27 @@ pub enum Cmd {
     SendMsg {
         msg: NetworkMsg,
         msg_id: MsgId,
-        recipients: Peers,
+        recipients: Recipients,
     },
+    // /// Performs serialisation and signing and sends the msg.
```
joshuef, line 174: 🧹

sn_node/src/node/flow_ctrl/cmds.rs
```diff
@@ -193,26 +198,37 @@ pub enum Cmd {
...
 
 impl Cmd {
-    pub(crate) fn send_msg(msg: NodeMsg, recipients: Peers) -> Self {
+    // this fn shall be fixed to be more type safe so that unnecessary casting upstream is avoided
+    pub(crate) fn send_msg(msg: NodeMsg, recipients: Recipients) -> Self {
         Cmd::send_network_msg(NetworkMsg::Node(msg), recipients)
     }
 
-    pub(crate) fn send_network_msg(msg: NetworkMsg, recipients: Peers) -> Self {
+    // pub(crate) fn send_msg(msg: NodeMsg, recipients: BTreeSet<NodeId>) -> Self {
```
joshuef, line 221: more tidy 🧹

joshuef (approved):

PR #2173, opened 2023-03-05
chore: rename `sn_node` binary to `safenode`

jacderida:
- 41130e9f1 **chore: rename `sn_node` binary to `safenode`**

  This also updates the CI and release processes to refer to `safenode` rather than `sn_node`.

  The assets being attached to the release are also now being renamed from `sn_cli` and `sn_node` to
  `safe` and `safenode`, respectively.

- f9459f020 **feat: reworked version of testnet binary**

  The `testnet` binary was reworked to remove its dependency on `sn_launch_tool`. Since it will now be
  used as a replacement for the `node run-baby-fleming` command, the interface was updated to make it
  a bit more usable. Now it does not attempt to build the node by default: the `--build-node` flag is
  used for that. The other arguments that were added should be fairly self explanatory and can be seen
  on the diff.

  All the support from the original binary is retained in this rework, including the use of flamegraph
  and joining an existing network.

  There is a module in the crate that exposes a couple of structs that can be used for launching a
  testnet: `TestnetBuilder` and `Testnet`. The functions on `Testnet` should have almost full code
  coverage, but since the builder is a simple utility, I didn't bother putting tedious tests around
  it. Since the bulk of the work has been unit tested I also haven't added any integration tests for
  testing the code in the `main` module for the binary, since it's pretty simple and doesn't really
  seem worth the additional testing infrastructure.

  The merge CI workflow was updated to use the reworked version of the binary. Since we already build
  the node in the previous step, we just pass the path to it, rather than attempt to rebuild.

  The script for verifying node joins was modified to check for IPs matching "0.0.0.0" in addition to
  "127.0.0.1", which was being caused by the use of the `--local-addr` argument. It also now considers
  the total number of nodes to be the genesis node plus the number of non-genesis nodes. In addition,
  the script was updated to split long commands over multiple lines, use newer-style Bash syntax and
  consistent indentation size.

- a2e728caf **chore: remove the `node` command from `safe`**

  BREAKING CHANGE: the `node` command is entirely removed.

  As agreed previously between both the dev team and community users, the `node` wrapper command is
  entirely removed in favour of using the `safenode` binary directly.

  All the associated code and helper code was removed. This functionality will now be provided by the
  `safeup` and `testnet` tools.

- 7539087c0 **chore: address pr feedback**

  Made some changes that were requested:

  * Tidy `use` statements
  * Define a node launch interval constant in one place
  * Define the `safenode` binary name in one place
  * Define the `sn-node-genesis` name as a constant
  * Define the `local-network-dir` name as a constant
  
  Also fixed a few test assertions that were checking for the wrong value and updated the benchmark
  workflow to correctly launch a local testnet.


bochaco (dismissed):

sn_client/examples/churn.rs
```diff
@@ -69,112 +63,33 @@ async fn main() -> Result<()> {
...
+        .join("release")
+        .join(SAFE_NODE_EXECUTABLE);
+    let (mut testnet, network_contacts_path) = Testnet::configure()
+        .node_bin_path(node_bin_path)
+        .node_launch_interval(INTERVAL_IN_MS)
+        .clear_nodes_dir()
+        .build()?;
+    testnet.launch_genesis(None, vec![])?;
+    testnet.launch_nodes(node_count, &network_contacts_path, vec![])?;
+    testnet.configure_network_contacts(&network_contacts_path)?;
```
oetyng, line 83: So nice simplification of setup here.

sn_client/examples/churn.rs
```diff
@@ -69,112 +63,33 @@ async fn main() -> Result<()> {
...
-    sn_launch_tool_args.push(additional_node_count_str);
-    debug!("Adding testnet nodes with args: {:?}", sn_launch_tool_args);
-
-    // We can now call the tool with the args
-    info!("Adding nodes to the local Safe network...");
-    Launch::from_iter_safe(&sn_launch_tool_args)
-        .map_err(|error| eyre!(error))
-        .and_then(|launch| launch.run())
-        .wrap_err("Error adding nodes to the testnet")?;
+    testnet.launch_nodes(ADDITIONAL_NODE_COUNT, &network_contacts_path, vec![])?;
```
oetyng, line 91: And here. ❤

testnet/src/lib.rs
```diff
@@ -0,0 +1,963 @@
...
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use assert_fs::prelude::*;
+    use color_eyre::Result;
+    use mockall::predicate::*;
+
+    const NODE_LAUNCH_INTERVAL: u64 = 0;
```
oetyng, line 374: I wonder if some of the strings that are frequently used in the tests should be local constants?

Like `"local-test-network"`, `"sn-node-genesis"`, `"safenode"`, etc.

testnet/src/lib.rs
```diff
@@ -0,0 +1,963 @@
...
+                Some(network_contacts_path),
+                node_args.clone(),
+            )?;
+            let launch_bin = self.get_launch_bin();
+            self.launcher.launch(&launch_bin, launch_args)?;
+
+            if i < end {
+                info!(
+                    "Delaying for {} seconds before launching the next node",
+                    self.node_launch_interval / 1000
```
oetyng, line 269: This is old stuff, but I've seen the interval being modified with a number here and there. It seems we for some reason want something else than node launch interval.

I wonder if those places could get a named variable with some docs, so we know why suddenly some other interval is used here or there?

testnet/src/main.rs
```diff
@@ -0,0 +1,281 @@
...
+    trivial_numeric_casts,
+    unused_extern_crates,
+    unused_import_braces,
+    unused_qualifications,
+    unused_results
+)]
+
+use clap::Parser;
+use color_eyre::{eyre::eyre, Help, Result};
+use std::path::PathBuf;
```
oetyng, line 32: Can be collapsed into ln 33.

.github/workflows/merge.yml
```diff
@@ -242,15 +242,11 @@ jobs:
...
-        timeout-minutes: 60
-
-      - name: Start the network
-        run: ./target/release/testnet --interval 30000
+      - name: Start a local network
+        run: cargo run --release --bin testnet -- --interval 30000 --node-path ./target/release/safenode
         id: section-startup
         env:
-          RUST_LOG: "sn_node,sn_comms,sn_consensus,sn_fault_detection,sn_interface=trace"
+          RUST_LOG: "safenode,sn_node,sn_comms,sn_consensus,sn_fault_detection,sn_interface=trace"
```
oetyng, line 249: Is the intention perhaps to replace `sn_node` in these declarations? I see it is replaced in some places further down.

oetyng (approved):
Great work! On many levels.
I really like how you write the docs, written as they should IMO. Thoroughly, describing what it is, how it is used and some motivations on why this or that choice has been made/not made.

.github/workflows/merge.yml
jacderida: As mentioned on Slack, this was intentional.

PR #2178, opened 2023-03-07
feat(rewards): add bls pub key to node

oetyng:
This is the first step to simple dbc tx fees and rewards for Elders.
It allows Elders to check if dbc transactions contain any output for them.

***

Follow up PRs:

- Including Elders in outputs (any value) when spending DBCs.


grumbach (approved):
LGTM

PR #2180, opened 2023-03-08
feat(rewards): add elders to dbc outputs

oetyng:
1.1 `Previous PR`: The query and handling of it on Elders for their reward keys and hard coded fee (#2184).
1.2 `This PR`: ✔️ The logic for including Elders to outputs (were they to be given keys and fee from Elders, see 1.3 below).
1.3 `Next PR`: Connecting the two above, by the client logic for querying Elders for their reward keys and fee (#2189).

Follow up PRs:

- <s>Including Elders in outputs (any value) when spending DBCs.</s>

### DIFF
--- a/sn_api/src/app/wallet.rs
+++ b/sn_api/src/app/wallet.rs
@@ -325,6 +325,7 @@ impl Safe {
                     "Output amount to reissue needs to be larger than zero (0).".to_string(),
                 ));
             }
+
             total_output_amount =
                 total_output_amount
                     .checked_add(output_amount)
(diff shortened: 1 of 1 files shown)


sn_api/src/app/wallet.rs
```diff
@@ -386,43 +441,137 @@ impl Safe {
...
+            // Total fee paid to all recipients in the section for this input.
+            let fee_per_input = mint_fees
+                .values()
+                .fold(Some(Token::zero()), |total, fee| {
+                    total.and_then(|t| t.checked_add(*fee))
+                })
+                .ok_or_else(|| Error::DbcReissueError(
+                    "Overflow occurred while summing the individual Elder's fees in order to calculate the total amount for the output DBCs."
+                        .to_string(),
+                ))?;
```
grumbach, line 494: For now instead of using a `fee_per_input` could it be possible to have a single placeholder fee output in the transaction. Once we decide on fees, we can change it to multiple or anything else. 
My point is that the fees for each inputs idea is not economically viable so not worth putting in the draft. 

For context: 
- minimum fee is 1$
- elders gets 1$ DBC fee
- elders wants to spend that 1$
- elder need to pay 1$ to spend 1$
- small fees DBCs at the end of the line become worthless because they cost as much to spend as their value

grumbach (changes requested):
Good PR, but can we remove the fee per inputs and replace it by just a fee DBC for now as the "fee per input" is not economically viable?

sn_api/src/app/wallet.rs
```diff
@@ -349,6 +350,56 @@ impl Safe {
...
+                &safeurl,
+                &change_dbc,
+                format!("change-dbc-{}", &hex::encode(change_dbc.hash())[0..8]),
+            )
+            .await?;
+        }
+
+        // (virtually) remove input DBCs in the source wallet
+        self.multimap_remove(&safeurl.to_string(), input_dbcs_entries_hash)
+            .await?;
```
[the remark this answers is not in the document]
oetyng: IDK, didn't  write it.

I think he ment that it's removed in the local wallet

oetyng:
> Good PR, but can we remove the fee per inputs and replace it by just a fee DBC for now as the "fee per input" is not economically viable? 

I think it's okay for now.

It burns the money so it is still working as a fee, albeit not as a reward, and gets things going.

Refactor to 1 per dbc can be a follow up.

sn_api/src/app/wallet.rs
oetyng: It's iterative this feature so no need to make it work as it should directly.
The fee works as a fee (burns tokens), albeit not as a reward.

What you propose will come in, but it takes a bit more time.

In the meanwhile we can have _something_ in place, which is what iterative is about.

grumbach (approved):
🚢

PR #2185, opened 2023-03-09
feat(node): push non-blocking Cmds off thread

joshuef:
Moves heaps of non critical Cmds out of the blocking mut MyNode context. Allows them to run w/ latest context; only pushing a subset of Cmds into the blocking process loop

### DIFF
--- a/sn_node/src/node/flow_ctrl/dispatcher.rs
+++ b/sn_node/src/node/flow_ctrl/dispatcher.rs
@@ -23,9 +23,9 @@ impl MyNode {
         let start = Instant::now();
         let cmd_string = format!("{cmd}");
         let result = match cmd {
-            Cmd::TryJoinNetwork => Ok(MyNode::try_join_section(context, None)
+            Cmd::TryJoinNetwork => MyNode::try_join_section(context, None)
                 .into_iter()
-                .collect()),
+                .collect(),
             Cmd::UpdateCaller {
                 caller,
                 correlation_id,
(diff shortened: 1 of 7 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -69,10 +70,20 @@ impl RejoinReason {
     }
 }
 
+/// All Cmds should be passed via an event. This determines if to run in blocking
+/// context or no
+#[derive(Debug)]
+pub(crate) enum CmdProcessorEvent {
+    /// Process this cmd, in blocking thread of off htead where possible
```
b-zee, line 77:
```suggestion
    /// Process this cmd, in blocking thread or off-thread where possible
```

sn_node/src/node/flow_ctrl/dispatcher.rs
```diff
@@ -235,23 +235,59 @@ impl MyNode {
...
+                        proof.new_age(),
+                        self.info().age()
+                    );
+                info!("We've joined a section, dropping the relocation proof.");
+                self.relocation_state = None;
+                None
+            }
+        } else {
+            None
+        }
```
b-zee, line 291: Not sure if this is a good suggestion, but I like returning early and reducing the indentation a level if possible (cognitive load). Rust doesn't seem particularly fit for this though. Do with it what you want.

```suggestion
        // check if we can join the dst section
        let proof = match &self.relocation_state {
            Some(RelocationState::ReadyToJoinNewSection(proof)) => proof,
            _ => return None,
        }
        
        let prev_name = proof.previous_name();
        let new_name = proof.new_name();
        if !self.network_knowledge.is_section_member(&new_name) {
            info!("Periodic check: sending request to join the section as a relocated node");
            Some(MyNode::send_to_elders_await_responses(
                self.context(),
                NodeMsg::TryJoin(Some(proof.clone())),
            ))
        } else {
            info!(
                    "{} for previous node: {:?} of age: {:?}: The new name is: {:?}, and new age is: {:?} (and pass in context age: {:?})",
                    LogMarker::RelocateEnd,
                    prev_name,
                    proof.previous_age(),
                    new_name,
                    proof.new_age(),
                    self.info().age()
                );
            info!("We've joined a section, dropping the relocation proof.");
            self.relocation_state = None;
            None
        }
```

b-zee (changes requested):

oetyng (dismissed):
Nice to get blocking / non blocking separated!

oetyng (approved):

PR #2189, opened 2023-03-12
feat(rewards): send fee query to elders

oetyng:
This PR adds concurrent querying of each elder for their respective reward
key and fee, and replaces the dummy fee with the result from the implemented query.

***

1.1 `Previous PR`: The query and handling of it on Elders for their reward keys and hard coded fee (#2184).
1.2 `Previous PR`: The logic for including Elders to outputs (#2180).
1.3 `This PR`: ✔️ Connecting the two above, by the client logic for querying Elders for their reward keys and fee.

PRs in the overall feature:
- <s> 1. Including Elders in outputs (any value) when spending DBCs.</s>

### DIFF
--- a/sn_client/src/api/spend_queries.rs
+++ b/sn_client/src/api/spend_queries.rs
@@ -1,0 +1,161 @@
+// Copyright 2023 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
+use super::Client;
+use crate::errors::{Error, Result};
+
+use sn_dbc::{PublicKey, Token};
+use sn_interface::{
+    messaging::{
+        data::{ClientMsg, DataQuery, Error as ErrorMsg, QueryResponse, SpendQuery},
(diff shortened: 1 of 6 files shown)


sn_client/src/api/spend_queries.rs
```diff
@@ -0,0 +1,153 @@
...
+            randomization_factor: 1.5,
+            ..Default::default()
+        };
+
+        // this seems needed for custom settings to take effect
+        backoff.reset();
+
+        let dst = query.dst_name();
+
+        loop {
```
grumbach, line 104: Instead of an infinite loop, why don't we set a fixed amount of retries and an error msg if they all fail?

grumbach (commented):
Good one, side note, I'm wondering if we should still push towards the fee per input way?

oetyng:
> Good one, side note, I'm wondering if we should still push towards the fee per input way?

It's a step forward, the other stuff is perfect follow up.

sn_client/src/api/spend_queries.rs
oetyng: There was a fixed retry number. But we have the query timeout and backoff, which gives a number of retries, so it seems we are controlling from different places when we do like that.
Just seemed like that was enough?

grumbach: Even though it is possibly not an infinite loop now, a refactor might accidentally make it so. 
I prefer avoiding loop statements altogether unless I explicitly want an infinite loop.

oetyng: I agree with avoiding such loop statements, usually do it. We have the same pattern like all over though, this is copied from the other query pattern.

grumbach (approved):
🚢

grumbach (approved):
🚢

PR #2193, opened 2023-03-14
fix(node): update flowctrlcmd NodeContext during join

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -98,7 +98,7 @@ impl FlowCtrl {
         incoming_msg_events: Receiver<CommEvent>,
         data_replication_receiver: Receiver<(Vec<DataAddress>, NodeId)>,
         fault_cmds_channels: (Sender<FaultsCmd>, Receiver<FaultsCmd>),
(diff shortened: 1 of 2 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -207,6 +209,14 @@ impl FlowCtrl {
                         rejoin_network_tx.clone(),
                     )
                     .await;
+
+                // update our context in flow ctrl cmd processor with each cmd
+                flow_ctrl_cmd_processing
+                    .send(FlowCtrlCmd::UpdateContext(node.context()))
```
maqi, line 215: Just wondering that this `UpdateContext` seems shall only get called if the `cmd` caused any update to the node's network knowledge ?
The current approach seems will cause many un-necessary calls ? 
though may not be a burden?

joshuef: I'm not sure how to check that.... we'd be comparing the whole node in the current setup. Maybe that's fast? Seemed like it was easier to assume it changed and just update it.

I don't disagree it could perhaps be improved, but it's an issue in the normal flow (which already exists) too, so maybe we can address outwith of this pr

maqi (approved):

PR #2197, opened 2023-03-15
chore(deps): bump consensus

davidrusu:
This should slim down the size of consensus decisions substantially. Validation should be much quicker as well.


sn_interface/src/network_knowledge/mod.rs
```diff
@@ -161,10 +161,9 @@ pub struct NetworkKnowledge {
 fn section_signed_to_decision(section_signed: SectionSigned<NodeState>) -> Decision<NodeState> {
     let mut proposals = BTreeMap::new();
     let _ = proposals.insert(section_signed.value, section_signed.sig.signature);
-    Decision::<NodeState> {
-        votes: BTreeSet::new(),
+    Decision {
+        generation: 0, // TODO
```
maqi, line 165: this whole function will be removed anyway.

maqi (dismissed):

maqi (dismissed):

PR #2205, opened 2023-03-16
Further concurrency

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -155,7 +155,7 @@ impl FlowCtrl {
         Self::send_out_data_for_replication(
             node_context.data_storage,
             data_replication_receiver,
-            blocking_cmd_sender_channel.clone(),
+            flow_ctrl_cmd_sender.clone(),
         )
         .await;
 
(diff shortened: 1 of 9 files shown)


sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -90,7 +90,7 @@ impl FlowCtrl {
 
         // Our channel to process _all_ cmds. If it can, they are processed off thread with latest context,
         // otherwise they are sent to the blocking process channel
-        let (flow_ctrl_cmd_sender, flow_ctrl_cmd_reciever) = mpsc::channel(STANDARD_CHANNEL_SIZE);
```
RolandSherwin: This commit is redundant as the change is being reverted back in the following [commit](https://github.com/maidsafe/safe_network/pull/2205/commits/eedb19b1b95df2d5de7731e328828fb0c6d2f908).

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -121,7 +121,7 @@ impl FlowCtrl {
         };
 
         // first start listening for msgs
-        let cmd_channel_for_msgs = blocking_cmd_sender_channel.clone();
```
RolandSherwin, line 124: A very small one but we can probably save a "clone" here by directly cloning at the function call.

sn_node/src/node/messaging/mod.rs
```diff
@@ -75,14 +75,28 @@ impl MyNode {
...
+        let sender_addr = sender.addr();
         // alternatively we could flag in msg kind for this...
         // todo: this sender is actually client + forwarder ip....
 
-        // we've forwaded it to ourselves as we're the holder. This prevents a loop.
-        // TODO: cut that wee loop down
-        let is_from_us = sender.addr() == context.info.addr;
+        // It's been forwarded, by another node, or ourself. This prevents repeatdly
+        // forwarding client msgs
+        let is_from_elder_node = !context
```
oetyng, line 84: I don't get this.
We say that it is from an elder node if none of our members sent it?

oetyng: I see this further down
`debug!("{msg_id:?} is from another network node?`

So, is the variable meant to say that it is a msg from another section?

joshuef: Aye, this makes no sense atm. I thought I'd sorted something in another branch with this but pulled it in, but it's wrong wrong wrong. 

Removed this change altogether 👍 🙇

RolandSherwin (approved):

RolandSherwin (approved):

PR #2209, opened 2023-03-16
feat(node): exposing a gRPC interface on safenode bin 

bochaco:


sn_interface/proto/safenode/safenode.proto
```diff
@@ -0,0 +1,69 @@
...
+  string log_dir = 3;
+}
+
+message SectionMembersRequest {}
+
+message SectionMembersResponse {
+  repeated Peer peers = 1;
+}
+
+message Peer {
```
oetyng, line 53: Since we've moved away from `Peer`, this should probably be called `Node`.

The member `node_name` can then be just `name`.

And the member of `SectionMembersResponse` then `repeated Node nodes`.

sn_node/src/bin/safenode/main.rs
```diff
@@ -29,21 +29,35 @@
...
+    time::Duration,
+};
 use tracing::{self, error, info, warn};
 
 const JOIN_TIMEOUT_SEC: u64 = 60;
 const JOIN_TIMEOUT_RETRY_TIME_SEC: u64 = 30;
 const JOIN_DISALLOWED_RETRY_TIME_SEC: u64 = 60;
 
+#[derive(Debug)]
+enum NodeCtl {
```
oetyng, line 56: I think this should be called `NodeCtrl`, notice the `r` in there.

> On keyboards that use English abbreviations for key labeling, it is usually labeled Ctrl (rarely, Control or Ctl is seen)

PR #2210, opened 2023-03-17
fix(testnet): move os cond. to if-clause body

oetyng:
Running cargo clippy on win machine was borked with the conditional
over an error in a fn, as the following code became unreachable.

Placing it in the if clause body removes this clippy error, as the
rest of the code in that can still be executed if the flame arg was not
set.

### DIFF
--- a/sn_testnet/src/main.rs
+++ b/sn_testnet/src/main.rs
@@ -27,9 +27,10 @@
     unused_results
 )]
 
+use sn_testnet::{Testnet, DEFAULT_NODE_LAUNCH_INTERVAL, SAFENODE_BIN_NAME};
+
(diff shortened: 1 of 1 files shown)


jacderida:
Thanks for this. Is there any chance we could use the conditional compilation to have two versions of this function, one for Windows and one for everything else? The Windows one would just return an error, as with the intention of the original code. Just to indicate to Windows users that they can't use flamegraph.

oetyng:
> Is there any chance we could use the conditional compilation to have two versions of this function, one for Windows and one for everything else? The Windows one would just return an error

Absolutely, will fix!

jacderida (approved):

PR #2215, opened 2023-03-19
feat(fees): send fees as blinded amounts

oetyng:
The fee amount is now blinded all the way between Elders and Clients.
The ciphers over derivation index and blinding factor is included in the spend request 
sent to elders. By that, an Elder can now verify that it has been paid the required fee amount.

**NB**: The elders do not yet verify that they have been paid (that is done in PR #2238).


oetyng:
Superseded by #2238.

PR #2218, opened 2023-03-20
Improve periodics

joshuef:

### DIFF
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -283,13 +283,12 @@ impl FlowCtrl {
                 context = node.context();
             }
 
+            self.perform_periodic_checks(&context).await;
             // Finally update our context in read only processor with each cmd
             cmd_processing
                 .send(FlowCtrlCmd::UpdateContext(context))
                 .await
                 .map_err(|e| Error::TokioChannel(format!("cmd_processing send failed {:?}", e)))?;
-
-            self.perform_periodic_checks(&node).await;
         }
 
         Ok(())
--- a/sn_node/src/node/flow_ctrl/mod.rs
+++ b/sn_node/src/node/flow_ctrl/mod.rs
@@ -93,7 +93,7 @@ impl FlowCtrl {
         // otherwise they are sent to the blocking process channel
         let (flow_ctrl_cmd_sender, flow_ctrl_cmd_reciever) = mpsc::channel(STANDARD_CHANNEL_SIZE);
         // separate channel to update the periodics loop
-        let (periodic_flow_ctrl_cmd_sender, periodic_flow_ctrl_cmd_reciever) =
+        let (context_updater_for_periodic, context_receiver_for_periodics) =
             mpsc::channel(STANDARD_CHANNEL_SIZE);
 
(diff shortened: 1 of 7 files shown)


joshuef:
this is atop #2220

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -92,6 +92,9 @@ impl FlowCtrl {
         // Our channel to process _all_ cmds. If it can, they are processed off thread with latest context,
         // otherwise they are sent to the blocking process channel
         let (flow_ctrl_cmd_sender, flow_ctrl_cmd_reciever) = mpsc::channel(STANDARD_CHANNEL_SIZE);
+        // separate channel to update the periodics loop
+        let (periodic_flow_ctrl_cmd_sender, periodic_flow_ctrl_cmd_reciever) =
```
maqi, line 96: the name better to be `context_update_for_periodic_check_sender`  & `context_update_for_periodic_check_sender`

the current name sounds bit like the flow_ctrl_cmd is periodic

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -244,11 +266,11 @@ impl FlowCtrl {
     /// This loop processes cmds pushed via the CmdChannel and
     /// runs the periodic events internal to the node.
     async fn process_blocking_cmds(
-        mut self,
         mut node: MyNode,
         cmd_ctrl: CmdCtrl,
         mut incoming_cmds_from_apis: Receiver<(Cmd, Vec<usize>)>,
         cmd_processing: Sender<FlowCtrlCmd>,
+        periodic_cmd_sender: Sender<FlowCtrlCmd>,
```
maqi, line 273: this is actually `context_update_for_periodic_check_sender`

sn_node/src/node/flow_ctrl/mod.rs
```diff
@@ -400,6 +427,32 @@ impl FlowCtrl {
...
+        let mut context = context;
+
+        // now loop and check periodics with context...
+        let _handle = tokio::task::spawn(async move {
+            loop {
+                while let Ok(FlowCtrlCmd::UpdateContext(new_context)) =
+                    periodic_flow_ctrl_cmd_reciever.try_recv()
+                {
+                    debug!("Periodic context updated");
+                    context = new_context;
```
maqi, line 446: this will actually hold the `context` during the sleep period.

what if we pass down the `node` directly, then the code will be:
```rust
    fn listen_for_flow_ctrl_cmds_for_periodics(
        mut self,
        node: MyNode,
    ) {
        // Put a termination flag in if we don't want it to run forever
        while (true) {
            self.perform_periodic_checks(&node.context()).await;
            tokio::time::sleep(Duration::from_millis(100)).await;
        }
    }
```
and we don't need that `periodic_flow_ctrl_cmd_sender` `periodic_flow_ctrl_cmd_reciever` stuff anymore.

joshuef: the node itself is needed in the `mut` blocking cmd loop, so we can't have that here as well

maqi (dismissed):

PR #2238, opened 2023-03-23
feat(fees): verify fee amount at elder

oetyng:
This is done by decrypting the fee ciphers sent from client in a spend,
and verifying that the spend outputs contain an equivalent amount
to the key derived from the elder reward key.

**NB**: Tests must be updated to handle this.


oetyng:
Superseded.

PR #2239, opened 2023-03-23
feat(fees): impl supply/demand-adjusted fee

oetyng:
This PR adds an algorithm to calculate a fee with supply/demand adjusting properties - although indirect and therefore arguably sluggish - based on:
1. Network size (contributing to the deflationary character of SNT).
2. Storage used.
3. Requirements for space margin (1/3).
4. Requirement for as low fee as possible for as long as possible.

(Point 3. and 4. are achieved by the specific design of the required tokens curve.)

***

To do transfers on the network, clients are required to pay a number of tokens as a fee. The fees are the source of rewards for Elders in a payment-network, which is essential to the network security as it gives incentive to node operators to run a node and handle the requests, thus allowing for the network to grow.

**NB 1**: This is not a replacement for a true market, where node operators and clients discover the price based on operation costs + margins and service value + affordability to client.

***

1.  ✔️ `Previous feat` Including Elders in outputs (any value) when spending Dbcs.
2. `Pending PR`: Elders confirming that spends contain outputs with sufficient amount, destined for them: #2238.
3. `This PR`: Use a fee calc algo with supply/demand adjusting properties.

Steps in the fee/reward feature:
- <s> 1. Including Elders in outputs (any value) when spending Dbcs.</s>
- <s> 2. Elders confirming that spends contain outputs with sufficient amount, destined for them.</s>
- <s> 3. Implementing a commonly known deterministic transfer/store cost and distribution to nodes.</s>
- <s> 4. Including supply/demand adjusting fee in outputs when spending Dbcs.</s>


oetyng:
Superseded.

PR #2248, opened 2023-03-27
fixes to resolve CI failures due to network startup problem

maqi:
This PR contains fixes to resolve CI faiures due to network startup problem:
1, do not cast outdated handover vote after aggregated
    This will cause blocking to followup votes due to ChangeCote error
2, Do not carry out update when handling DKG outcome
    This shall be carried out AFTER handover completion
3, Do not promote elder when having enough elders
    This is to reduce the DKG & voting workload. May not be necessary anymore with other fixes deployed.
4, Do not carry out relocation within the genesis section
    The CI script currently doesn't support parsing relocated node info from logs
5, Remove incorrect cmds variable
    That's a coding bug

### DIFF
--- a/sn_node/src/node/mod.rs
+++ b/sn_node/src/node/mod.rs
@@ -49,6 +49,7 @@ use sn_dbc::Dbc;
 use sn_fault_detection::IssueType;
 use sn_interface::{
     dbcs::gen_genesis_dbc,
+    elder_count,
     messaging::{
         signature_aggregator::{SignatureAggregator, TotalParticipationAggregator},
         system::{DkgSessionId, NodeMsg, SectionSigned},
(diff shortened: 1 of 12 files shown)


sn_node/src/node/mod.rs
```diff
@@ -413,10 +414,16 @@ impl MyNode {
             return vec![zero_dkg_id, one_dkg_id];
         }
 
+        let current_elders = BTreeSet::from_iter(sap.elders().copied());
```
joshuef, line 417: This would prevent any splits, right?

maqi: yeah, you are right.
However our current CI tests don't involve any split yet.
will correct it anyway.

oetyng (commented):
I'm missing a description of the PR.

oetyng (approved):

PR #2247, opened 2023-03-25
feat(fees): impl supply/demand-adjusted fee

oetyng:
This PR adds an algorithm to calculate a fee with supply/demand adjusting properties - although indirect and therefore arguably sluggish - based on:
1. Network size (contributing to the deflationary character of SNT).
2. Storage used.
3. Requirements for space margin (1/3).
4. Requirement for as low fee as possible for as long as possible.

(Point 3. and 4. are achieved by the specific design of the required tokens curve.)

***

To do transfers on the network, clients are required to pay a number of tokens as a fee. The fees are the source of rewards for Elders in a payment-network, which is essential to the network security as it gives incentive to node operators to run a node and handle the requests, thus allowing for the network to grow.

**NB 1**: This is not a replacement for a true market, where node operators and clients discover the price based on operation costs + margins and service value + affordability to client.

**NB 2**: The fee paid to the Elder is not actually accessible until the Elder can fetch the `SpendProof` from
the data holders.

**NB 3**: With 1 fee per spend, the fee amount in the dbc is not accessible to the Elder until the required fee for
a spend has decreased below the amount in the dbc. In effect, with this design there is currently a lock-in
effect on Elder rewards which require the network to grow for the amount paid to them to be accessible.
***
1. ✔️ `Previous PR`: The query and handling of it on Elders for their reward keys and hard coded fee: #2184.
2. ✔️ `Previous PR`: The logic for including Elders to outputs: #2180.
3. ✔️ `Previous PR`: Including fee ciphers for Elders in spend: #2249.
4. ✔️ `Previous PR`: Elders confirming that spends contain outputs with sufficient amount, destined for them: #2246.
5. `This PR`: Use a fee calc algo with supply/demand adjusting properties.
6. `Next PR`: Use a mempool to increase supply/demand properties of the system and protect the network #2256.

Steps in the fee/reward feature:
- <s> 1. Including Elders in outputs (any value) when spending Dbcs.</s>
- <s> 2. Including fee ciphers for Elders so that they can confirm fee paid.</s>
- <s> 3. Elders confirming that spends contain outputs with sufficient amount, destined for them.</s>
- <s> 5. Implementing a commonly known deterministic transfer/store cost paid to nodes.</s> (`This PR`)
- <s> 5. Including supply/demand adjusting fee in outputs when spending Dbcs.</s> (`This PR`)
6. A mempool which orders incoming spends by fee amount, and a Client UI option to choose priority of their spend.
7. Elders fetching and storing their payment to a local wallet.

PR #2246, opened 2023-03-25
feat(fees): verify amount at elder 

oetyng:
With this PR an Elder now verifies that a spend contains an output for them
with correct amount.

This is done by decrypting the fee ciphers sent from client in a spend,
confirming that there is an output to the key derived from the elder reward key,
and that the amount of the output is equivalent to the decrypted amount.

The basic rule for now is that if the paid fee is over 1% less than the required fee, we don't accept the fee.
This is a temporary diff value/design.
(And yes, this means that for now, the Client can try to get away cheaper by sending in up to 1% lower fee than they were asked for.)

***

1.  ✔️ `Previous feat`: Including Elders in outputs (any value) when spending Dbcs.
2. ✔️ `Previous feat`: Including fee ciphers for Elders in spend: #2249.
3. `This PR`: Elders confirming that spends contain outputs with sufficient amount, destined for them.
4. `Next PR`: Use a fee calc algo with supply/demand adjusting properties: #2247.

Steps in the fee/reward feature:
<s> 1. Including Elders in outputs (any value) when spending Dbcs.</s>
<s> 2. Including fee ciphers for Elders so that they can confirm fee paid.</s>
<s> 3. Elders confirming that spends contain outputs with sufficient amount, destined for them.</s> (`This PR`)
4. Implementing a commonly known deterministic transfer/store cost paid to nodes.
5. Including supply/demand adjusting fee in outputs when spending Dbcs.
6. Elders fetching and storing their payment to a local wallet.
(7. A mempool which orders incoming spends by fee amount, and a Client UI option to choose priority of their spend.)

***

PR #2249, opened 2023-03-27
feat(fees): include ciphers for elders in spend

oetyng:
The ciphers over derivation index and blinding factor are now included
in the spend request sent to elders.
**NB:** The elders do not yet verify that they have been paid.

Next PRs are #2246, and #2247.


sn_api/src/app/wallet.rs
```diff
@@ -368,9 +386,25 @@ impl Safe {
...
+            outputs_owners,
+            change_amount,
+            #[cfg(not(feature = "data-network"))]
+            all_fee_cipher_params,
+        } = components;
+
         // We can now reissue the output DBCs
         let (output_dbcs, change_dbc) = self
-            .reissue_dbcs(input_dbcs_to_spend, outputs_owners, reason, change_amount)
+            .reissue_dbcs(
```
maqi, line 400: maybe we shall change the `reissue_dbcs` function to take the `ReissueComponents` struct in directly?

sn_api/src/app/wallet.rs
```diff
@@ -468,55 +502,91 @@ impl Safe {
...
-            if required_num_mints > mint_fees.len() {
-                warn!("Not enough mints contacted for the section to spend the input. Found: {}, needed: {required_num_mints}", mint_fees.len());
+            let revealed_bearer = dbc.as_revealed_input_bearer()?;
+            let input_key = revealed_bearer.public_key();
+
+            // Each section will have elder_count() instances to pay individually (for now, later they will be more).
+            let elder_fees = client.get_section_fees(input_key).await?;
+            let num_responses = elder_fees.len();
+            let required_responses = supermajority(elder_count());
+            if required_responses > num_responses {
```
maqi, line 525: seems this block of check shall be carried out outside and before the for clause ?

sn_api/src/app/wallet.rs
```diff
@@ -468,55 +502,91 @@ impl Safe {
...
-            // TODO: Query the network, one section per input, for the current fee.
-            // Right now, now fees are added, as a dummy is used instead.
-
-            let input_key = dbc.as_revealed_input_bearer()?.public_key();
-            // each mint will have elder_count() instances to pay individually (for now, later they will be more)
-            let mint_fees: BTreeMap<PublicKey, Token> = client.get_mint_fees(input_key).await?;
-            let required_num_mints = supermajority(elder_count());
-            if required_num_mints > mint_fees.len() {
-                warn!("Not enough mints contacted for the section to spend the input. Found: {}, needed: {required_num_mints}", mint_fees.len());
+            let revealed_bearer = dbc.as_revealed_input_bearer()?;
```
maqi, line 518: better to catch the error and log/handle .
as this will terminate the for clause earlier

sn_api/src/app/wallet.rs
```diff
@@ -468,55 +502,91 @@ impl Safe {
...
+            // Fees that were not encrypted to us.
+            let mut invalid_fees = BTreeSet::new();
+            // As the mints encrypt the amount to our public key, we need to decrypt it.
+            let mut decrypted_elder_fees = vec![];
+
+            for (elder, fee) in elder_fees {
+                match fee.content.decrypt_amount(&revealed_bearer.secret_key) {
+                    Ok(amount) => decrypted_elder_fees.push(((elder, fee), amount)),
+                    Err(_) => {
+                        let _ = invalid_fees.insert(fee.content.elder_reward_key);
```
maqi, line 539: shall we log the error to show why the decrypt_amount failed?

sn_api/src/app/wallet.rs
```diff
@@ -468,55 +502,91 @@ impl Safe {
         spendable_dbcs: WalletSpendableDbcs,
         mut total_output_amount: Token,
         mut outputs_owners: Vec<(Token, OwnerOnce)>,
-    ) -> Result<SelectedInputs> {
+    ) -> Result<ReissueComponents> {
```
maqi, line 505: the function becomes a bit long one now, especially the for clause block.
any possible to break it down?

sn_api/src/app/wallet.rs
```diff
@@ -628,23 +704,41 @@ impl Safe {
...
+        //     .revealed_outputs
+        //     .iter()
+        //     .map(|output| (output.public_key, output.revealed_amount))
+        //     .collect();
+
         // Spend all the input DBCs, collecting the spent proof shares for each of them
         for (public_key, tx) in dbc_builder.inputs() {
+            // Generate the fee ciphers.
+            let input_fee_ciphers = BTreeMap::new();
+            // Enable this when fees are verified, and tests fixed.
```
maqi, line 733: TODO ?

sn_interface/src/types/keys/secret_key.rs
```diff
@@ -82,3 +82,26 @@ impl Display for SecretKey {
...
+        Error::FailedToParse(format!(
+            "Couldn't parse BLS secret key bytes from hex: {err}",
+        ))
+    })?;
+    let bytes_fixed_len: [u8; bls::SK_SIZE] = bytes.as_slice().try_into()
+        .map_err(|_| Error::FailedToParse(format!(
+            "Couldn't parse BLS secret key bytes from hex. The provided string must represent exactly {} bytes.",
+            bls::SK_SIZE
+        )))?;
+    let pk = bls::SecretKey::from_bytes(bytes_fixed_len).map_err(|err| {
```
maqi, line 100: the variable name shall be `sk` ?

sn_node/src/node/api/node_starter.rs
```diff
@@ -52,27 +54,33 @@ pub async fn new_node(config: &Config, join_retry_timeout: Duration) -> Result<N
     let root_dir = root_dir_buf.as_path();
     fs::create_dir_all(root_dir).await?;
 
-    let reward_key = match get_reward_pk(root_dir).await? {
-        Some(public_key) => public_key,
+    let reward_secret_key = match get_reward_secret_key(root_dir).await? {
```
maqi, line 57: not sure about the whole story, so just a confirmation:
as I understand, the reward_key will be shown to others, but now we use a `secret_key` ?

or you mean we just show the public part to outsider, and keep the secret part locally? 
if say, maybe name it as reward_key_pair?

sn_node/src/node/context.rs
```diff
@@ -30,7 +29,7 @@ pub struct NodeContext {
     pub(crate) name: XorName,
     pub(crate) info: MyNodeInfo,
     pub(crate) keypair: Arc<Keypair>,
-    pub(crate) reward_key: PublicKey,
+    pub(crate) reward_secret_key: Arc<bls::SecretKey>,
```
maqi, line 32: actually, this sounds to me not a section_key but an owner key.
i.e. no need to use bls key, an ed key shall be enough?

sn_node/src/node/messaging/client_msgs.rs
```diff
@@ -285,32 +313,14 @@ impl MyNode {
...
+    #[cfg(not(feature = "data-network"))]
     fn verify_fee(
         _store_cost: sn_dbc::Token,
-        _our_key: PublicKey,
+        _reward_secret_key: &bls::SecretKey,
         _tx: &DbcTransaction,
+        _our_name: XorName,
+        _fee_ciphers: BTreeMap<XorName, FeeCiphers>,
     ) -> Result<()> {
-        // TODO: check that we have an output to us, and that it is of sufficient value.
```
maqi, line 293: this TODO block shall be retained? 
or at least the comment part?

sn_api/src/app/wallet.rs
oetyng: No, this must be in the for clause, because we are querying for fees for _every dbc_ that we are about to spend. That is a query to potentially different section every time.
While in a small network we could try to optimize by batching (seeing which dbc key goes to same section), but I doubt that is worth it in larger network.

sn_api/src/app/wallet.rs
oetyng: Yeah, you're right 👍

sn_api/src/app/wallet.rs
oetyng: It comes in the follow up PRs. (You can have a look in them if you want, it's in #2247).

sn_api/src/app/wallet.rs
oetyng: Yeah, at least we could do something about that fn signature.
We're not using the `input_dbcs_entries_hash` in reissue fn, and we don't have `reason` in the components.

There's more work to slim down this part in #2247. Maybe we could try work on this stuff after that one?

sn_api/src/app/wallet.rs
oetyng: Yes. Good.

sn_api/src/app/wallet.rs
```diff
@@ -628,23 +704,41 @@ impl Safe {
...
             .flat_map(|dbc| dbc.inputs_spent_transactions.clone())
             .collect();
 
-        let proof_key_verifier = SpentProofKeyVerifier { client };
-
-        // Let's build the output DBCs
+        // Finalize the tx builder to get the dbc builder.
         let mut dbc_builder = tx_builder.build(rng::thread_rng())?;
 
+        // Enable this when fees are verified, and tests fixed.
```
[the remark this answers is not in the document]
oetyng: This part will conflict when I rebase the follow up #2246, and all tests will fail without it, so don't think it will be necessary.

sn_api/src/app/wallet.rs
oetyng: Same as above.

sn_api/src/app/wallet.rs
```diff
@@ -713,12 +808,34 @@ impl Safe {
     }
 }
 
-type SelectedInputs = (
-    Vec<Dbc>,
-    BTreeSet<EntryHash>,
-    Vec<(Token, OwnerOnce)>,
-    Token,
-);
+#[allow(unused)]
```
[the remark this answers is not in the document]
oetyng: No, it would still be warned as unused (and fail when using `-D` flag to clippy).

sn_node/src/node/api/node_starter.rs
oetyng: > or you mean we just show the public part to outsider, and keep the secret part locally?

Yes this is the case.

> maybe name it as reward_key_pair?

I was actually considering this at first. But I thought it was a bit off as we have a KeyPair type, and this is actually the secret key so not wrong to call it the secret key? And it's known that pks are derived from secret keys.

Well, that was my reasoning anyway.

sn_api/src/app/wallet.rs
maqi: ok, then. 
just a remind that this will then may terminate the whole on any failure

joshuef (approved):

PR #2251, opened 2023-03-28
Send enqueue by default for send msg

joshuef:
Sets us up to use bidi streams for more AE responses and the message handling there.

We were hitting more `ContactingUnkownNode` type errors than we should have been. This removes that issue


sn_node/src/node/messaging/mod.rs
```diff
@@ -192,11 +192,29 @@ impl MyNode {
...
-                cmds.push(MyNode::send_ae_update_to_nodes(
-                    &context,
-                    Recipients::Single(sender),
-                    section_key,
-                ));
+
+                if let Some(stream) = send_stream {
+                    let msg = MyNode::generate_ae_update_msg(&context, section_key);
+
+                    // cmds.push(Cmd::UpdateCallerOnStream{
```
oetyng, line 199: Can be cleaned up.

sn_node/src/node/messaging/mod.rs
```diff
@@ -192,11 +192,29 @@ impl MyNode {
...
+                    //     caller: sender,msg_id:MsgId::new(),kind: AntiEn,
+
+                    // })
+                    if let Some(cmd) =
+                        MyNode::send_ae_response(msg, MsgId::new(), sender, msg_id, stream, context)
+                            .await?
+                    {
+                        cmds.push(cmd)
+                    };
+                    // cmds.push(Cmd::SendNodeMsgResponse { msg, msg_id: MsgId::new(), correlation_id: msg_id, sender, send_stream })
```
oetyng, line 209: And here.

oetyng (dismissed):

oetyng (approved):

